PR #105, opened 2020-04-28
fix: more safe check for getTitle

suevalov:


packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -1,13 +1,20 @@
...
     return defaultTitle;
-  } else {
-    return title;
   }
+  if (title) {
+    // check if title consist only from whitespace symbols
+    if (typeof title.match === 'function' && title.match(/^\s*$/)) {
+      return defaultTitle;
+    }
+    return title.trim();
```
johanneswuerbach, line 15: Maybe instead of the extra regex?

```suggestion
    const trimmedTitle = title.trim();
    if (trimmedTitle.length === 0) {
      return defaultTitle;
    }
    return trimmedTitle;
```

suevalov: Yeah, it's a good idea. Thanks!

packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -1,13 +1,20 @@
...
-    return title;
   }
+  if (title) {
+    const trimmedTitle = title.trim();
+    if (trimmedTitle.length === 0) {
+      return defaultTitle;
+    }
+    return trimmedTitle;
+  }
+  return defaultTitle;
```
DanweDE, line 17: replacing the above with only

```js
  const trimmedTitle = title.trim();
  if (trimmedTitle.length === 0) 
     return defaultTitle;
  }
  return trimmedTitle;
```

should be sufficient and saves you one `if`.

PR #127, opened 2020-05-05
feat: use new sdk.access to read permissions

suevalov:
Use new `sdk.access.can` methods for all field-editors to read permissions.

### DIFF
--- a/packages/markdown/src/MarkdownEditor.tsx
+++ b/packages/markdown/src/MarkdownEditor.tsx
@@ -44,7 +44,7 @@ export function MarkdownEditor(
   const [currentValue, setCurrentValue] = React.useState<string>(props.initialValue ?? '');
   const [selectedTab, setSelectedTab] = React.useState<MarkdownTab>('editor');
   const [editor, setEditor] = React.useState<InitializedEditorType | null>(null);
-  const [canUploadAssets, canUploadAssetsSet] = React.useState<boolean>(false);
+  const [canUploadAssets, setCanUploadAssets] = React.useState<boolean>(false);
 
   React.useEffect(() => {
     if (editor && props.onReady) {
@@ -58,7 +58,7 @@ export function MarkdownEditor(
 
   React.useEffect(() => {
     props.sdk.access.can('create', 'Asset').then(value => {
-      canUploadAssetsSet(value);
+      setCanUploadAssets(value);
     });
   }, []);
 
--- a/packages/markdown/src/MarkdownEditor.tsx
+++ b/packages/markdown/src/MarkdownEditor.tsx
@@ -29,14 +29,7 @@ export interface MarkdownEditorProps {
   isInitiallyDisabled: boolean;
 
   sdk: FieldExtensionSDK;
-  /**
-   * sdk.parameters
-   */
-  parameters: {
-    instance: {
-      canUploadAssets: boolean;
-    };
-  };
+
   previewComponents?: PreviewComponents;
   onReady?: Function;
 }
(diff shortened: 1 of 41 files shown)


packages/markdown/src/MarkdownEditor.tsx
```diff
@@ -29,14 +29,7 @@ export interface MarkdownEditorProps {
   isInitiallyDisabled: boolean;
 
   sdk: FieldExtensionSDK;
-  /**
```
Jwhiles, line 32: So we don't want to make it possible to disable the markdown editor being able to upload assets from an extension?

suevalov: Yeah, not at the moment. It was exposed only for Web App being able to pass the value.
We might get back to it and expose all buttons as instance parameters.

packages/markdown/src/MarkdownEditor.tsx
```diff
@@ -51,6 +44,7 @@ export function MarkdownEditor(
   const [currentValue, setCurrentValue] = React.useState<string>(props.initialValue ?? '');
   const [selectedTab, setSelectedTab] = React.useState<MarkdownTab>('editor');
   const [editor, setEditor] = React.useState<InitializedEditorType | null>(null);
+  const [canUploadAssets, canUploadAssetsSet] = React.useState<boolean>(false);
```
DanweDE, line 47: I was very confused when I read `canUploadAssetsSet`. This should be named `setCanUploadAssets` as the relevant verb is `set`, not `can` in this case "set whether we can upload an asset".

packages/rich-text/src/RichTextEditor.spec.js
```diff
@@ -18,13 +18,17 @@ const fakeProps = props => ({
...
       locale: 'FIELD_LOCALE'
     },
-    parameters: {
-      instance: {
-        permissions: {
-          canAccessAssets: true,
-          canCreateAssets: true,
-          canCreateEntryOfContentType: () => false
+    access: {
+      can: (access, entityType) => {
```
DanweDE, line 22: Nit just make this an `async` fn and `return`?

packages/rich-text/src/plugins/CommandPalette/CommandPaletteService.js
```diff
@@ -86,40 +86,45 @@ export const isEmbeddingEnabled = field =>
...
 
     const isAsset = !contentType;
     if (!isAsset) {
-      if (!isValidLinkedContentType(this.field, contentType, embedType)) {
+      if (!isValidLinkedContentType(this.sdk.field, contentType, embedType)) {
         return false;
       }
-      if (!this.permissions.canCreateEntryOfContentType(contentType.sys.id)) {
+
+      const canCreateContentType = await this.sdk.access.can('create', contentType);
```
DanweDE, line 118: :exclamation: Wait... whether or not you can create a CT vs. whether or not you can create an entry of a certain CT is something completely different. Is it just a naming issue here and this should be named `canCreateEntryOfCT` or is your intention to call `this.sdk.access.can('create', 'ContentType')`?

I guess it's a naming issue but then I am confused by `await this.sdk.access.can('create', contentType);`. I am looking [at the documentation](https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/) and I must say it is pretty confusing for this case. I believe `can` does too many things in one single function, or maybe it can't do what you're trying to do here?

E.g. `can('create', 'Entry', contentType)` would make more sense to me, otherwise, look at `can('update', contentType)` vs. `can('create', contentType)` where it's pretty ambiguous (and not documented) what either of these would actually verify exactly :thinking:

suevalov: Good catch! It was incorrect indeed. Now it should be fixed.

Issue #69, opened 2020-03-20
Disable the ability to focus on Markdown field if the field is disabled


suevalov:
Allow the same styling for the field as we do for rich text editor

djagya:
@suevalov Now I can close this issue, correct?

suevalov:
@djagya yep, I'm closing it. thanks 🎖️

PR #166, opened 2020-05-13
fix: add disabled state styles for the markdown field

djagya:
Properly fade the editor area and hide the blinking caret cursor.


suevalov:
Fixes #69

PR #203, opened 2020-05-26
fix: use dialogs.openCurrent to open dialogs

FLoppix:
The field editors throw an error if you open any form of a dialog from within an app. The method `openCurrent` handles this behaviour and decides if you should `openExtension` or `openCurrentApp`.


suevalov:
@FLoppix I can publish new packages once you merge this.

suevalov:
@FLoppix After publishing, we need to bump all versions in user_interface and rename `openExtension` in `createNewWidgetApi` there

FLoppix:
@suevalov ya exactly that was my follow up task. Then I think we have everything covered 👍

PR #204, opened 2020-05-28
fix: prevent rich text references from opening on click

m10l:
When clicking on reference cards within the rich text editor, the entry is immediately opened. This PR changes that so that clicking on a reference card only selects the card which allows the user to then press backspace to remove the referenced entry from the rich text editor field.

### DIFF
--- a/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
@@ -35,7 +35,12 @@ interface WrappedEntryCardProps {
   allContentTypes: ContentType[];
   entry: Entry;
   cardDragHandle?: React.ReactElement;
+  isClickable: boolean;
 }
+
+const defaultProps = {
+  isClickable: true,
+};
 
 export function WrappedEntryCard(props: WrappedEntryCardProps) {
   const [file, setFile] = React.useState<null | File>(null);
@@ -93,7 +98,9 @@ export function WrappedEntryCard(props: WrappedEntryCardProps) {
 
   return (
     <EntryCard
-      href={props.getEntryUrl ? props.getEntryUrl(props.entry.sys.id) : undefined}
+      href={
+        props.getEntryUrl && props.isClickable ? props.getEntryUrl(props.entry.sys.id) : undefined
+      }
       title={title}
       description={description}
       contentType={contentType?.name}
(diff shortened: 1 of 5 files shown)


packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
```diff
@@ -93,7 +98,9 @@ export function WrappedEntryCard(props: WrappedEntryCardProps) {
 
   return (
     <EntryCard
-      href={props.getEntryUrl ? props.getEntryUrl(props.entry.sys.id) : undefined}
+      href={
+        props.getEntryUrl && props.isClickable ? props.getEntryUrl(props.entry.sys.id) : undefined
```
djagya, line 102: I think the same should be done for assets if we want to be consistent on how both cards are handled?

i.e. `packages/rich-text/src/plugins/EmbeddedEntityBlock/FetchingWrappedAssetCard.js` and `packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx`

djagya, line 102: I think the same should be done for assets if we want to be consistent on how both cards are handled?

i.e. `packages/rich-text/src/plugins/EmbeddedEntityBlock/FetchingWrappedAssetCard.js` and `packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx`

packages/rich-text/src/plugins/EmbeddedEntityBlock/FetchingWrappedEntryCard.js
```diff
@@ -52,6 +52,7 @@ export function FetchingWrappedEntryCard(props) {
       entry={entry}
       onEdit={props.onEdit}
       onRemove={props.onRemove}
+      isClickable={false}
```
djagya, line 55: Should also be applied to `FetchingWrappedAssetCard.js` IIUC

packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
m10l: Ah good catch, I forgot to check assets 🤦

m10l: Updated assets, I'll mark this and the other comment as resolved

PR #205, opened 2020-05-28
fix: 🐛 Improve link rendering performance

DanweDE:
improves performance on rendering lots of links as we were doing `n^2` renderings of the link card where `n` is the number of links. We now do ~`n*2` renderings (one for loaded state, one for fetched link state)


DanweDE:
*Note:* With this change we're still calling each component `n^2` times because of how the entity store works, which really isn't ideal.

Also, the issue probably is also present for assets and should be fixed in that case as well.

PR #206, opened 2020-05-29
fix(reference): optimize multiple references field rendering

djagya:
The `useEntities` hook used inside `FetchingWrappedEntryCard` and `FetchingWrappedAssetCard` provides the list of all entities `entries`/`assets` from the store and the `loadEntry`/`loadAsset` function. The problem is that each successful load call triggers a re-render of all components using the hook, even though we fetch all entities in one request because of optimization. So a list of 500 cards ends up being rendered 500^2 times before the page is fully loaded.

The recent fix https://github.com/contentful/field-editors/commit/f6e844dcd2f25ff5453d2ce6301ea021b5fcaca3 added `useMemo` inside `FetchingWrappedEntryCard` to re-render only when an entity changes. But it doesn't consider component props and introduced a few bugs, like card not updating after an edit or cards not switching to initialized state (when a `DragHandle` is displayed and `Remove` button is enabled).

To fix the bugs, `props` dependency must be added to `useMemo`, but even then there are **15** initial re-renders, where most of them are not needed and happen only because the parent `SortableLinkList` is re-rendered and passed down new instances of `onRemove` and `cardDragHandler`, whereas all other card props are the same. Maybe I'm missing some obvious way of optimizing it (like memozing `onRemove` and `cardDragHandler`)?

In this PR I added the same technic for assets (`useMemo` in `FetchingWrappedAssetCard`).

### DIFF
--- a/packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
+++ b/packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
@@ -25,6 +25,12 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
   }, [props.assetId]);
 
   const asset = assets[props.assetId];
+  const entityKey =
+    asset === 'failed'
+      ? 'failed'
+      : asset === undefined
+      ? 'undefined'
+      : `:${asset.sys.id}:${asset.sys.version}`;
 
   React.useEffect(() => {
     if (asset) {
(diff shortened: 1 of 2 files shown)


packages/reference/src/assets/SortableElements.tsx
```diff
@@ -34,25 +34,32 @@ const SortableLink = SortableElement((props: { children: React.ReactElement }) =
...
+                  })
+                );
+              }}
+              cardDragHandle={props.isDisabled ? undefined : <DragHandle />}
+            />
+          </SortableLink>
+        ))}
+      </div>
+    ),
+    [itemsKey, props.isDisabled, props.viewType]
```
djagya, line 63: Probably all props must be listed here, but then there's no point in `useMemo`. I need to track what props make the list re-render because when I inspected them, it seemed like all of them should not be changing after the first render (and user actions).

djagya, line 63: This is not correct, all props should be listed here.

packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
```diff
@@ -25,61 +25,69 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
   }, [props.assetId]);
 
   const asset = assets[props.assetId];
+  const entityKey =
```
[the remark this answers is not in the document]
DanweDE: Not sure, I'd prefer if the store would give you some kind of result object rather than overloading this into `Enity|string|undefined`

If you want the helper you're mentioning, I think you could add it to the store as it's directly coupled to the data structure the store is exposing.

djagya:
I will merge and release this and will consider creating a follow up PR where we can check how the EntityStore return object could be changed.

PR #223, opened 2020-06-05
fix: allow draggable embeds in rich text

m10l:
This PR allows embeds in the rich text editor to be draggable for purposes of moving within the rich text content, but will still prevent the embedded entry/asset cards from opening on click


m10l:
> The missing `href` made them un-draggable?

Yeah, dragging embeds around in the rich text field was ever intended behaviour but it's working by coincidence. I think the behaviour showed up when we open-sourced the rich text editor with the incorrect behaviour for clicking on cards

PR #296, opened 2020-07-07
feat: add validation-errors component

z0al:
This PR introduces a new component `ValidationErrors` that will be used render validation errors in the Contentful web application and other internal services.


packages/validtion-errors/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "test": "tsdx test --env=jsdom --watch",
+    "test:ci": "tsdx test --env=jsdom --ci",
+    "tsc": "tsc -p ./ --noEmit"
+  },
+  "dependencies": {
+    "@contentful/field-editor-shared": "^0.11.0",
+    "@contentful/forma-36-react-components": "^3.40.1",
+    "@contentful/forma-36-tokens": "^0.7.1",
+    "emotion": "^10.0.17",
+    "lodash": "^4.17.15",
```
suevalov, line 28: Is lodash used here?

packages/validtion-errors/package.json
```diff
@@ -0,0 +1,47 @@
+{
```
suevalov, line 1: I noticed a type in the folder name: `validAtion-errors`

packages/validtion-errors/src/ValidationErrors.mdx
```diff
@@ -0,0 +1,48 @@
+---
```
suevalov, line 1: Not sure if we need a playground example for this component. I'd rather integrate this example into a complex `entry-editor` example that we have in `extensions` folder

packages/validtion-errors/src/ValidationErrors.tsx
```diff
@@ -0,0 +1,142 @@
...
+        title: getTitle(entry),
+        href: props.getEntryURL(entry),
+      }));
+
+      setState({
+        loading: false,
+        entries,
+      });
+    });
+  }, [props]);
```
suevalov, line 71: Are we sure that we want to react on all props changes here?

packages/validtion-errors/src/ValidationErrors.tsx
```diff
@@ -0,0 +1,142 @@
+import * as React from 'react';
+import type * as shared from '@contentful/field-editor-shared';
```
suevalov, line 2: Could you use named imports here? We never used such imports in this repo and it would be nice to be consistent.

packages/validtion-errors/src/ValidationErrors.tsx
z0al: Realistically we want to only refetch when the error changes but I also don't want to lie to React about the dependencies since it may cause some tricky bugs in end-users code.

What we can do is comparing the `error.conflicting` ids with those we have in `state.entries` and only fetch different, how about that @suevalov ?

suevalov: sounds good

z0al: Done. Let me know what you think now

packages/validation-errors/package.json
```diff
@@ -25,8 +25,7 @@
     "@contentful/forma-36-react-components": "^3.40.1",
     "@contentful/forma-36-tokens": "^0.7.1",
     "emotion": "^10.0.17",
-    "lodash": "^4.17.15",
-    "lodash-es": "^4.17.15"
+    "lodash": "4.17.16"
```
suevalov, line 28: Could you use "lodash": "^4.17.15" so we don't have a concrete locked version here. With a concrete locked version you often end up with several lodash packages in the final bundle.

z0al: Oh, yeah, my bad :+1:

anho: Or you simply get rid of lodash completely in this package by doing a diff function on your own 😄

z0al: @anho done. You can enjoy your vacation now :wink:

packages/validtion-errors/src/ValidationErrors.mdx
z0al: Will be done in a follow up PR.

PR #300, opened 2020-07-10
fix(reference): trigger 'rendered' event in multiple references once all entities are loaded

djagya:
We want to track when **all** entities are loaded in a multiple reference field. To do so `rendered` events coming from `FetchingWrapped*Card` must be counted until to determine when all of them are loaded, then the actual `rendered` event can be triggered.

This is implemented as a `useMultipleReferenceLoadTracking` hook which intercepts and counts events passed to `onAction`.

### DIFF
--- a/packages/reference/src/types.ts
+++ b/packages/reference/src/types.ts
@@ -8,7 +8,7 @@ export {
   ContentTypeField,
   Link,
   EntityType,
-  NavigatorSlideInfo
+  NavigatorSlideInfo,
 } from 'contentful-ui-extensions-sdk';
 
 export { Entry, File, Asset } from '@contentful/field-editor-shared';
@@ -55,4 +55,4 @@ export type Action =
       slide?: NavigatorSlideInfo;
     }
   | { type: 'delete'; contentTypeId: string; id: string; entity: EntityType }
-  | { type: 'rendered'; entity: EntityType };
+  | { type: 'rendered'; entity?: EntityType };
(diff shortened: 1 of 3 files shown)


djagya:
I was thinking about it, I know it doesn't look that nice. But we need some way to tell when all entities are rendered/fetched. We can't know the number of links outside `FieldConnector` and we can't know the number in `EntityStore`. So this place was the only one I found where I could put this logic.

packages/reference/src/types.ts
```diff
@@ -55,4 +55,4 @@ export type Action =
       slide?: NavigatorSlideInfo;
     }
   | { type: 'delete'; contentTypeId: string; id: string; entity: EntityType }
-  | { type: 'rendered'; entity: EntityType };
+  | { type: 'rendered'; entity?: EntityType };
```
DanweDE, line 58: Shouldn't we just add another event for the final overall "rendered"? If we listen to this event from the outside we still the `numberOfLinks+1` actions for `rendered` (if I am not mistaken) while we're only interested in the last one without the `entity`.

djagya: No, in the hook we listen to `rendered` event, but don't emit it further, instead waiting for all `numberOfLinks` events to be triggered to switch the `rendered` state: https://github.com/contentful/field-editors/pull/300/files#diff-07c0b93de1f5dd0e8d47eba16bd83554R28-R34

Once `rendered = true` **state**, we emit the actual `rendered` **event**: https://github.com/contentful/field-editors/pull/300/files#diff-07c0b93de1f5dd0e8d47eba16bd83554R17-R21

Adding a new event like `finalRendered` would require emitting this even in other places instead of `rendered`, or distinguishing between `rendered` coming from other fields and the multiple reference field.

djagya:
Found the way to implement it in our web app, so maybe this PR would not be required.

PR #336, opened 2020-08-06
refactor: 💡 share code between RT heading dropdown items

DanweDE:
- DRY up headings related editor toolbar plugins.
 - Remove some broken logic around bold headings.


packages/rich-text/src/plugins/shared/MarkPlugin.js
```diff
@@ -1,17 +1,16 @@
 import isHotkey from 'is-hotkey';
 import markDecorator from './MarkDecorator';
 import { haveMarks } from './UtilHave';
-import { css, cx } from 'emotion';
+import { css } from 'emotion';
 import tokens from '@contentful/forma-36-tokens';
 
 const styles = {
+  // TODO: Ensure a bold mark on a headline makes the heading appear even
+  //  bolder. Consider to make headings non-bold by default.
```
cribbles, line 9: This would necessitate a chat with @m10l or @joaoramos , no?

There may also be a ticket for it somewhere waaay in the vaults - I remember us talking about this during the initial implementation of the rich text editor.

packages/rich-text/src/plugins/shared/MarkPlugin.js
```diff
@@ -1,17 +1,16 @@
...
   bold: css({
     color: 'inherit',
     fontWeight: tokens.fontWeightDemiBold,
   }),
-  headingBold: css({
-    fontWeight: 900, // Make headings which are already bold even bolder.
-  }),
```
cribbles, line 14: And to be clear, this is the "broken" logic? i.e., this never really made the headings bolder? (hence why we can get away with no more RegExp parsing)

DanweDE: It did at some point (I think I originally implemented it and have tested it for sure) but I guess in some slate upgrade or refactoring it just didn't work anymore.

packages/rich-text/src/plugins/shared/MarkPlugin.js
DanweDE: Yeah, that's why I am writing "consider" ;)
We might have decided against it in the past. Personally I still find headlines way too bold in CF RT, all the way down to h5 or h5 where bold might make sense since the text is just getting too small.

PR #337, opened 2020-08-06
fix: 🐛 vertical jitter effect when applying rich text code mark

DanweDE:
Fixes the following unpleasant effect whenever applying a code mark (also applies for normal text, not just headings)

jitter


DanweDE:
Not sure whether there's any better way @m10l, it's just a quick thing I've stumbled across and seems to work quite well.

PR #350, opened 2020-08-12
feat: 🎸 refined editor focusing when using toolbar actions

DanweDE:
When the user clicks on a Rich Text editor toolbar action while the
editor is not focused, this should ideally result in focusing the editor
and applying the clicked action.

This handling has been implemented for some actions and it has been
breaking in some cases. This change introduces consistent handling and
dealing with Slate related limitations.

The main limitaton would be after initializing Rich Text, before the
user has manually focused the rich text field, we can not set the focus
programmatically as the Slate internal value seems to lack some
properties that can result in a crash at this point. We circumvent bugs
by preventing any action when clicking on a toolbar action. Instead, the
user has to focus the input first. This is pretty much consistent with
the previous behavior but implemented in a way that is able to prevent
any crashes. Ideally we would improve the behaivor once we figure out
how to prevent Slate from having these missing properties or after
upgrading to a newer Slate 0.5x version.

### DIFF
--- a/packages/rich-text/src/plugins/Hyperlink/ToolbarIcon.js
+++ b/packages/rich-text/src/plugins/Hyperlink/ToolbarIcon.js
@@ -2,21 +2,21 @@ import React, { Component } from 'react';
 import { INLINES } from '@contentful/rich-text-types';
 import ToolbarIcon from '../shared/ToolbarIcon';
 import { TOOLBAR_PLUGIN_PROP_TYPES } from '../shared/PluginApi';
+import { toolbarActionHandlerWithSafeAutoFocus } from '../shared/Util';
 import { hasHyperlink, toggleLink, hasOnlyHyperlinkInlines } from './Util';
 
 export default class HyperlinkToolbarIcon extends Component {
   static propTypes = TOOLBAR_PLUGIN_PROP_TYPES;
 
-  handleClick = async event => {
-    event.preventDefault();
+  handleClick = toolbarActionHandlerWithSafeAutoFocus(this, async () => {
     const {
       onToggle,
       editor,
-      richTextAPI: { sdk, logToolbarAction }
+      richTextAPI: { sdk, logToolbarAction },
     } = this.props;
     await toggleLink(editor, sdk, logToolbarAction);
     onToggle(editor);
-  };
+  });
 
   render() {
     const { disabled, editor } = this.props;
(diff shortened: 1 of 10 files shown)


packages/rich-text/src/plugins/Hyperlink/ToolbarIcon.js
```diff
@@ -27,7 +27,7 @@ export default class HyperlinkToolbarIcon extends Component {
         type={INLINES.HYPERLINK}
         icon="Link"
         title="Hyperlink"
-        onToggle={event => this.handleClick(event)}
+        onToggle={(event) => this.handleClick(event)}
```
cribbles, line 30: Can we just pass the function by reference here? i.e.
```jsx
onToggle={this.handleEvent}
```
since it's already instance-bound?

packages/rich-text/src/dialogs/HypelinkDialog/HyperlinkDialog.jsx
```diff
@@ -211,7 +211,7 @@ export class HyperlinkDialog extends React.Component {
             textInputProps={{
               placeholder: 'https://',
               testId: 'link-uri-input',
-              autoFocus: true,
+              canAutoFocus: true,
```
DanweDE, line 214: Almost introduced a bug here as I did a global `autoFocus` -> `canAutoFocus` replacement at some point :scream:

packages/rich-text/src/plugins/Hyperlink/ToolbarIcon.js
DanweDE: :+1: I knew I missed it in one of those files.

DanweDE:
### Test-plan
after page/entry with a RT field loaded in the editor, for each RT toolbar action <action> **cases**:
- click <action> before manually focusing RT.
  - [x] **Result:** Nothing is expected to happen. RT won't be programmatically focused.
 - click into RT to focus the input, then click RT <action>
   - [x] **Result:** RT is still focused, the action gets applied, toolbar icon toggles if applicable (not relevant for HR and hyperlink)
   - [x] In case of <action> opening a modal, the RT editor is focused after cancelling the dialog or confirming it
 - click into RT to focus, then place focus outside RT, then click RT <action>
   - [x] **Result:** RT gets focus, the action gets applied at previous cursor position, toolbar icon toggles if applicable (not relevant for HR) 
   - [x] In case of <action> opening a modal, the RT editor is focused after cancelling the dialog or confirming it

PR #354, opened 2020-08-13
fix: users should be able to remove checkbox with invalid predefined values

ChidinmaOrajiaku:
When a predefined checked checkbox value is invalid and page is reloaded, the value disappears which makes it impossible for users to unselect or remove invalid value.

This PR makes it possible for the user to see the invalid value and be able to remove it.

image

### DIFF
--- a/packages/checkbox/src/CheckboxEditor.tsx
+++ b/packages/checkbox/src/CheckboxEditor.tsx
@@ -27,7 +27,7 @@ export interface CheckboxEditorProps {
 
 type ListValue = string[];
 
-type CheckboxOption = { id: string; value: string; label: string };
+type CheckboxOption = { id: string; value: string; label: string; invalid?: boolean };
 
 function isEmptyListValue(value: ListValue | null) {
   return value === null || value.length === 0;
@@ -55,17 +55,22 @@ export function getOptions(field: FieldAPI): CheckboxOption[] {
   }));
 }
 
-const getMergedOptions = (values: string[], options: CheckboxOption[]): CheckboxOption[] => {
+const getInvalidValues = (
+  field: FieldAPI,
+  values: string[],
+  options: CheckboxOption[]
+): CheckboxOption[] => {
   const getValueFromOptions = (options as CheckboxOption[]).map((item) => item.value);
   const invalidValues = values
     .filter((value) => !getValueFromOptions.includes(value))
     .map((value, index) => ({
-      id: `invalid-${index}`,
+      id: ['entity', field.id, field.locale, index, 'invalid'].join('.'),
       label: value,
+      invalid: true,
       value,
     }));
 
-  return [...options, ...invalidValues];
+  return invalidValues;
 };
 
 export function CheckboxEditor(props: CheckboxEditorProps) {
@@ -99,7 +104,8 @@ export function CheckboxEditor(props: CheckboxEditorProps) {
           setValue(newValues);
         };
 
-        const mergedOptions = getMergedOptions(values, options);
+        const invalidValues = getInvalidValues(field, values, options);
+        const mergedOptions = [...options, ...invalidValues];
 
         return (
           <Form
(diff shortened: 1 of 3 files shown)


packages/checkbox/src/CheckboxEditor.tsx
```diff
@@ -86,29 +99,41 @@ export function CheckboxEditor(props: CheckboxEditorProps) {
           setValue(newValues);
         };
 
+        const mergedOptions = getMergedOptions(values, options);
```
freakyfelt, line 102: Given these are all internal methods what would you think about having `getMergedOptions` return `{ options, invalid }`

packages/checkbox/src/CheckboxEditor.tsx
```diff
@@ -86,29 +99,41 @@ export function CheckboxEditor(props: CheckboxEditorProps) {
...
+                    if (e.target.checked) {
+                      addValue(item.value);
+                    } else {
+                      removeValue(item.value);
+                    }
+                  }}
+                />
+                {item.id.includes('invalid') && (
+                  <>
+                    <span className={styles.invalidText}>(invalid)</span>
```
suevalov, line 130: I'm not sure if this is a sufficient explanation for the customer. It's not actionable and I don't understand why it's invalid even from the code :)

packages/checkbox/src/CheckboxEditor.tsx
```diff
@@ -86,29 +99,41 @@ export function CheckboxEditor(props: CheckboxEditorProps) {
...
+                  name={`${field.id}.${field.locale}`}
+                  onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                    if (e.target.checked) {
+                      addValue(item.value);
+                    } else {
+                      removeValue(item.value);
+                    }
+                  }}
+                />
+                {item.id.includes('invalid') && (
```
suevalov, line 128: Could we add `invalid: true` to `item` object instead of the modifying id.

packages/checkbox/src/CheckboxEditor.tsx
freakyfelt: Original problem is when a user removes an option from a valid set of values (e.g. must be one of 'one', 'two', or 'three'). When an entry has that value there is currently no way in the checkbox view to remove that invalid value

suevalov: Let's consult with a designer (@joaoramos) about the copy. 

One alternative solution would be to call `removeValue` automatically in case of misconfiguration.

freakyfelt: I'd rather be explicit to the customer about it. @dominicbonnice is driving the effort from Pulitzer

packages/checkbox/src/CheckboxEditor.tsx
```diff
@@ -86,29 +104,44 @@ export function CheckboxEditor(props: CheckboxEditorProps) {
           setValue(newValues);
         };
 
+        const invalidValues = getInvalidValues(field, values, options);
+        const mergedOptions = [...options, ...invalidValues];
+
```
freakyfelt, line 109: Thoughts on something like below?

```
const { options, invalid } = partitionOptions(field, values, options)

return (
  <Form ...>
    {options.map(...)}
    {invalidOptions.map(...)}
  </Form>
)
```

freakyfelt: You might even be able to do the Checkbox render inside of the partitionOptions command (or whatever name you give it)

ChidinmaOrajiaku: This could work but I don't understand why this approach. Could you please shed some light?

ChidinmaOrajiaku: I think this approach is repetitive. What benefit does it provide?
````
options.map((item) => (<CheckboxField/>))
invalidOptions.map((item)) => (<CheckboxField/>)
```

freakyfelt: The main thing is that we don't need to munge the two lists together when iterating, instead we can just render the valid options and then render the invalidOptions right below

ChidinmaOrajiaku: Is there any disadvantage we are trying to avoid by not copying both lists into one array and rending them as such?

PR #365, opened 2020-08-18
feat: extract field component to default-field-editors package

djagya:
This PR introduces a new package `default-field-editors`. Its purpose is to render all supported field editors (check the `Field` component) in the same structure they are rendered in Contentful web-app (look at the `FieldWrapper` component).

Additionally:
- Exposes `FieldExtensionSDK` from `shared` package
- Imports `codemirrorImports`, which contains styles and extensions, in `json` package


djagya:
@suevalov @DanweDE can you please suggest a name for the new package? 

`field-editors-field` is clearly a bad one. I was thinking `field-editors-renderer` or `field-editors-base-field`, but both are meh.

DanweDE:
do we need to prefix with `field-editors`? Thinking...

djagya:
@DanweDE actually the prefix is different, fixed my previous message

DanweDE:
Yeah, I know (not that different :laughing:)

If we've got to use that prefix then `field-editor-default-editors`, `field-editor-builtin-editors`, `field-editor-auto-renderer`, `field-editor-by-widget-id` ... :shrug:

suevalov:
`default-field-editors` would be good I think

packages/field/src/Field.tsx
```diff
@@ -0,0 +1,229 @@
+import * as React from 'react';
+import { FieldExtensionSDK } from 'contentful-ui-extensions-sdk';
```
suevalov, line 2: I think this type is available via `field-editor-shared` and it would be to consume it that way. Ideally, `contentful-ui-extensions-sdk` shouldn't be a dependency of this package.

packages/field/src/Field.tsx
```diff
@@ -0,0 +1,229 @@
...
+import {
+  SingleEntryReferenceEditor,
+  MultipleEntryReferenceEditor,
+  SingleMediaEditor,
+  MultipleMediaEditor,
+} from '@contentful/field-editor-reference';
+import { RichTextEditor } from '@contentful/field-editor-rich-text';
+import { MarkdownEditor } from '@contentful/field-editor-markdown';
+
+export type WidgetType =
```
suevalov, line 27: Could you move this type to a separate file, please?

packages/field/src/Field.tsx
```diff
@@ -0,0 +1,229 @@
...
+        <DropdownEditor field={field} locales={locales} isInitiallyDisabled={isInitiallyDisabled} />
+      );
+    case 'entryLinkEditor': {
+      return (
+        <SingleEntryReferenceEditor
+          viewType="link"
+          sdk={sdk}
+          isInitiallyDisabled={isInitiallyDisabled}
+          hasCardEditActions={true}
+          parameters={{
```
suevalov, line 139: This should be `sdk.parameters` by default and should fall back to these default values in case if `sdk.parameters.instance` are missing

packages/field/src/FieldWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </HelpText>
+
+      {children}
+
+      <ValidationErrors
+        field={field}
+        space={sdk.space}
+        locales={sdk.locales}
+        getEntryURL={getEntryURL}
+      />
```
suevalov, line 39: `FieldWrapper` should be also responsible for rendering the help text which comes from `sdk.parameters.instance.helpText`.

In general, before publishing this package we need to make sure that we can use it as is in the main web application and that it has no regressions compared to what we currently have there.

djagya: @suevalov Help text might also come from `editorInterface.controls[fieldId].settings.helpText`, right?

suevalov: `editorInterface.controls[fieldId].settings` is actually  from where`sdk.parameters.instance` get their values

packages/default-field-editors/src/Field.tsx
```diff
@@ -0,0 +1,196 @@
...
+import type { WidgetType } from './types';
+
+type FieldProps = {
+  sdk: FieldExtensionSDK;
+  widgetId: WidgetType;
+  isInitiallyDisabled: boolean;
+  renderFieldEditor?: (
+    widgetId: WidgetType,
+    sdk: FieldExtensionSDK,
+    isInitiallyDisabled: boolean
```
DanweDE, line 36: I am wondering, wouldn't this be a much nicer api? should also save us some code in web creator as we just want to inject a couple of props in some editors. 

```ts
  renderFieldEditor?: (
    widgetId: WidgetType,
    DefaultEditorComponent: React.Component,
    defaultEditorProps: Object // So `return <DefaultEditorComponent {...defaultEditorProps} />` should have the same effect than `return false;`
  ) => JSX.Element | false;
```

On the other hand, with this approach one had to be super careful as `DefaultEditorComponent` might change from version to version and therefore potentially the `defaultEditorProps`

djagya: I see, good suggestion. There's no way to type the props however, so it would require checking the code/types of editor you want to render

djagya: Also this would not allow to render something completely custom? Or that's not the case to support here?

DanweDE: > you'd need to extract sdk from `defaultEditorProps`

`defaultEditorProps` might not have `sdk`. I was thinking since you instantiated `Field` which takes an `sdk` in the first place you should already have it at your disposal :shrug: But might not hurt to make it available anyways.

> There's no way to type the props however, so it would require checking the code/types of editor you want to render

Yes, that's what I meant by you would need to be "user careful". I think for that reason my idea actually doesn't make much sense, you would be like

```js
<Field {...someProps} renderFieldEditor={
  ({ widgetId, DefaultEditorComponent, defaultEditorProps }) => {
    if (sdk.field.type === FIELD_TYPE_I_WANT_TO_CUSTOMIZE) {
      if (DefaultEditorComponent === COMPONENT_I_WOULD_EXPECT_HERE) {
        return <DefaultEditorComponent {...defaultEditorProps} someProp={SOME_OVERRIDE_FOR_THIS_PROP} />
      } else {
        return <COMPONENT_I_WOULD_EXPECT_HERE {...ALL_THE_PROPS} />
      }
    }
  }
/>

```

**IN SHORT:** Let's forget about this, it was a terrible idea.

packages/default-field-editors/README.md
```diff
@@ -0,0 +1,11 @@
+# @contentful/field-editor-field
```
suevalov, line 1: This file needs to be updated: proper package name and a description

packages/default-field-editors/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@contentful/default-field-editors",
+  "version": "0.1.1",
```
suevalov, line 3: I think it makes sense to drop the number to `0.1.0`

packages/default-field-editors/src/FieldWrapper.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { styles } from './FieldWrapper.styles';
+
+type FieldWrapperProps = {
+  name: string;
+  sdk: FieldExtensionSDK;
+  getEntryURL: (entry: Entry) => string;
+  className?: string;
+  children: React.ReactNode;
+  renderHeading?: (name: string) => JSX.Element | null;
+  renderHelpText?: (helpText: string) => JSX.Element | null;
```
DanweDE, line 15: Maybe a bool prop to control whether or not to render the focus bar would make sense, [also see this comment](https://github.com/contentful/creator-app-mvp/pull/536/files#r474497472)

PR #375, opened 2020-08-24
feat: export default-field-editors EditorOptions type

djagya:
Makes `EditorOptions` type  from `default-field-editors` package available.


packages/default-field-editors/src/types.ts
```diff
@@ -24,3 +47,30 @@ export type WidgetType =
   | 'richTextEditor'
   | 'markdown'
   | string;
+
+export type EditorOptions = {
+  multipleLine: Partial<Parameters<typeof MultipleLineEditor>[0]>;
```
suevalov, line 52: Shouldn't every key be optional too?

PR #385, opened 2020-08-31
feat: 🎸 customize actions on reference editors

anho:
Add render prop to provide override for reference actions.

### DIFF
--- a/packages/reference/src/common/ReferenceEditor.tsx
+++ b/packages/reference/src/common/ReferenceEditor.tsx
@@ -3,6 +3,7 @@ import deepEqual from 'deep-equal';
 import { FieldConnector } from '@contentful/field-editor-shared';
 import { EntityProvider } from './EntityStore';
 import { ViewType, FieldExtensionSDK, Action, Entry, ContentType, ActionLabels } from '../types';
+import type { LinkActionsProps } from '../components/LinkActions/LinkActions';
 
 // TODO: Rename common base for reference/media editors to something neutral,
 //  e.g. `LinkEditor<T>`.
@@ -16,6 +17,7 @@ export interface ReferenceEditorProps {
   sdk: FieldExtensionSDK;
   viewType: ViewType;
   renderCustomCard?: (props: CustomEntryCardProps) => React.ReactElement | false;
+  renderCustomActions?: (props: CustomActionProps) => React.ReactElement;
   getEntityUrl?: (entryId: string) => string;
   onAction?: (action: Action) => void;
   actionLabels?: Partial<ActionLabels>;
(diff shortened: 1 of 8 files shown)


DanweDE (approved):

packages/reference/src/common/ReferenceEditor.tsx
```diff
@@ -16,6 +17,7 @@ export interface ReferenceEditorProps {
   sdk: FieldExtensionSDK;
   viewType: ViewType;
   renderCustomCard?: (props: CustomEntryCardProps) => React.ReactElement | false;
+  renderCustomActions?: (props: CustomActionProps) => React.ReactElement;
```
suevalov, line 20: Maybe it's worth to support the same `false` logic for consistency with `renderCustomCard` property?

suevalov (approved):

packages/reference/src/common/ReferenceEditor.tsx
anho: That is what I had in the beginning. After asking @DanweDE about the rationale behind the support for false for the card I didn't feel like we actually need it.

giotiskl (approved):

packages/reference/src/components/LinkActions/LinkActions.tsx
```diff
@@ -4,7 +4,7 @@ import * as styles from './styles';
 import { EntityType, ContentType, ActionLabels } from '../../types';
 import { CreateEntryLinkButton } from '../CreateEntryLinkButton/CreateEntryLinkButton';
 
-interface LinkActionsProps {
+export interface LinkActionsProps {
```
DanweDE, line 7: I think we should actually export this in the main `index.tsx` just like we expose the `CustomEntryCardProps` [here](https://github.com/contentful/field-editors/blob/fd1d9353d459f96f255430eba0f6b5217c30f973/packages/reference/src/index.tsx#L16)

As we're making this a public interface, we should also clean it up a bit. E.g. `multiple` is not consistent with most our public naming, e.g. `isMultipleLinks: boolean` would be a bit clearer.

A nice property to add (not sure if you need it though right now) would be one to indicate the current number of links to e.g. render different action buttons for an empty state link field editor.

anho: All of the things you mentioned are good points, I'd skip them for time being and once we see actual need for these "features" I'd happily come back and implement them.

DanweDE: The first two points are not features, I think you need to export `LinkActionsProps` from the main package to be able to use it. Renaming `multiple` should be done for consistency/clarity.

PR #387, opened 2020-09-01
fix: clarify multiple prop

anho:
Minor clean up to ease consumption of reference editor.

### DIFF
--- a/packages/reference/src/common/MultipleReferenceEditor.tsx
+++ b/packages/reference/src/common/MultipleReferenceEditor.tsx
@@ -105,11 +105,11 @@ class Editor extends React.Component<EditorProps, EditorState> {
         })}
         <LinkEntityActions
           entityType={this.props.entityType}
+          canLinkMultiple={true}
           allContentTypes={this.props.allContentTypes}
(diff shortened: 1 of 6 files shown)


DanweDE (approved):
:+1:

suevalov:
Should it be a breaking change?

anho:
> Should it be a breaking change?

Afaik we only rely internally so far, so I am bit hesitant to do that and opted-out of it. You have a strong opinion about it?

suevalov:
@anho This property was already exposed to a function that you've introduced earlier so if to be strict - it's a breaking change. 

On the other side, we haven't achieved a major version yet so we can be less strict and make `fix: ` PR that changes API of the previously released function.

It's definitely not a chore though.

DanweDE:
My 2 cents: Since the `CustomActionProps` wasn't previously exposed we don't necessarily have to declare it breaking. If that was the case, and `CustomActionProps` was already exposed on the top package level, then declaring a breaking change would be in order for sure. I also think it's a `fix:`.

suevalov (approved):

PR #390, opened 2020-09-03
feat: 🎸 add withCharValidation property to hide validation

suevalov:
To be able to hide validations when it's not necessary, for example - asset editor.

### DIFF
--- a/packages/single-line/src/SingleLineEditor.test.tsx
+++ b/packages/single-line/src/SingleLineEditor.test.tsx
@@ -202,4 +202,34 @@ describe('SingleLineEditor', () => {
     expect(getByText('0 characters')).toBeInTheDocument();
     expect(getByText('Requires at least 1000 characters')).toBeInTheDocument();
   });
+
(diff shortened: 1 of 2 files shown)


packages/single-line/src/SingleLineEditor.test.tsx
```diff
@@ -202,4 +202,34 @@ describe('SingleLineEditor', () => {
     expect(getByText('0 characters')).toBeInTheDocument();
     expect(getByText('Requires at least 1000 characters')).toBeInTheDocument();
   });
+
+  it('renders no validation message if withCharValidation is falsy', () => {
+    const [field] = createFakeFieldAPI((field) => {
+      return {
+        ...field,
+        type: 'Symbol',
```
freakyfelt, line 210: Heads up that Symbol by definition is max 256 characters. Text, on the other hand, would normally be unlimited in length unless the constraint is applied. Granted here it doesn't matter.

freakyfelt (approved):

PR #413, opened 2020-09-24
feat: 🎸 introduce experimental CombinedAssetLinkActions

DanweDE:
Introduces a new component `CombinedAssetLinkActions` that can be used in conjunction with reference and media editor `renderCustomCard` props. It introduces a new link actions layout which is planned to replace the current link actions in some future version of these editors.

**For references:**
new-entries

**For assets:**
media

### DIFF
--- a/cypress/integration/MultipleMediaEditor.spec.ts
+++ b/cypress/integration/MultipleMediaEditor.spec.ts
@@ -3,25 +3,49 @@ describe('Multiple Media Editor', () => {
     cy.visit('/media-multiple');
   });
 
-  const getDefaultWrapper = () => cy.findByTestId('multiple-media-editor-integration-test');
-  const getCustomActionsWrapper = () =>
-    cy.findByTestId('multiple-media-editor-custom-actions-integration-test');
-  const findCreateAndLinkBtn = () => getDefaultWrapper().findByTestId('linkEditor.createAndLink');
-  const findLinkExistingBtn = () => getDefaultWrapper().findByTestId('linkEditor.linkExisting');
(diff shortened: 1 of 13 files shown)


suevalov (commented):
I think it looks ok code-wise (even though `CreateEntryLinkButton.tsx` becomes too complicated) but I wonder if the change makes sense logically because it significantly de-prioritize selecting an existing content (especially with entries), and reusable content concept in general. I don't see an improvement UX wise and I think it makes sense to experiment with the concept in the new application without making any changes in the shared package.

Right now, it's possible to hide link/create button per field, is it possible to with a new component?

DanweDE:
> even though `CreateEntryLinkButton.tsx` becomes too complicated

Agreed, that's why I have those - hopefully - very clear TODOs there to clean-up once we get rid of one of the two designs (probably the old one unless users tell us otherwise)

> I wonder if the change makes sense logically because it significantly de-prioritize selecting an existing content (especially with entries), and reusable content concept in general.

I had similar concerns but I trust design on that one. Also, something to try out, which is why I'd really like to try it out in the web app behind a feature flag (we just won't get quick feedback in the new application).

> Right now, it's possible to hide link/create button per field, is it possible to with a new component?

Are you referring to `canLinkEntity` and `canCreateEntity` e.g. [here](https://github.com/contentful/field-editors/blob/c343dd4d5a3365b8e88026e746b2802cc1472294/packages/reference/src/components/LinkActions/LinkActions.tsx#L47) and

options
?

Good point, I've added support but then removed it again when simplifying some code. See latest commit for this working. No designs for it so took the liberty and not render a dropdown in this case, e.g.
existing-media-only
new-media-only

suevalov (approved):

PR #427, opened 2020-09-28
feat: 🎸 pass link actions props to a custom card renderer

djagya:
Extracts props of `LinkEntityActions` to a hook `useLinkActionsProps`, so the same props can be passed to `renderCustomCard`.

Would allow to rendered custom card actions utilising `onCreate`, `onLinkExisting`, `availableContentTypes`.

### DIFF
--- a/packages/reference/src/common/MultipleReferenceEditor.tsx
+++ b/packages/reference/src/common/MultipleReferenceEditor.tsx
@@ -2,9 +2,12 @@ import * as React from 'react';
 import arrayMove from 'array-move';
 import { ReferenceValue, EntityType, ContentType } from '../types';
 import { fromFieldValidations } from '../utils/fromFieldValidations';
-import { ReferenceEditor, ReferenceEditorProps } from './ReferenceEditor';
+import { CustomEntryCardProps, ReferenceEditor, ReferenceEditorProps } from './ReferenceEditor';
 import { LinkEntityActions } from '../components';
 import { SortEndHandler, SortStartHandler } from 'react-sortable-hoc';
+import { useLinkActionsProps } from '../components/LinkActions/LinkEntityActions';
+import { useCallback } from 'react';
+import { useEntityPermissions } from './useEntityPermissions';
 
 type ChildProps = {
   entityType: EntityType;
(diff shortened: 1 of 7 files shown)


packages/reference/src/components/LinkActions/helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+  validations: ReferenceValidations;
+}) {
+  const value = props.sdk.field.getValue();
+
+  const linkCount = Array.isArray(value) ? value.length : value ? 1 : 0;
+
+  // TODO: Why not always set `min: 1` by default? Does it make sense to enforce
+  //  user to select as many entities as the field's "min" requires? What if e.g.
+  // "min" is 4 and the user wants to insert 2 entities first, then create 2 new ones?
+  const min = Math.max((props.validations.numberOfLinks?.min || 1) - linkCount, 1);
```
suevalov, line 55: @DanweDE @djagya maybe it's time to clean this `min/max` complications? I think a lot of customers were complaining about it

DanweDE: I thought I recently updated the TODO but apparently, I didn't :thinking: 

I don't think we should be passing `1`. Instead, `sdk.dialogs.selectMultipleEntries` should always allow inserting an arbitrary number of links (while potentially informing you about limits by looking at these `min`/`max` values). What do you think?

suevalov: Yeah, it looks logical to me. I can incorporate this logic into the new entity editor, instead of disabling the button I would just inform the use about the restrictions.

packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -21,106 +24,66 @@ type EditorProps = ReferenceEditorProps &
...
-  render() {
-    const validations = fromFieldValidations([
-      ...this.props.sdk.field.validations,
-      ...(this.props.sdk.field.items?.validations ?? []),
-    ]);
+  const validations = fromFieldValidations([
+    ...props.sdk.field.validations,
+    ...(props.sdk.field.items?.validations ?? []),
+  ]);
+  const linkActionsProps = useLinkActionsProps({
```
DanweDE, line 61: Seems a bit expensive to do this for every link. Can't we just do it once and then reuse the same set of props on all links and for `renderCustomActions`?

packages/reference/src/components/LinkActions/helpers.ts
DanweDE: That would be grand :+1:

suevalov: Deal 👍

djagya: So what's the plan? Do I remove anything here now?

suevalov: No, you can leave it as it is. Maybe the comment is non needed anymore though

packages/reference/src/common/MultipleReferenceEditor.tsx
djagya: Actually I think that's how it works right now. `Editor` is rendered just once and then renders the children - items. So `useLinkActionsProps` is also supposed to be called once (and it also memoized the returned value). Then these props are passed to the custom card renderer and to `LinkEntityActions`.

suevalov (approved):
👍

PR #428, opened 2020-09-28
fix: 🐛 flaky location editor test

DanweDE:
Should result in green ci again.


suevalov (approved):
Let's merge it as `chore: ` to avoid version bump

PR #430, opened 2020-09-29
feat: 🎸 pass renderDefaultCard to renderCustomCard

IBratWurst:
Created.  renderDefaultCard function and exposed to renderCustomCard to allow customization of props

This is needed for the new app to differentiate non flattened entries between pages and non pages
https://contentful.atlassian.net/browse/PEN-1190

### DIFF
--- a/packages/reference/src/common/MultipleReferenceEditor.tsx
+++ b/packages/reference/src/common/MultipleReferenceEditor.tsx
@@ -68,9 +68,11 @@ function Editor(props: EditorProps) {
     onLink: onLink,
   });
   const customCardRenderer = useCallback(
-    (cardProps: CustomEntryCardProps) =>
-      props.renderCustomCard ? props.renderCustomCard(cardProps, linkActionsProps) : false,
-    [linkActionsProps]
+    (cardProps: CustomEntryCardProps, linkActionsProps, renderDefaultCard) =>
+      props.renderCustomCard
+        ? props.renderCustomCard(cardProps, linkActionsProps, renderDefaultCard)
+        : false,
+    []
   );
 
   return (
--- a/packages/reference/src/common/MultipleReferenceEditor.tsx
+++ b/packages/reference/src/common/MultipleReferenceEditor.tsx
@@ -68,11 +68,11 @@ function Editor(props: EditorProps) {
     onLink: onLink,
   });
   const customCardRenderer = useCallback(
-    (cardProps: CustomEntryCardProps, linkActionsProps, renderDefaultCard) =>
+    (cardProps: CustomEntryCardProps, _, renderDefaultCard) =>
       props.renderCustomCard
         ? props.renderCustomCard(cardProps, linkActionsProps, renderDefaultCard)
         : false,
-    []
+    [linkActionsProps]
   );
 
   return (
--- a/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
@@ -8,6 +8,7 @@ import {
   CustomActionProps,
   CustomEntryCardProps,
   ReferenceEditorProps,
+  DefaultCardRenderer,
 } from '../../common/ReferenceEditor';
 import get from 'lodash/get';
 import { WrappedEntryCardProps } from './WrappedEntryCard';
@@ -22,7 +23,7 @@ export type EntryCardReferenceEditorProps = ReferenceEditorProps & {
   renderCustomCard?: (
     props: CustomEntryCardProps,
     linkActionsProps: CustomActionProps,
-    renderDefaultCard: (props?: CustomEntryCardProps) => React.ReactElement
+    renderDefaultCard: DefaultCardRenderer
   ) => React.ReactElement | false;
   hasCardEditActions: boolean;
 };
--- a/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
@@ -21,7 +21,8 @@ export type EntryCardReferenceEditorProps = ReferenceEditorProps & {
   cardDragHandle?: React.ReactElement;
   renderCustomCard?: (
     props: CustomEntryCardProps,
-    linkActionsProps: CustomActionProps
+    linkActionsProps: CustomActionProps,
+    renderDefaultCard: (props?: CustomEntryCardProps) => React.ReactElement
   ) => React.ReactElement | false;
   hasCardEditActions: boolean;
 };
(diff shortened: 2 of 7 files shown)


packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -132,20 +133,34 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
       onEdit,
       onRemove,
     };
+
+    function renderDefaultCard(customProps?: Partial<WrappedEntryCardProps>) {
+      const builtinCardProps: WrappedEntryCardProps = {
+        ...sharedCardProps,
+        // eslint-disable-next-line react/prop-types
```
IBratWurst, line 140: at these line eslint is complaining:
```
  140:35  error  'hasCardEditActions' is missing in props validation                   react/prop-types
  141:25  error  'sdk' is missing in props validation                                  react/prop-types
  141:29  error  'sdk.space' is missing in props validation                            react/prop-types
  141:35  error  'sdk.space.getAsset' is missing in props validation                   react/prop-types
  142:42  error  'sdk' is missing in props validation                                  react/prop-types
  142:46  error  'sdk.space' is missing in props validation                            react/prop-types
  142:52  error  'sdk.space.getEntityScheduledActions' is missing in props validation  react/prop-types
```

Any idea how to prevent that?

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -21,7 +21,8 @@ export type EntryCardReferenceEditorProps = ReferenceEditorProps & {
   cardDragHandle?: React.ReactElement;
   renderCustomCard?: (
     props: CustomEntryCardProps,
-    linkActionsProps: CustomActionProps
+    linkActionsProps: CustomActionProps,
+    renderDefaultCard: (customProps?: Partial<WrappedEntryCardProps>) => React.ReactElement
```
DanweDE, line 25: Could we make this `renderDefaultCard: (props: CustomEntryCardProps) => React.ReactElement` ?

The idea would be that we don't need to expose a new type `WrappedEntryCardProps` from this package and you could easily customize like `(props, _, renderCustomCard) => renderDefaultCard({...props, ...myCustomOverwriteProps})`

IBratWurst: Can you have a look at it now? Is that what you suggested?

djagya (approved):
Looks good to me

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
Silhoue: `CustomEntryCardProps` doesn't include `hasCardEditActions` which is what we want to override in Web Creator.

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -132,20 +133,33 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
       onEdit,
       onRemove,
     };
+
+    const { hasCardEditActions, sdk } = props;
+
+    function renderDefaultCard(props?: CustomEntryCardProps) {
```
DanweDE, line 139: Will this really still work if `props` is omitted or should they be not optional?

But yes :+1: looking good like this. Consider to update one of the integration tests using the `return renderDefaultCard(...props)` instead of `return false`, e.g. [here](https://github.com/contentful/field-editors/blob/fd1d9353d459f96f255430eba0f6b5217c30f973/packages/reference/src/entries/MultipleEntryReferenceEditor.mdx) in a follow-up PR.

DanweDE (approved):

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
DanweDE: @Silhoue It includes an `onEdit?: () => void;`. I believe if that's not passed the card won't feature edit functionality. Or am I mistaken?

Silhoue: From what I see `onEdit` is created by `field-editors`, not passed: https://github.com/contentful/field-editors/blob/e62856c3c3ff89a82ccf9eb6af43f26695eea225/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx#L75-L88

DanweDE: Isn't that just the default value which will be part of `props` in `(props, _, renderCustomCard) => renderDefaultCard({...props, ...myCustomOverwriteProps})`?

I would advise making the objective work in one of the mdx spec files first (like rendering a default card there that comes without edit functionality)

Silhoue: Edit button renders when both are present https://github.com/contentful/field-editors/blob/e62856c3c3ff89a82ccf9eb6af43f26695eea225/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx#L137

So do you suggest to pass `hasCardEditActions=true` and then overwrite the `onEdit` to `undefined` for non-page cards?

Silhoue: The initial idea was to overwrite `hasCardEditActions` to `true` for page cards, which seems easier to understand.

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
Silhoue: It can be optional, spreading `undefined` doesn't alter the object.

packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -68,9 +68,11 @@ function Editor(props: EditorProps) {
...
   });
   const customCardRenderer = useCallback(
-    (cardProps: CustomEntryCardProps) =>
-      props.renderCustomCard ? props.renderCustomCard(cardProps, linkActionsProps) : false,
-    [linkActionsProps]
+    (cardProps: CustomEntryCardProps, linkActionsProps, renderDefaultCard) =>
+      props.renderCustomCard
+        ? props.renderCustomCard(cardProps, linkActionsProps, renderDefaultCard)
+        : false,
+    []
```
IBratWurst, line 75: @djagya We remove linkActionProps from here because they are passed to the customCardRenderer function everytime anyway. Is this ok with you?

djagya (approved):

packages/reference/src/common/MultipleReferenceEditor.tsx
djagya: Hmm, where do they get passed from?

djagya: If you mean https://github.com/contentful/field-editors/pull/430/files/3e343b599d9c8786e8e66cf93d32b4a5071f7b72#diff-f6ec592edf9c670a2aa37adfea260fb4R154, then it's just a placeholder so TS doesn't complain.

Actual `linkActionsProps` are generated in this place only.

djagya: So no, it should stay as it was before, taking `linkActionsProps` from here: https://github.com/contentful/field-editors/pull/430/files/3e343b599d9c8786e8e66cf93d32b4a5071f7b72#diff-1eb231122762c9ef75c2fc235137581bR61

djagya (changes requested):
`linkActionsProps` should be passed from `MultipleReferenceEditor`

packages/reference/src/common/MultipleReferenceEditor.tsx
IBratWurst: Ok, thank you. I updated the PR. If ok we can merge

djagya (approved):
Thanks, now it looks great

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
Silhoue: > So do you suggest to pass hasCardEditActions: true and then overwrite the onEdit to undefined for non-page cards?

FYI We went for the above solution as it seems to be the intention with field-editors. `hasCardEditActions` is only deciding on whether or not to show the dropdown, while `onEdit` indicates if the entity is generally openable or not and thus the pointy cursor problem you mentioned.

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
Silhoue: Testcases added in `SingleEntryReferenceEditor.mdx` and `MultipleEntryReferenceEditor.mdx`.

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
Silhoue: Jira ticket for the pointy cursor problem mentioned above: https://contentful.atlassian.net/browse/CX-509

PR #431, opened 2020-09-30
chore: 🤖 bump @contentful/forma-36-react-components

anho:
Bump Forma to utilise the size=auto in references.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "mochawesome-report-generator": "^5.1.0"
   },
   "resolutions": {
-    "@contentful/forma-36-react-components": "3.47.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -98,7 +98,7 @@
     "mochawesome-report-generator": "^5.1.0"
   },
   "resolutions": {
-    "@contentful/forma-36-react-components": "3.47.1",
+    "@contentful/forma-36-react-components": "3.57.1",
```
suevalov, line 101: I think that the minimum version should be also bumped in the `packages/reference/package.json` too.

suevalov (approved):

PR #437, opened 2020-10-07
feat: 🎸 do not show link actions on full reference/media fields

DanweDE:
We do not want to show the link actions on a reference or media field which is already full (meaning the field has a max number of links validation and that number is met or exceeded). This is true for both the current editor link design as well as the experimental CombinedLinkActions design.

This is actually regarded as a regression, although one already in place for such a long while that we can probably treat it as a feature at this point (hence the `feat` commit).

This also removes a lot of redundant mocks used by the media field examples.

### DIFF
--- a/cypress/integration/MultipleMediaEditor.spec.ts
+++ b/cypress/integration/MultipleMediaEditor.spec.ts
@@ -1,6 +1,8 @@
 describe('Multiple Media Editor', () => {
+  const openPage = () => cy.visit('/media-multiple');
+
   beforeEach(() => {
-    cy.visit('/media-multiple');
+    openPage();
   });
 
   const findCreateAndLinkBtn = (parent: Cypress.Chainable) =>
@@ -35,6 +37,8 @@ describe('Multiple Media Editor', () => {
 
   describe('custom actions injected actions dropdown', () => {
     beforeEach(() => {
+      cy.setFieldValidations([{ size: { max: 2 } }]);
+      openPage();
       cy.findByTestId('multiple-media-editor-custom-actions-integration-test').as('wrapper');
     });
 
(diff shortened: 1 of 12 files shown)


suevalov (changes requested):
Hmm, I'm not sure if simply hiding them is a good solution actually. Could we maybe change isDisabled property if we hit the limit and wrap the button with `Tooltip` component that explains why it is disabled?

I'm afraid that people might start asking "Where is my buttons?" and there will be support tickets if we simply hide them.

Could we discuss this change with Design and maybe PM?

DanweDE:
> I'm afraid that people might start asking "Where is my buttons?" and there will be support tickets if we simply hide them.

I would say we should follow the design patterns we have in disabled vs. hidden questions (I hope we have guiding principles here) and try out hidden buttons first if that's what we would usually do in such cases.

> Could we discuss this change with Design and maybe PM?

The change is coming form design (cc @joaoramos could you please verify that we want to hide the button, or would you prefer changing these to disabled buttons with a tooltip in this case?) 

> Hmm, I'm not sure if simply hiding them is a good solution actually. Could we maybe change isDisabled property if we hit the limit and wrap the button with Tooltip component that explains why it is disabled?

If we want to change this we'll need some design for this new disabled/tooltipped state in case of both the new and old design of these buttons in the new app (e.g. I imagine we wouldn't want them to take this much space in that case and hide the placeholder). Ideally by tomorrow as I am out today and next week, otherwise I'll have to stale this for a week or hand over to another dev.

Reminder new design:
style-update

Reminder old design:
link-buttons-old-design

joaoramos:
Those are valid concerns, @suevalov. If removing the buttons raises concerns with customers, we'll address that and course correct. Having disabled states and further UI to address these situations will potentially raise cognitive load which, unfortunately, is rather high already. Plus, they can still use the options menu to add references, where they will get a disabled action for it. I'd move forward with the PR.

suevalov (approved):

PR #442, opened 2020-10-13
feat(reference-editor): insert at index

cribbles:
Adds support for inserting a new created or linked entity or entities at a given index in the `MultipleReferencesEditor`.

By default we're currently adding such entities to the end of the list. This PR allows the options for clients to insert wherever they'd like, e.g., immediately after the element in a list, depending on how the card's rendered.

Note this is non-breaking as the default behavior remains the same.

Also adds `@types/deep-equal` as my compiler was complaining about that.

### DIFF
--- a/packages/reference/src/common/MultipleReferenceEditor.tsx
+++ b/packages/reference/src/common/MultipleReferenceEditor.tsx
@@ -24,6 +24,21 @@ type EditorProps = ReferenceEditorProps &
     children: (props: ReferenceEditorProps & ChildProps) => React.ReactElement;
   };
 
+function onLinkOrCreate(
+  setValue: ChildProps['setValue'],
+  entityType: ChildProps['entityType'],
+  items: ChildProps['items'],
+  ids: string[],
+  index?: number
+): void {
+  const links: ReferenceValue[] = ids.map((id) => ({
+    sys: { type: 'Link', linkType: entityType, id },
+  }));
+  const newItems = Array.from(items);
+  newItems.splice(index !== undefined ? index : items.length, 0, ...links);
+  setValue(newItems);
+}
+
 function Editor(props: EditorProps) {
   const { items, setValue, entityType } = props;
   const { canCreateEntity, canLinkEntity } = useEntityPermissions(props);
@@ -38,19 +53,12 @@ function Editor(props: EditorProps) {
   );
 
   const onCreate = useCallback(
-    (id: string) => {
-      setValue([...items, { sys: { type: 'Link', linkType: entityType, id } }]);
-    },
+    (id: string, index?: number) => onLinkOrCreate(setValue, entityType, items, [id], index),
     [setValue, items, entityType]
   );
 
   const onLink = useCallback(
-    (ids: string[]) => {
-      setValue([
-        ...items,
-        ...ids.map((id) => ({ sys: { type: 'Link', linkType: entityType, id } as const })),
-      ]);
-    },
+    (ids: string[], index?: number) => onLinkOrCreate(setValue, entityType, items, ids, index),
     [setValue, items, entityType]
   );
 
@@ -64,8 +72,8 @@ function Editor(props: EditorProps) {
     validations,
     canCreateEntity,
     canLinkEntity,
-    onCreate: onCreate,
-    onLink: onLink,
+    onCreate,
+    onLink,
   });
   const customCardRenderer = useCallback(
     (cardProps: CustomEntryCardProps, _, renderDefaultCard) =>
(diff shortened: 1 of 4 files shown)


packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -64,29 +64,38 @@ export function useLinkActionsProps(props: LinkEntityActionsProps): LinkActionsP
...
-    props.onAction &&
-      props.onAction({ type: 'select_and_link', entity: entityType, entityData: entity });
-  }, [sdk, entityType, props.onLink, props.onAction]);
+  const onLinkExisting = React.useCallback(
+    async (index?: number) => {
+      const entity = await selectSingleEntity({ sdk, entityType, validations });
+      if (!entity) {
+        return;
+      }
+      props.onLink([entity.sys.id], index);
```
cribbles, line 73: This diff is ugly because of the linter but the only thing to take note of is we're now parameterizing the optional index and delegating it to `onLink`.

packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -64,29 +64,38 @@ export function useLinkActionsProps(props: LinkEntityActionsProps): LinkActionsP
...
-      return;
-    }
-    props.onLink(entities.map((item) => item.sys.id));
+      if (!entities || entities.length === 0) {
+        return;
+      }
+      props.onLink(
+        entities.map((item) => item.sys.id),
+        index
+      );
```
cribbles, line 90: Same here.

packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -24,6 +24,21 @@ type EditorProps = ReferenceEditorProps &
     children: (props: ReferenceEditorProps & ChildProps) => React.ReactElement;
   };
 
+function onLinkOrCreate(
+  setValue: ChildProps['setValue'],
+  entityType: ChildProps['entityType'],
+  items: ChildProps['items'],
+  ids: string[],
+  index = -1
```
djagya, line 32: Why not `index?: number` instead? Might be also confusing as I saw it being used as `insert before the first from end`

djagya (approved):

packages/reference/src/common/MultipleReferenceEditor.tsx
cribbles: Sure - I'm happy with rewriting. My thought was to echo the behavior of primitives like `indexOf`, but I can see why this would be confusing or conflict with the ability to do Ruby-like backward index scanning at some point in the future.

cribbles: Fixed in https://github.com/contentful/field-editors/pull/442/commits/c166bdc923aef114f88c9fdeb0b63a2bd25302e7.

djagya: Yeah, I agree with using `-1` as a search result, but for inserting imho it becomes a different meaning.

djagya (approved):

packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -14,8 +14,8 @@ type LinkEntityActionsProps = {
   canCreateEntity: boolean;
   canLinkEntity: boolean;
   validations: ReferenceValidations;
-  onCreate: (id: string) => void;
-  onLink: (ids: string[]) => void;
+  onCreate: ((id: string) => void) | ((id: string, index?: number) => void);
+  onLink: ((ids: string[]) => void) | ((ids: string[], index?: number) => void);
```
suevalov, line 18: Do we need an overload type here? 

```suggestion
  onLink: (ids: string[], index?: number) => void;
```

should be okay, or not?

suevalov (commented):
One small nit

packages/reference/src/components/LinkActions/LinkEntityActions.tsx
cribbles: I think I saw an error come up around this when this changeset was in development, but I changed it locally and I don't see it now, so as long as CI is okay with it we can do it this way :)

cribbles: Updated in https://github.com/contentful/field-editors/pull/442/commits/5131e7501b03d3f6964b40738161125da639f2f9.

suevalov (approved):

PR #453, opened 2020-10-22
feat: 🎸 auto position combined link actions dropdown

DanweDE:
The new combined reference link action's dropdown will be positioned automatically, if necessary, above the trigger button (in most cases, if there's enough space, it will still be shown below). This should prevent the menu from being cut off when e.g. positioned at the end of a page.

### DIFF
--- a/packages/reference/src/components/LinkActions/CombinedLinkActions.tsx
+++ b/packages/reference/src/components/LinkActions/CombinedLinkActions.tsx
@@ -42,6 +42,10 @@ function CombinedEntryLinkActions(props: LinkActionsProps) {
         contentTypes={props.contentTypes}
         hasPlusIcon={true}
         useExperimentalStyles={true}
+        dropdownSettings={{
+          isAutoalignmentEnabled: true,
+          position: 'bottom-left',
+        }}
         onSelect={(contentTypeId) => {
           return contentTypeId ? props.onCreate(contentTypeId) : Promise.resolve();
         }}


m10l:
@DanweDE I'd hold off on this - there's a PR open to refactor the dropdown component to use a more stable positioning package (popper.js) - https://github.com/contentful/forma-36/pull/597

DanweDE:
@m10l this PR shouldn't conflict with that work, but without it we wouldn't benefit from the positioning in contentful/forma-36#597 as auto-positioning is currently disabled on this button. Or am I missing something?

DanweDE:
Since the `isAutoalignmentEnabled` prop is celebrating its comeback in contentful/forma-36#597 I think we can just merge this PR @denkristoffer @m10l 

> @burakukula @DanweDE I think I managed to replicate disabling positioning with the isAutoAlignmentEnabled prop. I have un-deprecated it, so we can get this in without breaking changes. I think it's ready.

denkristoffer:
@DanweDE Looks right, but the values passed to the dropdown are also the default values, so are you sure this commit changes anything?

DanweDE:
> but the values passed to the dropdown are also the default values

@denkristoffer Pretty sure this is necessary, see [this line](https://github.com/contentful/field-editors/blob/39b5bfa32336fdd87848d0a937fe70c2a13ca5ef/packages/reference/src/components/CreateEntryLinkButton/CreateEntryMenuTrigger.tsx#L83) where the default is `false`, not the actual dropdown default which would be `true`.

suevalov:
@denkristoffer @DanweDE should it be closed by https://github.com/contentful/field-editors/pull/466?

denkristoffer:
@suevalov No it looks like the "Create entry" dropdown disables automatic positioning by default, so this is a good change.

denkristoffer (approved):

PR #462, opened 2020-10-27
chore: bump forma-36-react-components to 3.63.5

### LINKED ISSUE #188: Asset options are going out of view port in SingleMediaEditor
I used the `SingleMediaEditor` in an App field extension and it works great. One thing I noticed is that the edit options (`edit` / `remove`) behind the three dots are going out of the iframe viewport.

It would be great to implement the behavior of the `SingleEntryReferenceEditor`. This editor seems to take of that by placing the tooltip in the visible area. :) Thanks.

Screenshot 2020-05-22 at 10 49 53

☝️ Example shows that the reference editor is 100% functional, whereas the asset editor places the tool tip out of reach.

### LINKED ISSUE #415: Multi-reference editor 'Create new entry and link' dropdown won't work in an app
To reproduce this:

1. Clone https://github.com/kamsar/ctfl-multi-reference-embed-bug ([specific repro code location](https://github.com/kamsar/ctfl-multi-reference-embed-bug/blob/master/src/components/Field.tsx))
2. `npm run start`
3. In Contentful, add a private app pointed at `http://localhost:1338`, with a Entry field editor for Entry reference, list
4. Assign the app's field editor to any Entry reference, list field. Ensure that more than one content type is allowed to be linked.
5. Edit an entry with the custom editor. It will render correctly and work as a duplicate of the default multiple reference editor _until_ you click `Create new entry and link`. Then the dropdown for content type selection becomes hidden and can't be clicked.

image

The same view, but with `sdk.window.startAutoResizer()` replaced with `sdk.window.updateHeight(800)` to force a very tall iframe:

image

There are two issues here (and they're probably closely related):

1. The extension SDK's `autoResizer` is [not picking up the opening of the dropdown and resizing the iframe](https://github.com/kamsar/ctfl-multi-reference-embed-bug/blob/cfbd90b5549de8b55ca04f60e2489e27396a35d8/src/components/Field.tsx#L10)
2. This seems to be because the technique used to open the dropdown is also not updating any of the usual browser height detection properties, such as `document.body.scrollHeight` or `document.documentElement.getBoundingClientRect()` all of which remain constant within the iframe whether the dropdown is open or not.

denkristoffer:
3.63.5 has the new dropdown positioning.

Resolves #188
Resolves #415

### DIFF
--- a/extensions/entry-extension-collapsible/package.json
+++ b/extensions/entry-extension-collapsible/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "@contentful/field-editor-shared": "^0.13.0",
     "@contentful/forma-36-fcss": "^0.2.1",
-    "@contentful/forma-36-react-components": "^3.45.1",
+    "@contentful/forma-36-react-components": "^3.63.5",
     "@contentful/forma-36-tokens": "^0.7.2",
     "contentful-ui-extensions-sdk": "^3.18.0",
     "emotion": "^10.0.17",
--- a/extensions/markdown-extension/package.json
+++ b/extensions/markdown-extension/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@contentful/forma-36-fcss": "^0.2.1",
-    "@contentful/forma-36-react-components": "^3.45.1",
+    "@contentful/forma-36-react-components": "^3.63.5",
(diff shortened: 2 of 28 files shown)


taye (approved):

suevalov (approved):

denkristoffer:
Superseded by #466

PR #468, opened 2020-11-02
feat: "move to top", "move to bottom" reference actions

djagya:
Adds buttons "Move to top" and "Move to bottom" to the multiple reference editor cards. The buttons are hidden when the action is not possible due to the card position.


suevalov (approved):
I think our customers will love this 💚 

Once shipped we'll need add this to https://www.contentful.com/developers/changelog/

m10l (approved):

PR #469, opened 2020-11-02
feat: support custom card rendered in media ref editors

djagya:
Adds support for custom cards rendering in media reference editors. It can be achieved similarly to the entry reference editor - by using `renderCustomCard`.

**Breaking change:** 
- `CustomEntryCardProp` -> `CustomEntityCardProps`
- `CustomEntryCardProp.entry: Entry` -> `CustomEntityCardProps.entity: Entry | Asset`
- `CustomEntryCardProp.entryUrl` -> `CustomEntityCardProps.entityUrl`

Both entry and asset editors custom card properties `CustomEntityCardProps` now contain an entry or an asset in the unified `entity` field and its URL in the `entityUrl` field.


packages/reference/src/common/customCardTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+  isDisabled: boolean;
+  size: 'default' | 'small';
+  cardDragHandle?: React.ReactElement;
+  onEdit?: () => void;
+  onRemove?: () => void;
+  onMoveTop?: () => void;
+  onMoveBottom?: () => void;
+  // TODO: specific types should be used instead of these props, but for that the ReferenceEditor props should accept a generic
+  // Entry editor
+  entry?: Entry;
```
DanweDE, line 28: Would it be difficult to address above TODO and just make this `entity` instead of `entry?` and `asset?` ?

djagya: I was trying to do that, but it all requires making the [`ReferenceEditorProps`](https://github.com/contentful/field-editors/blob/master/packages/reference/src/common/ReferenceEditor.tsx#L11-L35) generic. This will break many other things relying on this type. 

I also tried to make `ReferenceEditorProps.renderCustomCard<CustomEntryCardProps | CustomAssetCardProps>` and make this function type more specific in each of the editors types: `SingleEntryReferenceEditor`, `MultipleEntryReferenceEditor`, `MultipleMediaEditor` and `SingleMediaEditor`. But it still conflicts because of hierarchy and composing of the multiple components.

So my answer is that it's not easy unless I miss an obvious and simple approach.

packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
```diff
@@ -73,23 +77,42 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
...
+      };
+
+      // LinkActionsProps are injected higher SingleReferenceEditor/MultipleReferenceEditor
+      const renderedCustomCard = props.renderCustomCard(
+        customProps,
+        {} as LinkActionsProps,
+        renderDefaultCard
+      );
+      // Only `false` indicates to render the original card. E.g. `null` would result in no card.
+      if (renderedCustomCard !== false) {
```
IBratWurst, line 111: What is the use case of having a renderCustomCard that returns false? If we want the default card, shouldn't we not provide a renderCustomCard at all?

djagya: We better ask @DanweDE. But I think the idea is that you can decide inside the custom card renderer where you have access to all the computed props, specifically loaded `asset`. For example you could say that an asset of "published" status should be rendered as default card, otherwise as custom.

DanweDE: Yes, that's the idea here :point_up: you can basically make the decision whether or not to render the custom card dynamically at "run time" and for each card individually rather than just during configuration of the component.

packages/reference/src/common/customCardTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+  entryUrl?: string;
+  contentType?: ContentType;
+  // Asset editor
+  asset?: Asset;
+  href?: string;
```
DanweDE, line 33: This seems to be the same as `entryUrl` above but for assets? Please rename to `assetUrl`, but basically the same as above, `entityUrl` would be great :laughing:

packages/reference/src/common/customCardTypes.ts
DanweDE: >  it all requires making the ReferenceEditorProps generic. 

They seem pretty generic already, with the exception of `renderCustomCard?.props: CustomEntryCardProps`. And that's because `CustomEntryCardProps` was only relevant to entry link (reference) editors so far.

So it all comes back to [this TODO](https://github.com/contentful/field-editors/blob/fcae1d5dc1423df1b4c4b346cdeee3ed9f2e6340/packages/reference/src/common/ReferenceEditor.tsx#L39-L41):
https://github.com/contentful/field-editors/blob/fcae1d5dc1423df1b4c4b346cdeee3ed9f2e6340/packages/reference/src/common/ReferenceEditor.tsx#L39-L43

But I guess that's nothing new for you, just trying to get an overview here.

> I also tried to make ReferenceEditorProps.renderCustomCard<CustomEntryCardProps | CustomAssetCardProps> and make this function type more specific in each of the editors types

Feels to me like a generic `CustomEntityCardProps` (or `CustomCardProps`) would be enough for now. But maybe I am missing something.

> This will break many other things relying on this type.
I think a major version bump and telling users to adjust to the new interfaces would be preferable over having less clean interfaces that might grow even more into this direction with further changes.

It's hard for me to keep an overview, but wouldn't we need to introduce a major version bump anyways considering your change of `CustomEntryCardProps` to `CustomCardProps`?

DanweDE (approved):
I think it's looking good now :+1: Don't forget to release as a major version, considering the breaking interface changes.

PR #474, opened 2020-11-06
chore(deps): update Forma 36 dependencies

denkristoffer:


suevalov (changes requested):
@denkristoffer We don't want to pin exact versions of dependencies to avoid having multiple versions in applications bundles.

A rule of thumb: always use `^` in libraries except a really rare cases and when you don't update the dependency often.

denkristoffer:
@suevalov Yeah, I don't know why I thought of field-editors as an app, that was a mistake 🙂 Updated!

suevalov (approved):

PR #530, opened 2020-12-16
feat: use access api to hide inaccessible content

anho:


packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -92,6 +85,11 @@ function Editor(props: EditorProps) {
     [linkActionsProps]
   );
 
+  const testProps = {
```
z0al, line 88: Is the naming here intentional ? :)

packages/reference/src/common/useContentTypePermissions.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    async function checkContentTypeAccess() {
+      const creatable = await getContentTypes('create');
+      const readable = await getContentTypes('read');
+      setCreatableContentTypes(creatable);
+      setReadableContentTypes(readable);
+    }
+
+    void checkContentTypeAccess();
+  }, [availableContentTypes]);
```
z0al, line 64: I think the `useEffect` dependency list needs to be updated, oder?

z0al (commented):
Took a quick look. Looking good. Will review it again after the power hour

packages/reference/src/common/MultipleReferenceEditor.tsx
anho: 🦅 👀

suevalov (commented):
Looks good, did you test it with a local copy of the main web app?

packages/reference/src/common/useContentTypePermissions.ts
anho: It is fine like this.

anho:
> Looks good, did you test it with a local copy of the main web app?

Quick sanity check yes. Will integrate it back after an actual testing session.

suevalov (approved):

packages/reference/src/common/useContentTypePermissions.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    if (props.validations.contentTypes) {
+      return props.allContentTypes.filter((ct) =>
+        props.validations.contentTypes?.includes(ct.sys.id)
+      );
+    }
+
+    return props.allContentTypes;
+  }, [props.validations.contentTypes, props.entityType]);
```
djagya, line 46: Did you intentionally leave `allContentTypes` out of deps array?

giotiskl (approved):

djagya (approved):
Nice. The `.mdx` files used by docz don't need to be updated, right? (like `SingleEntryReferenceEditor.mdx`)

packages/reference/src/common/useContentTypePermissions.ts
anho: Nope, I see a pattern where I miss out on updating effect dependencies 😄

anho:
@djagya 
> Nice. The .mdx files used by docz don't need to be updated, right? (like SingleEntryReferenceEditor.mdx)

No need to update, there are no external changes.

PR #550, opened 2021-01-11
chore: [] Upgrade uie sdk

Jwhiles:
Upgrade UIE-sdk and fix some type errors that arose from the upgrade. In some cases I've chosen to use `BaseExtensionSDK & SharedEditorSDK` as this seems to best reflect what these components expect.


andipaetzold:
Isn't that exactly `FieldExtensionSDK`?

anho:
> Isn't that exactly `FieldExtensionSDK`?

Yep, would state the same. @Jwhiles you feel like you wanna tackle it as part of this PR?

suevalov (approved):

Jwhiles:
@andipaetzold @anho I was trying to allow the possibility for the possibility of those components being used for sidebar, entry, or field location - which is why I didn't use the `FieldExtensionSDK`. Happy to change it over though :)

PR #564, opened 2021-01-19
feat(asset reference): add support for scheduled icon [PUL-1206]

Spring3:
Next steps:
- bump the dependency in the web app
- bump the dependency in entity selector

Here I did the same thing we do already for entry references


djagya (approved):
Looks good to me

suevalov (approved):
Scheduled assets are coming 😎

Spring3:
@suevalov :) in fact they are already there since maybe September last year. But this is just some follow up to add indicators.

PR #580, opened 2021-02-02
feat(reference): adjust permissions

marcolink:
The check for permissions on read entity are too pessimistic.
There are situation where we can have enitites to reference, but we don't offer the "select content to link" button.

### DIFF
--- a/packages/reference/src/common/useEditorPermissions.ts
+++ b/packages/reference/src/common/useEditorPermissions.ts
@@ -20,7 +20,7 @@ export function useEditorPermissions(props: EditorPermissionsProps) {
   const {
     creatableContentTypes,
     readableContentTypes,
-    availableContentTypes,
+    availableContentTypes
   } = useContentTypePermissions({ ...props, validations });
   const { canOnEntity } = useAccessApi(sdk.access);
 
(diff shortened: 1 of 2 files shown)


packages/reference/src/common/useEditorPermissions.ts
```diff
@@ -50,25 +50,29 @@ export function useEditorPermissions(props: EditorPermissionsProps) {
...
+      if (props.allContentTypes?.length) {
+        if (entityType === 'Asset') {
+          const canRead = await canOnEntity('read', 'Asset');
+          setCanLinkEntity(canRead);
+        }
+        if (entityType === 'Entry') {
+          setCanLinkEntity(readableContentTypes.length > 0);
+        }
+      } else {
+        setCanCreateEntity(true)
```
anho, line 62: You may want to set `setCanLinkEntity(true)` here as well, don't you?

marcolink: @anho uh, good catch, It should actually only be `setCanLinkEntity`  ;)

Vinz93 (approved):
LGTM

damienxy (approved):

anho (approved):

PR #581, opened 2021-02-03
feat(reference): change text links to button

denkristoffer:


suevalov:
Screenshot 2021-02-04 at 11 31 47

can we also change "link existing" to be a button with the same style?

suevalov:
> Can I remove the existing code that lets us fall back to the previous reference field editor?

After one more thought, I don't think we need to delete it now. I still think that merging two buttons into one worsened UX and wasn't received well by the customers. We might want to revaluate this and fallback to the previous implementation.

packages/reference/src/components/LinkActions/CombinedLinkActions.tsx
```diff
@@ -97,32 +96,34 @@ function CombinedAssetLinkActions(props: LinkActionsProps) {
   if (!props.canLinkEntity || !props.canCreateEntity) {
     if (props.canLinkEntity) {
       return (
-        <TextLink
+        <Button
```
suevalov, line 99: I personally like that the existing one has primary color, so reusing the content is prioritized, but was it approved by design?

denkristoffer: Yes. I personally think it looks out of place in the editor when the whole button is primary.

denkristoffer:
> After one more thought, I don't think we need to delete it now. I still think that merging two buttons into one worsened UX and wasn't received well by the customers. We might want to revaluate this and fallback to the previous implementation.

Okay but at this point I still think we will want to maybe split the button into two rather than go back to text links.

packages/reference/src/components/LinkActions/CombinedLinkActions.tsx
suevalov: "when the whole button is primary" - what do you mean here?

denkristoffer: With the button the primary colour is quite overpowering and draws your attention since the buttons are so big. Even with `primary` I don't think `TextLink` draws your attention that much, so while `primary` works well for `TextLink` I don't think that's the case with buttons here.

suevalov: so, should we use muted for both of them maybe?

suevalov (approved):

PR #583, opened 2021-02-04
fix: 🐛 proper resolving default fallback for references

suevalov:


packages/default-field-editors/src/getDefaultWidgetId.ts
```diff
@@ -32,6 +28,21 @@ export function getDefaultWidgetId(
...
+  if (fieldType === 'Array') {
+    if (field.items?.linkType === 'Asset') {
+      return 'assetLinksEditor';
+    }
+    return 'entryLinksEditor';
+  }
+
+  if (fieldType === 'Link') {
+    // @ts-expect-error
+    if (field.linkType === 'Asset') {
```
suevalov, line 40: we don't have this in app-sdk types but it's actually there and there's no other way to check linkType

Jwhiles (approved):

PR #584, opened 2021-02-04
fix(tags): Editor permissions handles entities with metadata

Vinz93:
The access API is not working properly when we have permissions that have tags in them, the issue seems to be solved by  adding property `metadata.tags` to the dump entry generated in **makeEntryFromType**

### DIFF
--- a/packages/reference/src/common/useAccessApi.ts
+++ b/packages/reference/src/common/useAccessApi.ts
@@ -14,6 +14,9 @@ function makeEntryFromType(contentTypeId: string) {
         sys: { id: contentTypeId },
       },
     },
+    metadata: {
+      tags: [],
+    },
     fields: {},
   } as Entry;
 }
--- a/packages/reference/src/common/useEditorPermissions.spec.ts
+++ b/packages/reference/src/common/useEditorPermissions.spec.ts
@@ -63,10 +63,11 @@ describe('useEditorPermissions', () => {
       expect(sdk.access.can).not.toHaveBeenCalledWith();
     });
 
-    it(`defaults link entry action visibility to true`, async () => {
-      const { result } = await renderEditorPermissions({ entityType: 'Entry' });
+    it(`checks basic access`, async () => {
+      const { sdk } = await renderEditorPermissions({ entityType: 'Asset' });
 
-      expect(result.current.canLinkEntity).toBeTruthy();
+      expect(sdk.access.can).toHaveBeenCalledWith('create', 'Asset');
+      expect(sdk.access.can).toHaveBeenCalledWith('read', 'Asset');
     });
 
     it(`defaults link asset action visibility to true`, async () => {
--- a/packages/reference/src/common/useEditorPermissions.ts
+++ b/packages/reference/src/common/useEditorPermissions.ts
@@ -20,7 +20,7 @@ export function useEditorPermissions(props: EditorPermissionsProps) {
   const {
     creatableContentTypes,
     readableContentTypes,
-    availableContentTypes
+    availableContentTypes,
   } = useContentTypePermissions({ ...props, validations });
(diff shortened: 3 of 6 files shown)


packages/reference/src/common/useAccessApi.ts
```diff
@@ -7,15 +7,18 @@ import type {
...
       contentType: {
         sys: { id: contentTypeId },
       },
     },
+    metadata: {
+      tags: [],
+    },
     fields: {},
-  } as Entry;
+  } as unknown) as Entry;
```
marcolink, line 21: I wonder if we should add `metadata: {tags: []}` to the [Entry](https://github.com/contentful/field-editors/blob/master/packages/_shared/src/typesEntity.ts) type?

marcolink (approved):
LGTM

packages/reference/src/common/useEditorPermissions.ts
```diff
@@ -50,29 +50,25 @@ export function useEditorPermissions(props: EditorPermissionsProps) {
...
+        setCanLinkEntity(canRead);
+      }
+      if (entityType === 'Entry') {
+        setCanLinkEntity(readableContentTypes.length > 0);
       }
     }
 
     void checkLinkAccess();
-  }, [entityType, parameters.instance, readableContentTypes, props.allContentTypes]);
+  }, [entityType, parameters.instance, readableContentTypes]);
```
suevalov, line 63: Looks like we don't need props.allContentTypes in this component anymore, right?

Vinz93: You are right! good point, then I'll delete it from the **EditorPermissionsProps** interface and I'll make sure that we are not passing it in the  props anymore 👍

packages/reference/src/common/useEditorPermissions.spec.ts
```diff
@@ -63,10 +63,11 @@ describe('useEditorPermissions', () => {
       expect(sdk.access.can).not.toHaveBeenCalledWith();
     });
 
-    it(`defaults link entry action visibility to true`, async () => {
-      const { result } = await renderEditorPermissions({ entityType: 'Entry' });
+    it(`checks basic access`, async () => {
```
anho, line 66: I don't get the intention behind that change, can you briefly explain?

packages/reference/src/components/LinkActions/helpers.ts
```diff
@@ -46,7 +46,7 @@ export async function selectMultipleEntities(props: {
   sdk: FieldExtensionSDK;
   entityType: EntityType;
   editorPermissions: EditorPermissions;
-}) {
+}): Promise<(Entry | Asset)[] | null> {
```
anho, line 49: Is it necessary to explicitly type here?

packages/reference/src/common/useEditorPermissions.spec.ts
Vinz93: I am reverting the changes done in this [pr](https://github.com/contentful/field-editors/pull/580) since it was not solving our problem

packages/reference/src/common/useEditorPermissions.ts
Vinz93: I tried to delete props.allContentType from the EditorPermissionsProps, and I found that we are passing them to **useContentTypePermissions** hook

packages/reference/src/components/LinkActions/helpers.ts
Vinz93: Indeed no, I just removed it, I needed to add this explicit return because when I changed the shape of Entry by adding the `metadata.tags[]` property the type inferral was not working as expected, but I updated the Asset's interface after, and that fixed the type inferral issue that I had before 👍

anho (approved):

PR #585, opened 2021-02-05
fix rich text command palette and embed dropdown icons

denkristoffer:
* fix(rich-text): add z-index token to rich text command palette
* fix(embedded-entry-inline): use Flex component to align icons
* fix(embedded-entity-block): use Flex component to align icons


suevalov (approved):
💯

PR #601, opened 2021-02-15
fix: 🐛 handle lack of proper acccess checks in the webapp

anho:
The access check on content-types is slightly buggy in the web app and needs to be addressed separately.


suevalov (approved):
So, basically we fallback to `allow` if `canPerformActionOnEntryOfType` is missing?

anho:
@suevalov correct.

PR #610, opened 2021-02-23
chore: migrate examples from extensions to apps

VaguelySerious:
Use create-contentful-app as the base for all widget examples, instead of using create-contentful-extension.
Rename them and make sure that the correct app-framework related information is linked.

Renaming everything from "extensions" to "apps" will be done in a separate PR, as otherwise it would be extremely hard to review.

- [x] Merge this first: https://github.com/contentful/field-editors/pull/611


VaguelySerious:
I made a README change in `package:_shared`. I wonder if that'll cause a version bump through github actions?

z0al:
> I made a README change in `package:_shared`. I wonder if that'll cause a version bump through github actions?

`chore` commits should not result in a version bump.

PR #612, opened 2021-02-23
feat: default-field-editors doc improvements

djagya:
Added few more examples, explained the `widgetId` and made the `getEntryURL` prop optional.

### DIFF
--- a/packages/default-field-editors/src/FieldWrapper.tsx
+++ b/packages/default-field-editors/src/FieldWrapper.tsx
@@ -8,6 +8,9 @@ import { styles } from './FieldWrapper.styles';
 type FieldWrapperProps = {
   name: string;
   sdk: FieldExtensionSDK;
+  /**
+   * Generates a link to another entry with the same value when a "non unique" validation error occurs
+   */
   getEntryURL?: (entry: Entry) => string;
   className?: string;
   showFocusBar?: boolean;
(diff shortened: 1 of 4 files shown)


packages/default-field-editors/src/FieldWrapper.tsx
```diff
@@ -8,24 +8,26 @@ import { styles } from './FieldWrapper.styles';
...
   const {
     name,
     sdk,
-    getEntryURL,
     className,
     children,
     renderHeading,
     renderHelpText,
     showFocusBar = true,
+    getEntryURL = defaultGetEntryUrl,
```
stefanjudis, line 30: Nit: I still don't understand why/when I should use `getEntryUrl` when using an editor.

stefanjudis (approved):
Tiny nit/question, but apart from that works for me.

packages/default-field-editors/src/FieldWrapper.tsx
djagya: `getEntryUrl` is required for the "this value is non-unique" validation error, where we fetch a list of duplicate entries and show them in a message together with links to them.

In theory `getEntryUrl` should be a method of SDK, so users won't need to specify one unless they want a custom experience. So I'll add a comment to this prop and also will specify a more proper `defaultGetEntryUrl`.

PR #467, opened 2020-11-02
chore(deps): upgrade Forma 36 to 3.64.1

denkristoffer:
fix: use click instead of mousedown handlers
build: allow node 12.8.1

PR #622, opened 2021-03-04
chore(extensions): fix build errors [EXT-2496]

FLoppix:
The build process for the extensions / examples was broken since November due to a dependency update that included a forma version bump (https://github.com/contentful/field-editors/pull/467).  

The error resulted in a `Error: Duplicate plugin/preset detected.`

My approach was to remove duplicated `babel` dependencies that we already have declared in the root `package.json` file. I also added a build step that checks if the extensions are able to build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,7 @@ jobs:
       - run: yarn lint
       - run: yarn bootstrap 
       - run: yarn build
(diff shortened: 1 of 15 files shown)


Jwhiles (approved):
Nice! Building the extensions as part of the CI check is a great improvement

suevalov (approved):

PR #683, opened 2021-04-12
fix(reference): multi-reference create and link actions always with index

taye:


taye:
Actually, this doesn't fix anything at all 🙈

packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -30,13 +30,13 @@ function onLinkOrCreate(
   entityType: ChildProps['entityType'],
   items: ChildProps['items'],
   ids: string[],
-  index?: number
+  index = items.length,
```
z0al, line 33: Shouldn't the index be `length -1`?

taye: this is called before `items` is updated, so this works

packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -90,7 +90,9 @@ function Editor(props: EditorProps) {
     editorPermissions,
     onCreate,
     onLink,
+    itemsLength: items.length,
```
IBratWurst, line 93: Couldn't `itemsLength` be read in the `onLink` and `onCreate` handlers above instead of pass it down to the children?
Something like 
```
const onCreate = useCallback(
    (id: string, index?: number = itemsLength) => onLinkOrCreate(setValue, entityType, items, [id], index),
    [setValue, items, entityType]
  );
```

IBratWurst (approved):
Approved, but please see the comment. If I'm not missing something it may improve the readability

taye:
@IBratWurst I don't think we can do this with multi-reference fields where adding multiple items should emit actions with index offset by the length of items

https://github.com/contentful/field-editors/pull/683/files#diff-2bf90183b49109823b06a7a0b234b1010a286ae866bfb899dfec7589364fbec4R57-R65

PR #702, opened 2021-04-28
feat: allow custom plugins for slate

mcheng-caring:
## Description

Adds a `customPlugins` prop to the `RichTextEditor` component. This allows the user to override existing node renderers or and/or add their own node renderers used by Slate.

### DIFF
--- a/packages/rich-text/src/RichTextEditor.jsx
+++ b/packages/rich-text/src/RichTextEditor.jsx
@@ -136,7 +136,7 @@ export class ConnectedRichTextEditor extends React.Component {
     onAction: this.props.onAction,
   });
 
-  slatePlugins = buildPlugins(this.richTextAPI);
+  slatePlugins = buildPlugins(this.richTextAPI, this.props.customPlugins);
 
   onChange = (editor) => {
     const { value, operations } = editor;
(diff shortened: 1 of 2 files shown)


cribbles:
Hi @mcheng-caring, thanks for the PR 👍 

This approach seems fine to me, looks like you'll need to add `customPlugins` to the `propTypes` though: https://circleci.com/gh/contentful/field-editors/5967

JSYK, we're currently working on upgrading the rich text editor. When it's ready (should be sometime later this year), this will be a major version bump that will not break compatibility with the existing format, but this means you may need to rewrite any custom plugins built using this approach.

JimmyHowe:
Hey @mcheng-caring, any update on this? I'm currently looking into a solution for code blocks for a blog and thinking around ways to do this, I created my own field that has a code editor but would be great to do it in the rich text editor directly, thanks

mcheng-caring:
> Hey @mcheng-caring, any update on this? I'm currently looking into a solution for code blocks for a blog and thinking around ways to do this, I created my own field that has a code editor but would be great to do it in the rich text editor directly, thanks

Hi @JimmyHowe, at my company, we've been using a fork with the code change from this PR for over a year now with success. I haven't had the time to loop back to this PR to check what tests were failing due to time constraints and because we deprioritize getting this PR through due to the news that rich text editor was supposed to have been updated (AFAIK, the update is still only present in the `next` branch). If you have any questions about the fork, I can try my best to help.

Noprop:
Jumping in on this -

Why does the Rich Text editor not allow direct modification of the HTML?
"we're currently working on upgrading the rich text editor" is this something that is still happening @cribbles ?

Wordpress at least does something right where you have rich text + can easily modify it to your liking. e.g. add classnames

PR #703, opened 2021-04-28
feat: add alpha rich text editor

cribbles:
# Summary

Introduces an alpha `RichTextEditor` package. This is just a placeholder component exporting an instance of the latest version of Slate.js.

Adds to the widgets supplied in the `default-field-editors` package, albeit without any parameters for now (we'll have to change this when the rich text editor can take sdk props and export editor state). This allows us to export the field in the webapp and Compose, among other places we might want to test this component (extensions, etc).

The babel, eslint and tsconfig plugins are all cloned from and left in mostly the same state as in the `RichTextEditor`. As we're planning on going with Cypress for integration tests but have no need for Jest - yet - I've added placeholder `tsdx test` commands in the `package.json` so that CI will pass but note we're not actually running anything.

# Visual preview

For now this is just visible in `docz`. It looks like this:

image

# Notes

We get out-of-the-box `paragraph` node support (in SlateJS's internal format, remember we're not using the adapter at this point). So for the above example the markup propagated through the Slate context looks like:

```json
[
    {
        "type": "paragraph",
        "children": [
            {
                "text": "A line of text in a paragraph."
            }
        ]
    },
    {
        "type": "paragraph",
        "children": [
            {
                "text": ""
            }
        ]
    },
    {
        "type": "paragraph",
        "children": [
            {
                "text": "Another line of text: Check this one out. It's even better"
            }
        ]
    }
]
```


packages/rich-text-alpha/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@contentful/field-editor-rich-text-alpha",
```
DanweDE, line 2: **Nit:** I'd prefer `@contentful/field-editor-rich-text-2-alpha` or `@contentful/field-editor-rich-text-next` as a wip name but it probably doesn't matter much.

DanweDE (approved):

packages/rich-text-alpha/package.json
cribbles: I get your reasoning. I prefer `next` of the two, `-2-alpha` is awkward :D

z0al: Any reason not to use release channels? e.g. `alpha` or `next` under the current package name e.g. `@contentful/field-editor-rich-text@next` and later land with a major version bump when it's ready. That would mean using a different branch though e.g. `rich-text-v2` or something

z0al: One advantage I see (in addition to being cleaner IMO) is that we don't have to change the `widgetId` everywhere when we are slowly rolling it out. I imagine `richTextEditor` being hard-coded in a couple of places that can be a bit of a hassle to adjust while being behind a feature flag.

cribbles: One nice thing about releasing it as its own widget is that we'll want to use a separate `widgetId` anyway when deciding which editor to choose in the webapp, Compose etc. Maybe there's a way around that I'm not thinking of though.

z0al: > One nice thing about releasing it as its own widget is that we'll want to use a separate widgetId 

It could be because I'm not fully aware of how we plan to move forward on this but I actually see that as an issue (as explained in my previous comment). My understanding is that there won't be any major changes to the _public_ API shape, so in a way, this package is a drop-in replacement - but incomplete as of now- for the current `@contentful/field-editor-rich-text` and can be used (again, not complete yet) wherever `richTextEditor` widget is used, is that right?

If yes, then having a different `widgetId` might not work in our favour, because we will gradually roll out changes using the `alphaRichTextEditor` and then again gradually roll out replacing it with the actual `richTextEditor`.

cribbles: Well, if we decided to go with this approach we'd need to introduce some prop in `default-field-editors` to support it. I'm not opposed to it but also don't feel too strongly and don't feel like I get the complete vision here, happy to pair with you on this early next week or accept a PR to this PR.

cezarsmpio (approved):

djagya (approved):

PR #709, opened 2021-05-04
feat: {wip} use rich text adapter

cribbles:
# Summary

- [x] Converts SlateJS markup to Contentful markup using the adapter via supplied `onChange` prop
- [x] Adds a [SlateJS schema](https://docs.slatejs.org/v/v0.47/slate-core/schema) based on [the one we're using in `master` currently](https://github.com/contentful/field-editors/blob/master/packages/rich-text/src/constants/Schema.js)

# TODO

- [ ] Convert Contentful markup back to SlateJS markup using state supplied from the SDK
- [ ] Add tests

### DIFF
--- a/packages/rich-text/src/RichTextEditor.mdx
+++ b/packages/rich-text/src/RichTextEditor.mdx
@@ -15,5 +15,5 @@ import { css } from 'emotion';
 ## In Action
 
 <Playground>
-  <RichTextEditor />
+  <RichTextEditor onChange={console.log} />
 </Playground>
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -1,11 +1,10 @@
-// Import React dependencies.
 import React, { useMemo, useState } from 'react'
-
-// Import the Slate editor factory.
 import { createEditor, BaseEditor } from 'slate'
-
-// Import the Slate components and React plugin.
 import { Slate, Editable, withReact, ReactEditor } from 'slate-react'
+import debounce from 'lodash-es/debounce'
+import { toContentfulDocument } from '@contentful/contentful-slatejs-adapter'
+import * as Contentful from '@contentful/rich-text-types';
+import schema from './constants/Schema';
 
 type CustomElement = {
   type: 'paragraph';
@@ -21,11 +20,18 @@ declare module 'slate' {
   }
 }
 
+const toContentfulDocumentDebounced = debounce((document, schema) => {
+  return toContentfulDocument({ document, schema });
+}, 500);
+
+type Props = {
+  onChange: (doc: Contentful.Document) => unknown;
+};
+
 // Define our app...
-const RichTextEditor = () => {
+const RichTextEditor = (props: Props) => {
   const editor = useMemo(() => withReact(createEditor()), [])
 
-  // Keep track of state for the value of the editor.
   const [value, setValue] = useState([
     {
       type: 'paragraph',
@@ -33,12 +39,15 @@ const RichTextEditor = () => {
     },
   ] as CustomElement[])
 
-  // Render the Slate context.
   return (
     <Slate
       editor={editor}
       value={value}
-      onChange={newValue => setValue(newValue as CustomElement[])}>
+      onChange={newValue => {
+        setValue(newValue as CustomElement[]);
+        const doc = toContentfulDocumentDebounced(newValue, schema);
+        props.onChange(doc);
+      }}>
       <Editable />
     </Slate>
   )
--- a/packages/rich-text/src/constants/Schema.ts
+++ b/packages/rich-text/src/constants/Schema.ts
@@ -1,0 +1,140 @@
+import {
+  BLOCKS,
+  INLINES,
+  TOP_LEVEL_BLOCKS,
+  VOID_BLOCKS,
+  CONTAINERS
+} from '@contentful/rich-text-types';
+
+const inlines = Object.values(INLINES).map((type) => ({ type }));
+
+export default {
+  document: {
+    nodes: [
+      {
+        types: TOP_LEVEL_BLOCKS.map((type) => ({ type }))
+      }
+    ]
+  },
+
+  blocks: {
+    [BLOCKS.PARAGRAPH]: {
+      nodes: [
+        {
+          match: [
+            ...inlines,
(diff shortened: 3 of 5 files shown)


cypress/plugins/index.js
```diff
@@ -6,6 +6,10 @@ module.exports = on => {
       resolve: {
         extensions: ['.ts', '.js']
       },
+      output: {
+        hotUpdateChunkFilename: '[id].[fullhash].hot-update.js',
+        hotUpdateMainFilename: '[runtime].[fullhash].hot-update.json',
+      },
```
cribbles, line 12: needed to prevent errors with the latest cypress. cf https://github.com/webpack/webpack/issues/6693#issuecomment-745688108

cypress/support/commands.js
```diff
@@ -72,3 +72,11 @@ Cypress.Commands.add('getRichTextField', () => {
...
   });
 });
+
+Cypress.Commands.add('typeInSlate', { prevSubject: true }, (subject, text) => {
+  return cy.wrap(subject)
+    .then(subject => {
+      subject[0].dispatchEvent(new InputEvent('beforeinput', { inputType: 'insertText', data: text }));
+      return subject;
+    })
+});
```
cribbles, line 82: needed because cypress's `type` doesn't trigger `beforeinput` events by default

packages/rich-text/src/RichTextEditor.mdx
```diff
@@ -10,10 +10,109 @@ import Readme from '../README.md';
...
 
 <Playground>
-  <RichTextEditor />
+  {() => {
+    const rtPreviewStyle = css({
+      backgroundColor: 'whitesmoke',
+      padding: '0 0.5em'
+    });
+    const newEntitySelectorDummyDialog = (fnName, type) => async () => {
+      return confirm(`sdk.dialogs.${fnName}()\nSimulate selecting a random entity or cancel?`)
```
cribbles, line 34: Most of this is just copied verbatim from the file in `master`. I think it's okay to have some unused mocks in here since we know they'll be used rather soon as we develop forthcoming plugins

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -21,27 +22,75 @@ declare module 'slate' {
   }
 }
 
-// Define our app...
-const RichTextEditor = () => {
+const toContentfulDocumentDebounced = (document, schema) => {
+  // TODO: Do we need to debounce 500ms, like in master?
```
cribbles, line 26: I'm seeking ideas from reviewers here. We debounce the conversion in `master` by 500ms, not sure if we need to be doing this anymore due to ShareJS removal

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -21,27 +22,75 @@ declare module 'slate' {
...
+    schema
+  });
+  
+  const [value, setValue] = useState(document as CustomElement[])
 
-  // Render the Slate context.
   return (
     <Slate
       editor={editor}
+      // TODO: normalize like in the webapp?
```
cribbles, line 54: cf. https://github.com/contentful/field-editors/blob/master/packages/rich-text/src/RichTextEditor.jsx#L69-L85

cypress/plugins/index.js
z0al: I'd add such comments as code comments for future reader

packages/rich-text/src/constants/Schema.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    [BLOCKS.QUOTE]: {
+      nodes: [
+        {
+          match: [CONTAINERS[BLOCKS.QUOTE].map(type => ({ type }))],
+          min: 1
+        }
+      ],
+      normalize: (editor, error) => {
+        // TODO: still a relevant error code from Slate?
```
z0al, line 107: Looks like it. Didn't find a relevant note in the change log since `v0.44.0`

packages/rich-text/src/RichTextEditor.tsx
z0al: If we are not sure then let's maintain the same behaviour as in `master` for now until we actually investigate the reason behind it a bit more.

z0al (commented):
Looks good to me. Left a few tiny comments

packages/rich-text/src/RichTextEditor.tsx
cribbles: yeah, makes sense 👍

cezarsmpio (approved):
Looks good to me

cypress/support/commands.js
```diff
@@ -72,3 +72,17 @@ Cypress.Commands.add('getRichTextField', () => {
     return win.richTextField;
   });
 });
+
+Cypress.Commands.add('typeInSlate', { prevSubject: true }, (subject, text) => {
+  // Needed because cypress's `type` doesn't trigger `beforeinput` events by default.
```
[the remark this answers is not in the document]
cribbles: I thought I did, but then I triple checked and realized Cypress was nevertheless reverting to v5.2.0 locally due to resolution settings in the repo-level `package.json`. Fixed in https://github.com/contentful/field-editors/pull/709/commits/4207b67ce3cf34293b55839f2c6c6fcb369d0fb6 (will have to wait to see if CI passes though).

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -21,27 +22,76 @@ declare module 'slate' {
...
+
+const RichTextEditor = (props: Props) => {
+  const { sdk, isInitiallyDisabled, ...otherProps } = props;
+  return (
+    <EntityProvider sdk={sdk}>
+      <FieldConnector
+        throttle={0}
+        field={sdk.field}
+        isInitiallyDisabled={isInitiallyDisabled}
+        isEmptyValue={(value) => !value || deepEquals(value, Contentful.EMPTY_DOCUMENT)}
```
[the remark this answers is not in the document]
cribbles: Don't see any problem with this. `useCallback` should be fine tbh

z0al (approved):

cribbles:
Wrote https://contentful.atlassian.net/browse/SHE-80 to encapsulate forgoing cypress migration, also left a TODO about it.

PR #716, opened 2021-05-07
feat: add marks support [SHE-12]

cezarsmpio:
Checklist:

- [x] Bold
- [x] Italic
- [x] Underline
- [x] Code
- [x] Leaf
- [x] Cypress Tests

Initial content with some marks being rendered accordingly:

image

Hotkeys are also working. Tested on Chrome, Firefox, and Safari.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@typescript-eslint/parser": "^4.10.0",
     "contentful-cli": "^1.3.29",
     "cssnano": "4.1.11",
-    "cypress": "5.2.0",
+    "cypress": "^7.2.0",
     "cz-lerna-changelog": "^2.0.2",
     "docz": "1.3.2",
     "docz-theme-default": "1.2.0",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@typescript-eslint/parser": "^4.10.0",
     "contentful-cli": "^1.3.29",
     "cssnano": "4.1.11",
-    "cypress": "^7.2.0",
+    "cypress": "^5.2.0",
     "cz-lerna-changelog": "^2.0.2",
     "docz": "1.3.2",
     "docz-theme-default": "1.2.0",
@@ -104,7 +104,7 @@
   "resolutions": {
     "@types/react": "16.14.5",
     "babel-preset-react-app": "9.1.2",
-    "cypress": "7.2.0",
+    "cypress": "5.2.0",
     "react": "16.10.2",
     "react-dom": "16.10.2"
   },
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@typescript-eslint/parser": "^4.10.0",
     "contentful-cli": "^1.3.29",
     "cssnano": "4.1.11",
-    "cypress": "^5.2.0",
+    "cypress": "5.2.0",
     "cz-lerna-changelog": "^2.0.2",
     "docz": "1.3.2",
     "docz-theme-default": "1.2.0",
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
   "resolutions": {
     "@types/react": "16.14.5",
     "babel-preset-react-app": "9.1.2",
-    "cypress": "5.2.0",
+    "cypress": "7.2.0",
     "react": "16.10.2",
     "react-dom": "16.10.2"
   },
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -1,48 +1,82 @@
-import React, { useCallback, useMemo, useState } from 'react'
-import { createEditor, BaseEditor } from 'slate'
-import { Slate, Editable, withReact, ReactEditor } from 'slate-react'
-import { toContentfulDocument, toSlatejsDocument } from '@contentful/contentful-slatejs-adapter'
+import React, { useCallback, useMemo, useState } from 'react';
+import { createEditor, BaseEditor } from 'slate';
+import {
+  Slate,
+  Editable,
+  withReact,
+  ReactEditor,
+  DefaultElement,
+  RenderLeafProps,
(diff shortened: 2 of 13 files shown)


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -1,47 +1,120 @@
...
+type CustomText = { text: string };
 
 declare module 'slate' {
   interface CustomTypes {
-    Editor: BaseEditor & ReactEditor
-    Element: CustomElement
-    Text: CustomText
+    Editor: BaseEditor & ReactEditor;
+    Element: CustomElement;
+    Text: CustomText;
```
cezarsmpio, line 41: Prettier changes, sorry for this noise.

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -1,47 +1,120 @@
...
 const RichTextEditor = () => {
-  const editor = useMemo(() => withReact(createEditor()), [])
+  const editor = useMemo<ContentfulEditor>(() => withPlugins(createEditor()), []);
 
   // Keep track of state for the value of the editor.
   const [value, setValue] = useState([
     {
       type: 'paragraph',
-      children: [{ text: 'A line of text in a paragraph.' }],
+      children: [
```
cezarsmpio, line 60: Added a more complex dummy value. For testing purpose.

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -1,47 +1,120 @@
...
 
   // Render the Slate context.
   return (
     <Slate
       editor={editor}
       value={value}
-      onChange={newValue => setValue(newValue as CustomElement[])}>
-      <Editable />
+      onChange={(newValue) => setValue(newValue as CustomElement[])}>
+      <Editable
```
cezarsmpio, line 111: @m10l this is where we will add the toolbar. Example: https://github.com/ianstormtaylor/slate/blob/main/site/examples/richtext.tsx#L32

packages/rich-text/src/helpers/compose.ts
```diff
@@ -0,0 +1,6 @@
+export function compose(...fns: Function[]) {
```
cezarsmpio, line 1: Should we get this from some library?

packages/rich-text/src/plugins/Leaf/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  if ('italic' in leaf) {
+    children = <Italic {...props}>{children}</Italic>;
+  }
+
+  if ('underline' in leaf) {
+    children = <Underline {...props}>{children}</Underline>;
+  }
+
+  return <span {...attributes}>{children}</span>;
```
cezarsmpio, line 29: We can also just return `children`.

packages/rich-text/src/RichTextEditor.tsx
m10l: Lovely, ty

packages/rich-text/src/helpers/compose.ts
taye: You could use https://docs-lodash.com/v4/flow/

package.json
```diff
@@ -66,7 +66,7 @@
     "@typescript-eslint/parser": "^4.10.0",
     "contentful-cli": "^1.3.29",
     "cssnano": "4.1.11",
-    "cypress": "5.2.0",
+    "cypress": "^7.2.0",
```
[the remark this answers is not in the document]
cezarsmpio: tests

z0al (approved):
LGTM

packages/rich-text/src/plugins/Leaf/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Italic } from '../Italic';
+import { Underline } from '../Underline';
+
+export function Leaf(props: RenderLeafProps) {
+  const { attributes, leaf } = props;
+  let { children } = props;
+
+  if ('bold' in leaf) {
+    children = <Bold {...props}>{children}</Bold>;
+  }
```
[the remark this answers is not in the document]
z0al: Good catch Taye. Looking at this again, it seems like it will execute the if-block even if `leaf.bold=false`

I think we should be using what you suggested. Unless I'm missing something too :)

package.json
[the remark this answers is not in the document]
cezarsmpio: Unfortunately, `7.2.0` breaks other tests. That's not good :/ I've downgraded it to `v5.5.0` and it also breaks other tests...

I think I will add that workaround for the time being so we can think of a better solution.

How does that sound?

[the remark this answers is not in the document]
cezarsmpio: It seems that @cribbles has done the homework here 🤦 my bad

packages/rich-text/src/RichTextEditor.tsx
cezarsmpio: I'm up to that. Usually, Prettier will also read your settings from eslint rules.

packages/rich-text/src/plugins/Bold/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return <button>Bold</button>;
+}
+
+const styles = {
+  bold: css({
+    fontWeight: 600,
+  }),
+};
+
+export function Bold(props: Slate.RenderLeafProps) {
```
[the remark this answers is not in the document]
cezarsmpio: I don't have a strong opinion against the first example, I like both ways :) 

Now regarding the `Leaf` component, I'm wondering if it makes it harder to read? We will always have only 4 marks.

Let's see if somebody else has a different opinion about it.

cribbles (approved):

z0al (approved):
👏🏻

taye (approved):

PR #728, opened 2021-05-17
feat: add HR component support

cezarsmpio:
- [x] Toolbar button/icon
- [x] Styles
- [x] Cypress test
- [x] Current behavior (except the deleting one we will need to discuss it further)

### DIFF
--- a/packages/rich-text/src/hooks/useCustomEditor.ts
+++ b/packages/rich-text/src/hooks/useCustomEditor.ts
@@ -1,0 +1,8 @@
+import { useSlateStatic } from 'slate-react';
+import { CustomEditor } from '../types';
+
+export function useCustomEditor() {
+  const editor = useSlateStatic() as CustomEditor;
+
+  return editor;
+}
--- a/packages/rich-text/src/plugins/Editor/index.ts
+++ b/packages/rich-text/src/plugins/Editor/index.ts
@@ -13,7 +13,7 @@ export function withEditorPlugin(editor: CustomEditor): CustomEditor {
     return !!match;
   }
 
-  function isVoid(element: CustomElement): boolean {
+  function isVoid(element): boolean {
     return element.isVoid || originalIsVoid(element);
   }
 
--- a/packages/rich-text/src/plugins/Editor/index.ts
+++ b/packages/rich-text/src/plugins/Editor/index.ts
@@ -17,8 +17,17 @@ export function withEditorPlugin(editor: CustomEditor): CustomEditor {
     return element.isVoid || originalIsVoid(element);
   }
 
+  function hasSelectionText() {
+    return editor.selection
+      ? Editor.node(editor, editor.selection.focus.path).some(
+          (node) => node.text && node.text !== ''
+        )
+      : false;
+  }
+
   editor.isBlockSelected = isBlockSelected;
   editor.isVoid = isVoid;
+  editor.hasSelectionText = hasSelectionText;
 
   return editor;
 }
--- a/packages/rich-text/src/plugins/Editor/index.ts
+++ b/packages/rich-text/src/plugins/Editor/index.ts
@@ -1,4 +1,4 @@
-import { Editor, Element } from 'slate';
+import { Editor, Element, Text } from 'slate';
 import { CustomEditor, CustomElement } from '../../types';
 
 export function withEditorPlugin(editor: CustomEditor): CustomEditor {
@@ -20,7 +20,7 @@ export function withEditorPlugin(editor: CustomEditor): CustomEditor {
   function hasSelectionText() {
     return editor.selection
       ? Editor.node(editor, editor.selection.focus.path).some(
-          (node) => node.text && node.text !== ''
+          (node) => Text.isText(node) && node.text !== ''
         )
       : false;
   }
--- a/packages/rich-text/src/plugins/Editor/index.ts
+++ b/packages/rich-text/src/plugins/Editor/index.ts
@@ -1,0 +1,24 @@
+import { Editor, Element } from 'slate';
+import { CustomEditor, CustomElement } from '../../types';
+
+export function withEditorPlugin(editor: CustomEditor): CustomEditor {
+  const { isVoid: originalIsVoid } = editor;
+
+  function isBlockSelected(type: string): boolean {
+    const [match] = Array.from(
+      Editor.nodes(editor, {
+        match: (node) => Element.isElement(node) && (node as CustomElement).type === type,
+      })
+    );
+    return !!match;
+  }
+
+  function isVoid(element: CustomElement): boolean {
+    return element.isVoid || originalIsVoid(element);
+  }
+
+  editor.isBlockSelected = isBlockSelected;
+  editor.isVoid = isVoid;
+
+  return editor;
+}
--- a/packages/rich-text/src/plugins/Editor/index.ts
+++ b/packages/rich-text/src/plugins/Editor/index.ts
@@ -1,4 +1,4 @@
-import { Editor, Element, Text } from 'slate';
+import { Editor, Element, Text, Transforms } from 'slate';
 import { CustomEditor, CustomElement } from '../../types';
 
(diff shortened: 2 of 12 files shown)


packages/rich-text/src/hooks/useCustomEditor.ts
```diff
@@ -0,0 +1,8 @@
+import { useSlateStatic } from 'slate-react';
+import { CustomEditor } from '../types';
+
+export function useCustomEditor() {
+  const editor = useSlateStatic() as CustomEditor;
+
+  return editor;
+}
```
cribbles, line 8: what does this do?

packages/rich-text/src/plugins/Editor/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function withEditorPlugin(editor: CustomEditor): CustomEditor {
+  const { isVoid: originalIsVoid } = editor;
+
+  function isBlockSelected(type: string): boolean {
+    const [match] = Array.from(
+      Editor.nodes(editor, {
+        match: (node) => Element.isElement(node) && (node as CustomElement).type === type,
+      })
+    );
```
cribbles, line 12: This seems kinda odd to me but maybe I just don't understand [the API](https://docs.slatejs.org/api/nodes/editor#editor-nodes-editor-editor-options-generator-less-than-nodeentry-less-than-t-greater-than-void-undefined-greater-than). I would guess from this logic this doesn't tell us whether the block is selected, but whether a block of type `type` appears in the editor markup somewhere. Is that right? And where does this get used?

packages/rich-text/src/plugins/Editor/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    return !!match;
+  }
+
+  function isVoid(element): boolean {
+    return element.isVoid || originalIsVoid(element);
```
cribbles, line 17: `originalIsVoid` comes ultimately from the schema, right? When would `element.isVoid` be false but not the schema `isVoid`?

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+interface ToolbarHrButtonProps {
+  isDisabled?: boolean;
+}
+
+export function withHrEvents(editor, event) {
+  const [currentFragment] = Editor.fragment(editor, editor.selection.focus.path) as CustomElement[];
+
+  if (
+    event.keyCode === 13 &&
```
cribbles, line 51: Would you mind turning this into a const? Couldn't remember which one this was.

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    Editor.isVoid(editor, currentFragment) &&
```
cribbles, line 52: Why do we need to asset whether the the current fragment is void? Are there cases where the `type` would be `HR` but not void, and that would determine the outcome?

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    currentFragment.type === BLOCKS.HR
+  ) {
+    event.preventDefault();
+
+    Transforms.move(editor, { distance: 1 });
```
cribbles, line 57: What does this do? (could encapsulate this if block into a function with a clarifying name, or add a comment or something like that - behavior isn't obvious at first glance)

packages/rich-text/src/hooks/useCustomEditor.ts
cezarsmpio: Did this to have the editor as a CustomEditor type, to facilitate this type conversion.

packages/rich-text/src/plugins/Editor/index.ts
cezarsmpio: I noticed from the small experience that I had that it [always gets from the current selection](https://docs.slatejs.org/concepts/04-transforms#the-at-option) unless we manually override it. I think I forgot to remove it, but I used it at some point, however, I'm using the same thing for the headings.

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: Haha, yeah, sure :) `13` is `Enter`.

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: You are maybe right, if we have `currentFragment.type === BLOCKS.HR` we might not need this `Editor.isVoid` checking.

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: This moves the text cursor to the next "line". I can create a method and name it "moveToTheNextSelection" or something like that. Any name ideas?

m10l: I think just adding a comment might be fine for now, if we need to reuse it (could be likely that we will) we can do that later

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+interface ToolbarHrButtonProps {
+  isDisabled?: boolean;
+}
+
+export function withHrEvents(editor: CustomEditor, event: KeyboardEvent) {
+  if (!editor.selection) return;
+
+  const [currentFragment] = Editor.fragment(editor, editor.selection.focus.path) as CustomElement[];
+  const isEnter = event.keyCode === 13;
```
taye, line 51: You can use `event.key === 'Enter'`, which is clearer than the key code.

We should probably also check that there are no modifier keys on the event.

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export function ToolbarHrButton(props: ToolbarHrButtonProps) {
+  const editor = useCustomEditor();
+
+  function handleOnClick() {
+    const hr = {
+      type: BLOCKS.HR,
+      children: [{ text: '' }],
+      isVoid: true,
```
taye, line 68: I don't know what slate expects, but it seems odd that this void block would have a text child 🤔

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+      tooltip="HR"
+      label="HR"
+      disabled={props.isDisabled}
+      onClick={handleOnClick}
+      testId="hr-toolbar-button"
+    />
+  );
+}
+
+export function Hr(props: Slate.RenderLeafProps) {
```
taye, line 99: I'm thinking again that having consistent names of plugin exports would be a good idea https://github.com/contentful/field-editors/pull/716#discussion_r629498736. @contentful/team-shelley what do you think?

taye (approved):
LGTM. Added a few thoughts

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: Yeah, in this case, the name of the variable already tells us this an `Enter`.

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: It's how it's supposed to be from now on.. I think they have it explained on their documentation.

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: I think that might not be good to find references for it because we will have many methods with the same name. I also don't have a stronger opinion than that, it's simpler to write but not sure about maintenance.

PR #732, opened 2021-05-27
feat: add support for headings [SHE-16]

cezarsmpio:
- [x] Toolbar
- [x] Hotkeys
- [x] Cypress test
- [x] Styling
- [x] Mimic current editor behavior

### DIFF
--- a/packages/rich-text/src/RichTextEditor.styles.ts
+++ b/packages/rich-text/src/RichTextEditor.styles.ts
@@ -13,6 +13,8 @@ export const styles = {
     borderTop: 0,
     padding: '20px',
     fontSize: tokens.spacingM,
+    fontFamily:
+      '-apple-system,BlinkMacSystemFont,Segoe UI,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol',
     minHeight: '400px',
     background: tokens.colorWhite,
     outline: 'none',
--- a/packages/rich-text/src/RichTextEditor.styles.ts
+++ b/packages/rich-text/src/RichTextEditor.styles.ts
@@ -13,8 +13,7 @@ export const styles = {
     borderTop: 0,
     padding: '20px',
     fontSize: tokens.spacingM,
-    fontFamily:
-      '-apple-system,BlinkMacSystemFont,Segoe UI,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol',
+    fontFamily: tokens.fontStackPrimary,
     minHeight: '400px',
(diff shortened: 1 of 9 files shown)


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -9,7 +9,7 @@ import { styles } from './RichTextEditor.styles';
 import { FieldExtensionSDK, FieldConnector } from '@contentful/field-editor-shared';
 import schema from './constants/Schema';
 import deepEquals from 'fast-deep-equal';
-import debounce from 'lodash/debounce';
+// import debounce from 'lodash/debounce';
```
cezarsmpio, line 12: Remove it temporarily until we figure out whether we need it or not because this was causing some actions to act slow on some user cases.

cribbles (approved):
Nice clear code. Works great with `yarn docz:dev` too :D

packages/rich-text/src/RichTextEditor.styles.ts
```diff
@@ -13,6 +13,8 @@ export const styles = {
     borderTop: 0,
     padding: '20px',
     fontSize: tokens.spacingM,
+    fontFamily:
+      '-apple-system,BlinkMacSystemFont,Segoe UI,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol',
```
m10l, line 17: we can get this from F36 tokens - `fontStackPrimary` is the token we'd use

m10l: We can do this as a separate PR, there are a few other things that can be turned into tokens (and fontSize is currently using a spacing token, oops)

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -60,6 +62,18 @@ const ConnectedRichTextEditor = (props: ConnectedProps) => {
...
+      case Contentful.BLOCKS.HEADING_2:
+        return <H2 {...props} />;
+      case Contentful.BLOCKS.HEADING_3:
+        return <H3 {...props} />;
+      case Contentful.BLOCKS.HEADING_4:
+        return <H4 {...props} />;
+      case Contentful.BLOCKS.HEADING_5:
+        return <H5 {...props} />;
+      case Contentful.BLOCKS.HEADING_6:
+        return <H6 {...props} />;
```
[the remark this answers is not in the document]
cezarsmpio: I agree :)

packages/rich-text/src/RichTextEditor.styles.ts
cezarsmpio: Thanks, just added it ;)

PR #734, opened 2021-06-01
feat: gracefully handle void elements

anho:
Usage of void elements in the markdown crashes the preview because React won't let us render children for those elements. We hereby surface that problem to the user in order to give them a way out of that otherwise dead end.

image

### DIFF
--- a/packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
+++ b/packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import { useVoidElements } from './useVoidElements';
+import { Note, Paragraph, TextLink } from '@contentful/forma-36-react-components';
+import tokens from '@contentful/forma-36-tokens';
+
+import { css } from 'emotion';
+
+const styles = {
+  note: css({
+    margin: `${tokens.spacingM} 0`
+  }),
+  paragraph: css({
+    '&:not(:last-child)': {
+      paddingBottom: tokens.spacingS
+    }
+  })
+};
+
+export function VoidElementsWarning() {
+  const { isEmpty } = useVoidElements();
+
+  if (isEmpty) {
+    return null;
+  }
+
+  return (
+    <Note noteType="warning" hasCloseButton={false} className={styles.note} testId="markdown-void-elements-warning">
+      <Paragraph className={styles.paragraph}>
+        Your markdown contains <TextLink linkType="primary" icon="ExternalLink" iconPosition="right">void
+        elements</TextLink> that causes the preview to skip those parts. This indicates incorrect usage of one or more
+        of the following HTML elements: area, base, br, col, command, embed, hr, img, input, keygen, link, meta, param,
+        source, track, wbr.
+      </Paragraph>
+      <Paragraph className={styles.paragraph}>
+        In order to properly display the preview fix the elements or replace them with alternatives.
+      </Paragraph>
+    </Note>
+  );
+}
--- a/packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
+++ b/packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
@@ -26,7 +26,7 @@ export function VoidElementsWarning() {
   return (
     <Note noteType="warning" hasCloseButton={false} className={styles.note} testId="markdown-void-elements-warning">
       <Paragraph className={styles.paragraph}>
-        Your markdown contains <TextLink linkType="primary" icon="ExternalLink" iconPosition="right"
+        Your markdown contains <TextLink linkType="primary" icon="ExternalLink" iconPosition="right" target="_blank"
                                          href="https://www.w3.org/TR/2011/WD-html-markup-20110113/syntax.html#void-element">void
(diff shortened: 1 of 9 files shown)


packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  if (isEmpty) {
+    return null;
+  }
+
+  return (
+    <Note noteType="warning" hasCloseButton={false} className={styles.note} testId="markdown-void-elements-warning">
+      <Paragraph className={styles.paragraph}>
+        Your markdown contains <TextLink linkType="primary" icon="ExternalLink" iconPosition="right">void
+        elements</TextLink> that causes the preview to skip those parts. This indicates incorrect usage of one or more
```
zcei, line 30:
```suggestion
        elements</TextLink> that cause the preview to skip those parts. This indicates incorrect usage of one or more
```

packages/markdown/src/components/VoidElements/useVoidElements.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const [VoidElementsContext, useVoidElements] = constate(() => {
+  const [counter, setCounter] = useState(0);
+  const add = useCallback(() => {
+    console.log('adding');
+    setCounter((current) => current + 1);
+
+  }, [setCounter]);
+  const remove = useCallback(() => {
+    console.log('removing');
```
zcei, line 12: Leftover `console.log`s here and in l. 7?

packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
```diff
@@ -0,0 +1,39 @@
...
+        of the following HTML elements: area, base, br, col, command, embed, hr, img, input, keygen, link, meta, param,
```
zcei, line 31: Would be cool to only show the ones that are actively used / derive the whole list from what we guard against, but probably effort and value have no match here?

anho: Yep, for an edge-case like this it feels too much effort.

packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
anho: Sure? :)

zcei: Why does my brain do the Germans-learning-English-fart:

> He, she, it - das s muss mit.

As it's multiple elements (3rd person plural) that cause something, I'd argue it's without the "s", or do we have some edge-case grammar here? 🤔

zcei: DeepL says I'm wrong 😆

packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { isEmpty } = useVoidElements();
+
+  if (isEmpty) {
+    return null;
+  }
+
+  return (
+    <Note noteType="warning" hasCloseButton={false} className={styles.note} testId="markdown-void-elements-warning">
+      <Paragraph className={styles.paragraph}>
+        Your markdown contains <TextLink linkType="primary" icon="ExternalLink" iconPosition="right">void
```
artibella, line 29: Suggestion: Start the message with a TL;DR of the most important information: e.g. `Partial preview only`.
Then add some more details what causes this behaviour and how to fix it. @aromanko What do you think?

Where is the external link pointing to? I don't see a URL in the code

anho: @artibella I think "Your markdown contains void elements that causes preview to skip those parts" is pretty close to what you are suggesting. I'd like to get this merged and deployed soon and will come back to the copy if Anna wants a change of copy.

artibella: About the copy: Ok for me to do it as you suggested

About the link URL: I am still confused where this external link is pointing to as I can not see it from the code. But maybe I am just missing something.

anho: @artibella it's pointing to https://www.w3.org/TR/2011/WD-html-markup-20110113/syntax.html#void-element

artibella: Thanks @anho

suevalov (approved):

IBratWurst (approved):

packages/markdown/src/components/VoidElements/VoidElementsWarning.tsx
aromanko: > Suggestion: Start the message with a TL;DR of the most important information: e.g. `Partial preview only`.
> Then add some more details what causes this behaviour and how to fix it. @aromanko What do you think?
> 
> Where is the external link pointing to? I don't see a URL in the code

"Partial preview only" sounds good to me, it can be formatted as heading and will provide the reader with some extra context.

PR #737, opened 2021-06-07
feat: rich text blockquote plugin

m10l:
This PR adds the blockquote plugin to the new version of the Rich Text Editor

### DIFF
--- a/packages/rich-text/src/plugins/Editor/index.ts
+++ b/packages/rich-text/src/plugins/Editor/index.ts
@@ -34,20 +34,28 @@ export function withEditorPlugin(editor: CustomEditor): CustomEditor {
     const LIST_TYPES: string[] = [BLOCKS.OL_LIST, BLOCKS.UL_LIST];
     const isActive = isBlockSelected(type);
     const isList = LIST_TYPES.includes(type);
+    const isQuote = type === BLOCKS.QUOTE;
 
     Transforms.unwrapNodes(editor, {
       match: (node) =>
         !Editor.isEditor(node) &&
         Element.isElement(node) &&
-        LIST_TYPES.includes((node as CustomElement).type),
+        (LIST_TYPES.includes((node as CustomElement).type) ||
+          BLOCKS.QUOTE.includes((node as CustomElement).type)),
       split: true,
     });
     const newProperties: Partial<CustomElement> = {
-      type: isActive ? BLOCKS.PARAGRAPH : isList ? BLOCKS.LIST_ITEM : type,
+      type: isActive
+        ? BLOCKS.PARAGRAPH
+        : isList
+        ? BLOCKS.LIST_ITEM
+        : isQuote
+        ? BLOCKS.PARAGRAPH
+        : type,
     };
     Transforms.setNodes(editor, newProperties);
 
-    if (!isActive && isList) {
+    if (!isActive && (isList || isQuote)) {
       const block = { type, children: [] };
       Transforms.wrapNodes(editor, block);
     }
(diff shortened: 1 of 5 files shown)


packages/rich-text/src/plugins/Editor/index.ts
```diff
@@ -34,20 +34,28 @@ export function withEditorPlugin(editor: CustomEditor): CustomEditor {
...
     });
     const newProperties: Partial<CustomElement> = {
-      type: isActive ? BLOCKS.PARAGRAPH : isList ? BLOCKS.LIST_ITEM : type,
+      type: isActive
+        ? BLOCKS.PARAGRAPH
+        : isList
+        ? BLOCKS.LIST_ITEM
+        : isQuote
+        ? BLOCKS.PARAGRAPH
+        : type,
```
cezarsmpio, line 54: Should we find something more readable for this case? Maybe we will need to add more conditions here and a ternary if will not be the best to read.

packages/rich-text/src/plugins/Quote/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isDisabled?: boolean;
+}
+
+const createBlockQuote = (editor: CustomEditor) => {
+  if (!editor.selection) return;
+
+  const text = { text: '' };
+  const paragraph = { type: BLOCKS.PARAGRAPH, children: [text] };
+  const path = editor.selection.focus.path;
+  const parent = Editor.parent(editor, path);
```
cezarsmpio, line 33: I'm wondering if this is going to work if we start adding links to it, because links will add one more level so the parent won't be the parent you will be expecting, but this is just my theory.

You can also get the parent you want by using `Editor.levels` with the `reverse` property:

```js
const [currentListItemNode] = Editor.levels(editor, {
  at: editor.selection.focus.path,
  match: (node) =>
    Element.isElement(node) && (node as CustomElement).type === BLOCKS.LIST_ITEM,
  reverse: true,
});
```

I'm not sure if you will need it but we should test it later on. 

Can we add a TODO?

packages/rich-text/src/plugins/Quote/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    event.preventDefault();
+
+    const text = { text: '' };
+    const paragraph = { type: BLOCKS.PARAGRAPH, children: [text] };
+
+    if (editor.hasSelectionText()) {
+      const currentOffset = editor.selection.focus.offset;
+      const currentTextLength = Node.string(currentFragment).length;
+      const cursorIsAtTheBeginning = currentOffset === 0;
+      const cursorIsAtTheEnd = currentTextLength === currentOffset;
```
cezarsmpio, line 60: Maybe we should add this logic into the `Editor`? We are using the same thing on the headings and I will probably use the same thing for the lists.

packages/rich-text/src/plugins/Quote/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  if (isMod && isShift && isOneKey) {
+    createBlockQuote(editor);
+  }
+
+  // On backspace, check if quote is empty. If it's empty, switch the current fragment to a paragraph
+  if (isBackspace && currentFragment?.type === BLOCKS.QUOTE) {
+    if (
+      editor
+        .getElementFromCurrentSelection()[0]
+        .children.every((item) => item.children.every((item) => item.text === ''))
```
cezarsmpio, line 91: Can we move this to a new variable? Just for readability? :)

cezarsmpio (commented):
Just a few considerations.

packages/rich-text/src/plugins/Quote/index.tsx
m10l: Yeah I imagined this would break for links. I'll add a todo and link to this comment

cezarsmpio (approved):
Just a reminder that this might break when links are added.

PR #739, opened 2021-06-08
feat: 🎸 make validation component configurable

anho:


packages/default-field-editors/src/FieldWrapper.tsx
```diff
@@ -17,10 +18,12 @@ type FieldWrapperProps = {
   children: React.ReactNode;
   renderHeading?: (name: string) => JSX.Element | null;
   renderHelpText?: (helpText: string) => JSX.Element | null;
+  ValidationComponent?: React.Component<ValidationErrorsProps>
```
z0al, line 21: Would be nice to follow the same `render*` naming convention similar to the two props above it.

PR #742, opened 2021-06-10
feat: slatejs list support and port to the slatejs plugins next []

cezarsmpio:
Our team decided to use the [Slate Plugins](https://slate-plugins.udecode.io/docs/store) library instead of SlateJS directly.

Slate Plugins supports some of the components we will need for our next version, such as Lists and Tables, and it takes all the complexity of these components.

So in order to make it work, I also took advantage to port some other plugins we have, because of that, the PR is longer than it is used to be.

What's new:

- Lists

What was ported:

- Headings
- Marks (Bold, Italic, Underline and Code)
- HR

What's left:

- Blockquotes (@m10l will handle it)

Tests:

- I will add toolbar mark buttons tests on a different PR
- Also for list tests


cypress/integration/RichTextEditor.spec.ts
```diff
@@ -228,7 +228,7 @@ describe('Rich Text Editor', () => {
     });
   });
 
-  describe('Quote', () => {
+  describe.skip('Quote', () => {
```
cezarsmpio, line 231: @m10l I'm skipping quote tests, just fyi, otherwise tests won't pass on this branch.

packages/rich-text/src/helpers/editor.ts
```diff
@@ -0,0 +1,81 @@
+import { Text, Editor, Element, Transforms, Path } from 'slate';
```
cezarsmpio, line 1: Just moved stuff from the `Editor` plugin to this file, nothing "new" here.

packages/rich-text/src/plugins/Editor/index.ts
```diff
@@ -1,83 +0,0 @@
-import { Editor, Element, Text, Transforms } from 'slate';
```
cezarsmpio, line 1: This was moved to single functions on `helpers/editor.ts`.

m10l (approved):

cribbles (approved):

taye (approved):

PR #747, opened 2021-06-18
feat: Migrate extensions into CCA (entry-extension-collapsible) [EXT-2566]

alexdonets:
Ticket: https://contentful.atlassian.net/browse/EXT-2566

This PR is the first of several PRs which implement the migration of the extensions from CCE into CCA.
In this PR we are migrating the `entry-extension-collapsible` extension.

PS: I decided to do one PR per extension for readability and easier/faster code reviews. :)


extensions/entry-extension-collapsible/package.json
```diff
@@ -1,45 +1,55 @@
 {
   "name": "entry-extension-collapsible",
-  "version": "0.3.7",
+  "version": "0.1.0",
```
alexdonets, line 3: Question: should I manually set the old version here, pretend to do a major/minor update to bump it to 1.0.0/0.4.0 or keep the new version (0.1.0)?

extensions/entry-extension-collapsible/.env
```diff
@@ -0,0 +1,5 @@
+CONTENTFUL_ACCESS_TOKEN=3-Wv8MJDAXzQAE7jQHGLEkABrNlx9YFwwL9SWqUvLQ8
```
kdamball, line 1: Might want to remove this file and reset tokens?

extensions/entry-extension-collapsible/package.json
kdamball: I would bump it as it looks odd to go backwards 🤷🏾‍♂️

extensions/entry-extension-collapsible/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
kdamball, line 1: Add to `.gitignore`?

extensions/entry-extension-collapsible/README.md
```diff
@@ -4,27 +4,37 @@ This extension demonstrates how you would use all existing field editors to crea
 
 ![demo](./docs/demo.gif)
 
-## Commands
+## Available Scripts
```
kdamball, line 7: curious if the changes here are relevant? They seem very CCA based. Plus we say `npm` below but there's also a yarn file?

kdamball (commented):
Good one 🎉 . Left a bunch of questions and a minor clean up.

extensions/entry-extension-collapsible/src/react-app-env.d.ts
alexdonets: After googling seems like this file causes issues sometimes (although I didn't have any in this case), so make sense to put it in `.gitignore`. Will do it!

extensions/entry-extension-collapsible/README.md
alexdonets: I kept the original README instruction from CCA + added the description text for the extension from the old README. So the commands are relevant I think, except you are right that we should `yarn` there instead (and also description for `upload` script is missing).

Do you think adding description for `upload` and changing the commands to use `yarn` instead of `npm` would solve the issue? :)

alexdonets: Updated some things here :)

alexdonets:
Thanks for the review @kdamball! Made updates relevant to your comments. 🚀

kdamball (approved):
🎉

PR #750, opened 2021-06-22
chore: Migrate extensions into CCA (markdown-extension) [EXT-2566]

alexdonets:
#### Part 2/5 - `markdown-extension`

This PR is the 2nd of 5 PRs which implement the migration of the extensions from CCE into CCA.
In this PR we are migrating the `markdown-extension`.

#### Extensions migrated to CCA
- [x] `entry-extension-collapsible`
- [ ] `markdown-extension`
- [ ] `multiple-references-extension`
- [ ] `rich-text-extension`
- [ ] `singleline-extension`


extensions/markdown-extension/package.json
```diff
@@ -1,42 +1,51 @@
...
   "scripts": {
-    "start": "contentful-extension-scripts start",
-    "build": "contentful-extension-scripts build",
-    "deploy": "npm run build && contentful extension update --force",
-    "configure": "contentful space use && contentful space environment use",
-    "login": "contentful login",
-    "logout": "contentful logout",
-    "help": "contentful-extension-scripts help"
+    "start": "SKIP_PREFLIGHT_CHECK=true cross-env BROWSER=none react-scripts start",
+    "build": "SKIP_PREFLIGHT_CHECK=true GENERATE_SOURCEMAP=false react-scripts build",
```
shikaan, line 17: nit: those env vars can be put into the .env file, which is cross platform compatible. Alternatives, you should move them after `cross-env`

shikaan (approved):
LGTM: just a small nit for cross compatibility

extensions/markdown-extension/package.json
alexdonets: We decided to remove the .env in my first PR and put those inside the scripts for readability, plus they are specific to those 2 scripts. I'll move `SKIP_PREFLIGHT_CHECK=true` after `cross-env` if that makes sense?

PR #753, opened 2021-06-22
feat: add paragraph for slatejs next []

cezarsmpio:
Adds basic paragraph styling support and fix a few issues I've found.

### DIFF
--- a/packages/rich-text/src/RichTextEditor.styles.ts
+++ b/packages/rich-text/src/RichTextEditor.styles.ts
@@ -28,6 +28,10 @@ export const styles = {
         },
       },
     },
+    // We need to reset LIC style due to conflicts between PARAGRAPH styles
+    'ul > li > div': {
+      margin: 0,
+    },
   }),
   hiddenToolbar: css({
     borderTop: STYLE_EDITOR_BORDER,
--- a/packages/rich-text/src/plugins/Heading/index.tsx
+++ b/packages/rich-text/src/plugins/Heading/index.tsx
@@ -138,7 +138,7 @@ export function withHeadingEvents(editor: SPEditor) {
     if (isMod && isAltOrOption && headingKey) {
       event.preventDefault();
 
-      toggleNodeType(editor, { activeType: headingKey });
+      toggleNodeType(editor, { activeType: headingKey, inactiveType: BLOCKS.PARAGRAPH });
     }
   };
 }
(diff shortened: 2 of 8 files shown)


packages/rich-text/src/RichTextEditor.styles.ts
```diff
@@ -28,6 +28,10 @@ export const styles = {
         },
       },
     },
+    // We need to reset LIC style due to conflicts between PARAGRAPH styles
+    'ul > li > div': {
+      margin: 0,
+    },
```
cezarsmpio, line 34: I can explain this better. I think it's how this slate-plugins lib works internally. So when we map `ELEMENT_LIC` to type `BLOCKS.PARAGRAPH`, this will actually use what `BLOCKS.PARAGRAPH` is using as a component. This paragraph component has some margin styles. The `component` property of `ELEMENT_LIC` is completely ignored.

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -85,8 +85,6 @@ export const withListOptions: CustomSlatePluginOptions = {
   // ELEMENT_LIC is a child of li, slatejs does ul > li > lic + ul
   [ELEMENT_LIC]: {
     type: BLOCKS.PARAGRAPH,
-    // TODO: We might want to change it to our paragraph in case we develop one, otherwise we can simply use the slatejs default element
-    component: Slate.DefaultElement,
```
cezarsmpio, line 89: The `component` property is ignored because of this: https://github.com/contentful/field-editors/pull/753/files#diff-f887640d5b7f73e4a4d28d39eeea93db01913e49a8a3c1ea12ab7349af480fb8R40

packages/rich-text/src/plugins/Heading/index.tsx
```diff
@@ -142,7 +138,7 @@ export function withHeadingEvents(editor: SPEditor) {
     if (isMod && isAltOrOption && headingKey) {
       event.preventDefault();
 
-      toggleBlock(editor, headingKey);
+      toggleNodeType(editor, { activeType: headingKey, inactiveType: BLOCKS.PARAGRAPH });
```
cribbles, line 141: Not familiar with what this is doing - could you explain?

cribbles (approved):
LGTM but I was unsure about the `activeType` / `inactiveType` options use

PR #754, opened 2021-06-22
feat: add table markup

cribbles:
image

### DIFF
--- a/cypress/integration/RichTextEditor.spec.ts
+++ b/cypress/integration/RichTextEditor.spec.ts
@@ -540,4 +540,230 @@ describe('Rich Text Editor', () => {
       expectRichTextFieldValue(expectedValue);
     });
   });
+
+  describe('Tables', () => { 
+    const buildHelper = (type) => (...children) => block(type, {}, ...children);
+    const table = buildHelper(BLOCKS.TABLE);
+    const row = buildHelper(BLOCKS.TABLE_ROW);
+    const cell = buildHelper(BLOCKS.TABLE_CELL);
+    const paragraph = buildHelper(BLOCKS.PARAGRAPH);
+    const paragraphWithText = (t) => paragraph(text(t, []));
+    const emptyParagraph = () => paragraphWithText('');
+    const emptyCell = () => cell(emptyParagraph());
+    const cellWithText = (t) => cell(paragraphWithText(t));
+    const insertTableWithExampleData = () => {
+      editor()
+        .type(`{${mod}}{,}`)
+        .typeInSlate('foo')
+        .type('{rightArrow}')
+        .typeInSlate('bar')
+        .type('{rightArrow}')
+        .typeInSlate('baz')
+        .type('{rightArrow}')
+        .typeInSlate('quux')
+    };
+    const expectDocumentStructure = (...elements) => {
+      cy.wait(100);
+      expectRichTextFieldValue(
+        doc(
+          emptyParagraph(),
+          ...elements,
+          emptyParagraph(),
+        )
+      );
+    }
+    const expectTable = (...tableElements) => expectDocumentStructure(
+      table(...tableElements)
+    );
+    const expectTableToBeDeleted = () => expectDocumentStructure();
+
+    it('allows creating a table', () => {
+      editor().type(`{${mod}}{,}`);
+
+      expectTable(
+        row(
+          emptyCell(),
+          emptyCell(),
+        ),
+        row(
+          emptyCell(),
+          emptyCell(),
+        )
+      );
+    });
+
+    it('allows filling a table with text', () => {
+      insertTableWithExampleData();
+
+      expectTable(
+        row(
+          cellWithText('foo'),
+          cellWithText('bar'),
+        ),
+        row(
+          cellWithText('baz'),
+          cellWithText('quux'),
+        )
+      );
+    });
+
+    it('allows adding rows and columns', () => {
+      insertTableWithExampleData();
+
+      editor()
+        .click()
+        .type(`{upArrow}`)
+        .type(`{upArrow}`)
+        .wait(100)
+        .type(`{${mod}}{rightArrow}`);
+
+      expectTable(
+        row(
+          cellWithText('foo'),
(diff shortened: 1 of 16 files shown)


packages/json/src/JsonEditoField.tsx
```diff
@@ -12,7 +12,7 @@ const CODE_MIRROR_CONFIG = {
   indentWithTabs: true,
   height: 'auto',
   theme: 'none',
-  autoRefresh: true
+  autoRefresh: true,
```
cribbles, line 15: This is from running `prettier`.

packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -30,7 +30,7 @@ function onLinkOrCreate(
   entityType: ChildProps['entityType'],
   items: ChildProps['items'],
   ids: string[],
-  index = items.length,
+  index = items.length
```
cribbles, line 33: This is from running `prettier`.

packages/reference/src/components/LinkActions/LinkActions.tsx
```diff
@@ -27,7 +27,7 @@ export interface LinkActionsProps {
   onLinkedExisting: (entities: Array<Entry | Asset>, index?: number) => void;
   actionLabels?: Partial<ActionLabels>;
   combinedActionsLabel?: string | React.ReactElement;
-  itemsLength?: number
+  itemsLength?: number;
```
cribbles, line 30: This is from running `prettier`. (so are the changes below in this file)

packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -27,7 +27,15 @@ type LinkEntityActionsProps = {
...
-  const { sdk, editorPermissions, entityType, canLinkMultiple, isDisabled, actionLabels, itemsLength } = props;
+  const {
+    sdk,
+    editorPermissions,
+    entityType,
+    canLinkMultiple,
+    isDisabled,
+    actionLabels,
+    itemsLength,
+  } = props;
```
cribbles, line 38: More `prettier`

packages/rich-text/src/helpers/editor.ts
```diff
@@ -27,6 +28,20 @@ export function hasSelectionText(editor) {
...
+  nodeType: BLOCKS | INLINES
+): NodeEntry | [] {
+  if (!editor.selection) return [];
+  const { path } = editor.selection.focus;
+  for (let i = 0; i < path.length; i++) {
+    const nodeEntry = Editor.node(editor, path.slice(0, i + 1)) as NodeEntry;
+    if (nodeEntry[0].type === nodeType) return nodeEntry;
+  }
+  return []
+}
```
cribbles, line 43: This is a slight perf bump over udecode's `getNodes` that passes us the `[Node, Path]` tuple direct from Slate's `Editor` interface, which we need as often as not. I put it in `helpers` because we use it for a slight improvement to HRs later in this file.

packages/rich-text/src/helpers/editor.ts
```diff
@@ -56,7 +71,11 @@ export function toggleBlock(editor, type: string): void {
   Transforms.setNodes(editor, newProperties);
 
   if (!isActive && (isList || isQuote)) {
-    const block = { type, children: [] };
+    const block = {
+      type,
+      data: {},
+      children: []
+    };
```
cribbles, line 78: This is one of many places in this PR where the compiler was sad because a `block` was missing a `data` attribute. I don't know what I did to make that a problem for this PR only, but if you see something similar added below that's probably why.

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -55,19 +58,17 @@ interface ToolbarHrButtonProps {
...
+    const [, pathToSelectedHr] = getNodeEntryFromSelection(editor, BLOCKS.HR);
+
+    if (pathToSelectedHr) {
+      if (isEnter) {
+        event.preventDefault();
+        moveToTheNextLine(editor);
+      } else if (event.key === 'Backspace') {
+        event.preventDefault();
+        Transforms.removeNodes(editor, { at: pathToSelectedHr });
+      }
```
cribbles, line 71: This is more than a refactor, the second conditional lets us delete a selected HR (which we couldn't before, we could only delete an HR by backspacing from the paragraph after it). This felt weird but more importantly it's inconsistent with behavior in the legacy editor.

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -80,20 +81,22 @@ export function ToolbarHrButton(props: ToolbarHrButtonProps) {
...
       children: [{ text: '' }],
     };
 
     const hasText = !!getText(editor, editor.selection.focus.path);
 
-    hasText ? insertNodes(editor, hr) : setNodes(editor, hr);
+    hasText ? Transforms.insertNodes(editor, hr) : setNodes(editor, hr);
 
-    insertNodes(editor, paragraph);
+    Transforms.insertNodes(editor, paragraph);
```
cribbles, line 99: TypeScript got mad at me for not using Slate's `Transforms.insertNodes` after the refactor but trust me that it's doing the same thing as udecode's implementation was before.

cypress/integration/RichTextEditor.spec.ts
```diff
@@ -540,4 +540,232 @@ describe('Rich Text Editor', () => {
...
+    const expectTable = (...tableElements) => {
+      expectRichTextFieldValue(
+        doc(
+          emptyParagraph(),
+          table(...tableElements),
+          emptyParagraph(),
+        )
+      );
+    };
+    const expectTableToBeDeleted = () => {
```
cezarsmpio, line 573: You can try adding `cy.wait(100)` before assertions.

cezarsmpio (approved):

DanweDE (approved):
Didn't have a super in-depth look, but :+1:

taye (approved):

PR #767, opened 2021-07-02
feat: add hyperlink (only adding regular hyperlinks)

cezarsmpio:
Modal
image
image

Editor
image
image

### DIFF
--- a/packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
+++ b/packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
@@ -19,7 +19,7 @@ interface HyperlinkModalProps {
   linkText?: string;
   linkType?: string;
   linkTarget?: string;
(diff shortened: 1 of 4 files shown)


packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { Editor, Transforms } from 'slate';
+import { ReactEditor } from 'slate-react';
+import { HistoryEditor } from 'slate-history';
+import { insertLink } from '../../helpers/editor';
+
+interface HyperlinkModalProps {
+  linkText?: string;
+  linkType?: string;
+  linkTarget?: string;
+  onClose: Function;
```
taye, line 22: very small nit: it's better to use more specific types than `Function`

taye (approved):

PR #768, opened 2021-07-02
feat: add support for entry and asset hyperlinks

cezarsmpio:
# Entry

image
image

# Asset

image
image

### DIFF
--- a/packages/rich-text/src/helpers/editor.ts
+++ b/packages/rich-text/src/helpers/editor.ts
@@ -1,11 +1,11 @@
 import { Text, Editor, Element, Transforms, Path, Range } from 'slate';
 import { BLOCKS, INLINES } from '@contentful/rich-text-types';
-import { CustomElement } from '../types';
-import { SPEditor } from '@udecode/slate-plugins-core';
+import { CustomElement, CustomEditor } from '../types';
 
 const LIST_TYPES: string[] = [BLOCKS.OL_LIST, BLOCKS.UL_LIST];
+const LINK_TYPES: string[] = [INLINES.HYPERLINK, INLINES.ASSET_HYPERLINK, INLINES.ENTRY_HYPERLINK];
 
-export function isBlockSelected(editor, type: string): boolean {
+export function isBlockSelected(editor: CustomEditor, type: string): boolean {
   const [match] = Array.from(
     Editor.nodes(editor, {
       match: (node) => Element.isElement(node) && (node as CustomElement).type === type,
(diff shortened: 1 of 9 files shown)


taye (approved):

packages/rich-text/src/helpers/editor.ts
```diff
@@ -124,25 +129,31 @@ export function unwrapLink(editor) {
...
-    data: {
-      uri: url,
-    },
+    type,
+    data: {},
     children: isCollapsed ? [{ text }] : [],
   };
 
+  if (url) {
+    link.data = { url };
```
cribbles, line 150: It's okay that this is `url` and not `uri: url`?

packages/rich-text/src/helpers/editor.ts
```diff
@@ -124,25 +129,31 @@ export function unwrapLink(editor) {
...
+  }
+
+  if (target) {
+    link.data = { target };
+  }
```
cribbles, line 155: Would we face any issues with a scenario where `url` and `target` are both present?

packages/rich-text/src/plugins/Hr/index.tsx
```diff
@@ -101,6 +102,8 @@ export function ToolbarHrButton(props: ToolbarHrButtonProps) {
     Slate.ReactEditor.focus(editor);
   }
 
+  if (!editor) return null;
```
cribbles, line 105: What are these changes for?

packages/rich-text/src/helpers/editor.ts
cezarsmpio: Good catch!

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -30,6 +30,8 @@ export function ToolbarListButton(props: ToolbarListButtonProps) {
     Slate.ReactEditor.focus(editor);
   }
 
+  if (!editor) return null;
```
cribbles, line 33: same here, what does this fix?

packages/rich-text/src/plugins/Hr/index.tsx
cezarsmpio: Lint was accusing some errors when I was trying to improve the internal type. `editor` might be undefined. We also used the same pattern on other buttons but forgot to add on some. I can remove it if that's causing noise and confusion.

packages/rich-text/src/plugins/List/index.tsx
cezarsmpio: Answer above :)

packages/rich-text/src/plugins/Hr/index.tsx
cribbles: 👍

cribbles: (👍 to keeping it, not removing)

packages/rich-text/src/helpers/editor.ts
cezarsmpio: I don't think we can both at the same type. Nevertheless, `type` will dictate what to render, and the render will either use `target` or `uri`.

PR #769, opened 2021-07-05
feat: update to app-sdk@v4

djagya:
Bump the `app-sdk` version to v4.

BREAKING CHANGE: new SDK API.

### DIFF
--- a/apps/entry-app-collapsible/package.json
+++ b/apps/entry-app-collapsible/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
-    "@contentful/app-sdk": "^3.36.0",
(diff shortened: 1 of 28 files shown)


suevalov (commented):
Do we want to use the alpha version or wait for a stable v4 before upgrading? 

What's the reason to get it updated now?

djagya:
@suevalov The main web app already uses alpha v4, our experience components are also updated to v4. Now a couple of experience components are already bumped in Compose, and it makes sense to update the Compose itself. Then the only repo using v3 sdk is field-editors, which causes some type problems. Maybe I can approach it somehow else, avoiding releasing the new field-editors version with alpha sdk.

djagya:
Closing the PR for now.

PR #772, opened 2021-07-06
fix: orphaned text nodes

cribbles:
Ensures text nodes never get added to the document structure in a way that clashes with the validation layer.

### DIFF
--- a/packages/rich-text/src/helpers/sanitizeSlateDoc.ts
+++ b/packages/rich-text/src/helpers/sanitizeSlateDoc.ts
@@ -31,15 +31,6 @@ const wrapNode = (
   children: [node],
 });
 
-/**
- * Ensures "orphaned" text node elements (those without a parent capable of
- * validly rendering a text child node) are wrapped with a suitable parent
- * element to prevent failures on the validation layer.
- *
- * It is commonplace for third party plugins (including udecode) to reconcile
- * deletion events by inserting such nodes into their schema. This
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/helpers/sanitizeSlateDoc.ts
```diff
@@ -0,0 +1,77 @@
...
+  nodes: TextOrCustomElement[] = [],
+  parentNodeType: NodeType = BLOCKS.DOCUMENT,
+): TextOrCustomElement[] {
+  return nodes.map((node: TextOrCustomElement): TextOrCustomElement => {
+    if (TEXT_PARENT_NODES.includes(parentNodeType) && isTextElement(node)) {
+      return node;
+    }
+    if (isTextElement(node)) {
+      return wrapOrphanedTextNode(parentNodeType, node);
+    }
```
djagya, line 71: Maybe just a biiit clearer with one condition for the text element. But feel free to ignore

```suggestion
    if (isTextElement(node)) {
      return TEXT_PARENT_NODES.includes(parentNodeType) 
        ? node
        : wrapOrphanedTextNode(parentNodeType, node);
    }
```

djagya (approved):

packages/rich-text/src/helpers/sanitizeSlateDoc.ts
cribbles: This is better, I'll commit it 👍

PR #773, opened 2021-07-08
feat: disable table block insertion

cribbles:
Disables the toolbar button and keyboard shortcuts for block element insertion within tables.

### DIFF
--- a/packages/rich-text/src/helpers/editor.ts
+++ b/packages/rich-text/src/helpers/editor.ts
@@ -42,6 +42,15 @@ export function getNodeEntryFromSelection(
   return [];
 }
 
+export function isNodeTypeSelected(
+  editor: SPEditor,
+  nodeType: BLOCKS | INLINES
+): boolean {
+  if (!editor) return false;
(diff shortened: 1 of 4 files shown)


packages/rich-text/src/helpers/editor.ts
```diff
@@ -42,6 +42,15 @@ export function getNodeEntryFromSelection(
   return [];
 }
 
+export function isNodeTypeSelected(
+  editor: SPEditor,
+  nodeType: BLOCKS | INLINES
+): boolean {
+  if (!editor) return false;
+  const [table] = getNodeEntryFromSelection(editor, nodeType);
```
djagya, line 50: Not sure why the variable name is `table` here? A bit confusing, maybe `node` or `entry`?

djagya (approved):

djagya:
Nice cypress test 👍

packages/rich-text/src/helpers/editor.ts
cribbles: Ha, good catch 👍

PR #774, opened 2021-07-09
feat: edit hyperlinks

cribbles:
follows https://github.com/contentful/field-editors/pull/768

Allows editing hyperlinks, including changing a hyperlink to a different type of hyperlink (e.g. URL -> Entity and vice versa).

Plays well with tables

### DIFF
--- a/cypress/integration/RichTextEditor.spec.ts
+++ b/cypress/integration/RichTextEditor.spec.ts
@@ -1,5 +1,5 @@
-import { MARKS, BLOCKS } from '@contentful/rich-text-types';
-import { document as doc, block, text } from '../../packages/rich-text/src/helpers/nodeFactory';
+import { MARKS, BLOCKS, INLINES } from '@contentful/rich-text-types';
+import { document as doc, block, inline, text } from '../../packages/rich-text/src/helpers/nodeFactory';
 
 function expectRichTextFieldValue(expectedValue, editorEvents?) {
   cy.getRichTextField().then((field) => {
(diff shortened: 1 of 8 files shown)


packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
```diff
@@ -234,16 +228,29 @@ interface HyperLinkDialogData {
...
 
+  let linkType;
+  let linkText;
+  let linkTarget;
+  let linkEntity;
+
+  const [node, path] = getNodeEntryFromSelection(editor, LINK_TYPES);
+  if (node) {
+    linkType = node.type;
+    linkText = (node.children[0] as TextElement).text;
```
cezarsmpio, line 243: I guess we can use `Editor.string(editor, path)` to extract the text from the selection, I think it's safer.

packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
```diff
@@ -234,16 +228,29 @@ interface HyperLinkDialogData {
...
+    linkTarget = (node.data as { uri: string }).uri || '';
+    linkEntity = (node.data as { target: Link }).target;
+  }
+
   const selectionBeforeBlur = { ...editor.selection };
-  const currentLinkText = linkText ?? Editor.string(editor, editor.selection);
+  const currentLinkText = linkText || Editor.string(editor, editor.selection);
```
cezarsmpio, line 249: This is confusing me with line 243. I think we might not need both, what I mean is, I think we can simply use `Editor.string(editor, editor.selection)` for both because `editor.selection` should return what's selected, regardless of which component is selected. In this case, we only need the text, regardless of the action (edit or insert).

cezarsmpio (commented):
It looks good overall. Just have one question.

packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
cribbles: We're capturing a couple different cases here.

1) *_Create_ case:* the link doesn't exist yet. As a user I want to highlight one small selected part of a text node and convert that into a link. For this I need the `editor.selection` so the whole node isn't selected.

2) *_Update_ case:* the link exists. As a user, I want to click on it to prompt the edit modal. I haven't actually _selected_ anything, I just clicked on the link. So I need the `path` of the clicked node to get the link text, since the `editor.selection` buffer will possibly be empty or unrelated to the link at hand.

cezarsmpio: Ah, that's true! You're right.

packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -57,13 +57,29 @@ const styles = {
...
+  };
+}
+
+export function buildHyperlinkEventHandler(sdk) {
+  return function withHyperlinkEvents(editor) {
+    return function handleKeyDown(event: KeyboardEvent) {
+      if (!editor.selection) return;
+      const isMod = event.ctrlKey || event.metaKey;
+      const K = 75;
+      if (isMod && event.keyCode === K) {
```
cezarsmpio, line 75: I think it would be better if we transform `event.keyCode === K` into its own variable, something like `isK`.

cezarsmpio (approved):
Great!

cypress/integration/RichTextEditor.spec.ts
```diff
@@ -808,4 +808,273 @@ describe('Rich Text Editor', () => {
...
+    const expectDocumentStructure = (...nodes) => {
+      expectRichTextFieldValue(
+        doc(
+          block(
+            BLOCKS.PARAGRAPH,
+            {},
+            ...nodes.map(([nodeType, ...content]) => {
+              if (nodeType === 'text') return text(...content);
+              const [data, textContent] = content;
+              return inline(nodeType, data, text(textContent));
```
cribbles, line 827: I've updated this test to account for the new document structure in which text nodes are direct descendants of hyperlinks (rather than having a `hyperlink` -> `paragraph` -> `text` hierarchy).

packages/rich-text/src/helpers/sanitizeSlateDoc.ts
```diff
@@ -1,22 +1,21 @@
...
-    BLOCKS.PARAGRAPH,
-    BLOCKS.HEADING_1,
-    BLOCKS.HEADING_2,
-    BLOCKS.HEADING_3,
-    BLOCKS.HEADING_4,
-    BLOCKS.HEADING_5,
-    BLOCKS.HEADING_6,
-  );
+  (VOID_BLOCKS as string[])
+    .concat(values(INLINES) as string[])
```
cribbles, line 9: I had to change this to make hyperlinks pass validation. The previous syntax was simply wrong. This results in hyperlinks properly rendering text nodes as direct children, as noted above.

packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
```diff
@@ -153,36 +161,33 @@ export function HyperlinkModal(props: HyperlinkModalProps) {
...
-                        assetId={linkEntity.sys.id}
-                        isDisabled={true}
-                        isSelected={false}
-                      />
-                    )}
-                  </div>
-                )}
-
-                {!linkEntity && (
+                {linkEntity && linkEntity.sys.linkType === SYS_LINK_TYPES[linkType] ? (
```
cribbles, line 164: This ternary has been adjusted so that the asset -> entry -> asset switching logic in the modal "just works." Before, it would simply see that there was a `linkEntity` when updating and try to render the entity regardless of whether an entry or an asset was being linked. This would result in the "entry missing or unaccessible display."

packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
```diff
@@ -203,7 +208,7 @@ export function HyperlinkModal(props: HyperlinkModalProps) {
             disabled={!isLinkComplete()}
             onClick={handleOnSubmit}
             testId="confirm-cta">
-            Insert
+            {props.linkType ? 'Update' : 'Insert'}
```
cribbles, line 211: A final touch from the legacy editor that had been missing.

packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -57,13 +57,37 @@ const styles = {
...
+
+const isMod = (event: KeyboardEvent): event is ModEvent => event.ctrlKey || event.metaKey;
+const isK = (event: KeyboardEvent): event is KEvent => event.keyCode === 75;
+const wasHyperlinkEventTriggered = (event: KeyboardEvent): event is HyperlinkEvent =>
+  isMod(event) && isK(event);
+
+export function buildHyperlinkEventHandler(sdk) {
+  return function withHyperlinkEvents(editor) {
+    return function handleKeyDown(event: KeyboardEvent) {
+      if (!editor.selection || !wasHyperlinkEventTriggered(event)) return;
```
cribbles, line 84: Simplified based on the comment from a previous review.

cribbles:
@cezarsmpio , I'm reassigning you since I've made quite a few small (but important) changes since you last saw this one and they have been unfortunately squashed into the last commit - hence I've annotated the changes for visibility.

cezarsmpio (approved):
👍

PR #775, opened 2021-07-12
feat: fix non-unique checkbox labels for multiple editors

djagya:
Fixes non-unique labels of multiple checkbox editors, which was causing incorrect behavior when clicking on those labels.

### DIFF
--- a/packages/checkbox/src/CheckboxEditor.tsx
+++ b/packages/checkbox/src/CheckboxEditor.tsx
@@ -10,6 +10,7 @@ import {
 import { CheckboxField, Form, TextLink } from '@contentful/forma-36-react-components';
 import * as styles from './styles';
 import { nanoid } from 'nanoid';
+import { useMemo } from 'react';
 
 export interface CheckboxEditorProps {
   /**
@@ -78,7 +79,7 @@ const getInvalidValues = (
 export function CheckboxEditor(props: CheckboxEditorProps) {
   const { field, locales } = props;
 
-  const options = getOptions(field);
+  const options = useMemo(() => getOptions(field), [field]);
   const misconfigured = options.length === 0;
 
   if (misconfigured) {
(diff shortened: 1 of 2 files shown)


ChidinmaOrajiaku:
Thanks, @djagya for the quick fix 😊

packages/checkbox/src/CheckboxEditor.tsx
```diff
@@ -44,12 +45,13 @@ export function getOptions(field: FieldAPI): CheckboxOption[] {
...
-    ? predefinedValues[0]
-    : []) as string[];
+  const firstPredefinedValues = (
+    predefinedValues.length > 0 ? predefinedValues[0] : []
+  ) as string[];
 
   return firstPredefinedValues.map((value: string, index) => ({
-    id: ['entity', field.id, field.locale, index].join('.'),
+    // Append a random id to distinguish between checkboxes opened in two editors (e.g. slide-in)
+    id: ['entity', field.id, field.locale, index, nanoid(6)].join('.'),
```
DanweDE, line 54: `id` will become the element's `key`, right? I believe random keys should be avoided in React.
How about keeping a `const` that increases with each new component instance here instead?

Let's also

```suggestion
    id: ['entity', field.id, field.locale, index, nanoid(6)].join(':'),
```

since `.` [is an allowed resource ID character](https://www.contentful.com/developers/docs/references/content-management-api/#/introduction/authentication).

DanweDE (changes requested):

packages/checkbox/src/CheckboxEditor.tsx
djagya: Hmm, not sure how well it goes with react. We have to increment the `instanceCount` in `useEffect`, by the time when effect is triggered, the initial list is already rendered. When we render two editors at the same time, they both will have the same `instanceCount` in their options.

Preferably I'd keep using `nanoid` (we also use it in the boolean editor, same technique, https://github.com/contentful/field-editors/blob/master/packages/boolean/src/BooleanEditor.tsx#L32-L35), but will memoize the generated options.

DanweDE (approved):
LGTM :+1:

PR #777, opened 2021-07-21
feat: support Table Actions (add/delete rows/columns/tables)

z0al:
This PR adds support for the following Table actions:

* Add row above
* Add row below
* Add column left
* Add column right
* Delete row
* Delete column
* Delete table

rt-table-demo


packages/rich-text/src/plugins/Table/TableActions.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const styles = {
+  topRight: css({
+    position: 'absolute',
+    top: tokens.spacingXs,
+    right: tokens.spacingXs,
+  }),
+};
+
+type TableAction = (e: SPEditor, o: TablePluginOptions) => void;
```
cezarsmpio, line 30: Could we use more specific names for `e` and `o`? Not sure what's the best practice but I think it might not be intuitive for future maintenance.

packages/rich-text/src/plugins/Table/helpers.ts
```diff
@@ -175,4 +183,9 @@ function insertTableWithTrailingParagraph(editor: SPEditor, withTableOptions: Cu
     children: [{ text: '' }],
   };
   Transforms.insertNodes(editor, paragraph, { at: insertPath });
-};
+}
+
+export function isTableActive(editor: SPEditor) {
+  const tableElements = [ELEMENT_TABLE, ELEMENT_TH, ELEMENT_TR, ELEMENT_TD];
+  return tableElements.some((el) => isBlockSelected(editor, el));
+}
```
cezarsmpio, line 191: Clean 👍

packages/rich-text/src/plugins/Table/index.tsx
```diff
@@ -8,11 +8,16 @@ import {
   ELEMENT_TABLE,
   createTablePlugin as createTablePluginFromUdecode,
   getTableOnKeyDown,
+  // getTableCellEntry,
```
cezarsmpio, line 11: Should we remove it?

packages/rich-text/src/plugins/Table/useAnchorNode.ts
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
cezarsmpio, line 1: Don't we need to do `import * as React from 'react';` in order to use `React.useEffect`?

More a curiosity from my side :)

packages/rich-text/src/plugins/Table/index.tsx
```diff
@@ -52,15 +58,35 @@ export const TR = (props: Slate.RenderElementProps) => (
...
-    {...props.attributes}
-    // may include `colspan` and/or `rowspan`
-    {...(props.element.data as TableCell['data'])}
-    className={styles[BLOCKS.TABLE_CELL]}>
-    {props.children}
-  </td>
-);
+export const TD = (props: Slate.RenderElementProps) => {
+  const editor = useStoreEditor();
+  const selectedNode = useAnchorNode();
```
cezarsmpio, line 63: Why do we need this for? Can't we use some utils from SlateJS?

packages/rich-text/src/plugins/Table/index.tsx
z0al: Done

packages/rich-text/src/plugins/Table/useAnchorNode.ts
z0al: No need for `import * as React` really. I think that was a problem a few years back but not anymore. Can't remember exactly whether it's the `esModuleInterop` flag in `tsconfig.json` that does the trick or something else.

Anyway, I changed the import to `{ useState, useEffect }` since I don't need `React` anyway.

packages/rich-text/src/plugins/Table/index.tsx
z0al: Are you referring to `useAnchorNode` hook? that's just a helper to get the current _HTML_ Node under the cursor. If you have a better way to detect the focused cell let me know. Added a little comment on how it works.

> Can't we use some utils from SlateJS?

I'm not aware of any but just be clear. the Node here is an HTML node, not Slate's :)

packages/rich-text/src/plugins/Table/TableActions.tsx
z0al: I tend to not worry about naming, in short, unexported, _types_ like this. Especially when the arg types tell exactly what to expect (e.g. `SPEditor` and `TablePluginOptions` in this case)

Regardless, I changed them to `editor` and `options`.

packages/rich-text/src/plugins/Table/useAnchorNode.ts
z0al: Btw, looks like `import * as React` is [more future proof](https://twitter.com/dan_abramov/status/1308739731551858689?s=20) than `import React`

z0al: On the other hand, we don't need to import `React` anyway when migrate to React v17 :)

packages/rich-text/src/plugins/Table/index.tsx
z0al: Oh, now I know what you mean 😄 . There is `Slate.useSelect()` 🤦🏻‍♂️

cezarsmpio (approved):

cezarsmpio (approved):
Well done!

PR #779, opened 2021-07-22
feat: add embedded entity block

cribbles:
Adds support for embedded entity and asset block plugins.


packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -63,6 +69,24 @@ const Toolbar = ({ isDisabled }: ToolbarProps) => {
...
         <ToolbarHrButton isDisabled={isDisabled || !canInsertBlocks} />
       </div>
+      <div className={styles.embedActionsWrapper}>
+        <EmbeddedEntityDropdownButton
+          isDisabled={isDisabled}
+          onClose={onCloseEntityDropdown}
+          onToggle={onToggleEntityDropdown}
+          isOpen={isEmbedDropdownOpen}>
+          <EmbeddedEntityBlockToolbarIcon
+            isDisabled={!!isDisabled}
```
cezarsmpio, line 79: I think we don't need the double `!!`, `isDisabled` is already a boolean.

packages/rich-text/src/helpers/getLinkedContentTypeIdsForNodeType.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable you-dont-need-lodash-underscore/find */
```
cezarsmpio, line 1: We are using this function on `HyperlinkModal.tsx` already. We need to remove it from there too.

cezarsmpio (approved):

cezarsmpio (approved):

PR #780, opened 2021-07-23
feat: add serializer plugin for marks

cezarsmpio:
~~**Cypress tests are coming.**~~

Cypress doesn't support really well copying & pasting so we decided to skip it. Unit tests were also taken into consideration but we decided not to do it as well because we would be testing SlateJS functions only.

**Google Docs**
image

**Result**
image


djagya (approved):
👍

PR #785, opened 2021-07-28
fix: render dropdown list or fragment

cribbles:
Fixes https://github.com/contentful/field-editors/pull/784

For reasons not obvious to me, `false` actually causes this to fail with

```
TypeError: Cannot read property 'type' of null
```

I confess I am not sure why this solution works (ternary to default `undefined` or anything else falsy has the same problem), but it's low cost and accomplishes basically the same thing with the same DOM markup in the end. If anyone has any further suggestions / thoughts, I am open.


cezarsmpio (approved):

taye:
This seems to be the cause https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Card/CardActions/CardActions.tsx#L96

cribbles:
@taye awesome, fished around than way longer than I should've and couldn't find that

PR #786, opened 2021-07-28
feat: add embedded entity inline

cribbles:
- [x] adds embedded entity inline elements
  - keyboard shortcuts
  - cypress tests

Also
- [x] fixes keyboard shortcut for asset inlines
- [x] fixes "move to top" / "move to bottom" incorrectly appearing on embedded entry cards

### DIFF
--- a/cypress/integration/RichTextEditor.spec.ts
+++ b/cypress/integration/RichTextEditor.spec.ts
@@ -1145,17 +1145,13 @@ describe('Rich Text Editor', () => {
           cy.wait(100);
         },
       ],
-
-      // TODO:
-      // This works in the browser, but not in Cypress. Maybe upgrading will help.
-      // For now, it's okay to ensure addition via the toolbar button works.
-      // [
-      //   'using the keyboard shortcut',
-      //   () => {
-      //     editor().type(`{${mod}}{shift}a`);
-      //     cy.wait(100);
-      //   }
-      // ]
+      [
+        'using the keyboard shortcut',
+        () => {
+          editor().type(`{${mod}}{shift}a`);
+          cy.wait(100);
+        }
+      ]
     ];
 
     for (const [triggerMethod, triggerEmbeddedAsset] of methods) {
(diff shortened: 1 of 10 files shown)


cypress/integration/RichTextEditor.spec.ts
```diff
@@ -1228,4 +1224,77 @@ describe('Rich Text Editor', () => {
...
+    ];
+
+    for (const [triggerMethod, triggerEmbeddedAsset] of methods) {
+      describe(triggerMethod, () => {
+        it('adds and removes embedded entries', () => {
+          editor().click().typeInSlate('hello')
+          triggerEmbeddedAsset();
+          editor().click().typeInSlate('world')
+
+          cy.wait(500);
```
z0al, line 1254: Is there any other event/change we can wait for instead of the [magical `cy.wait(500)`](https://docs.cypress.io/api/commands/wait#Time) :) ? it slows our CI runs + It's just a matter of time before this test becomes flaky as well.

If there is no other option then maybe using [cy.waitUntil](https://github.com/NoriSte/cypress-wait-until) is slightly better since it doesn't have to always wait for `500ms`.

packages/rich-text/src/plugins/EmbeddedEntityBlock/LinkedEntityBlock.tsx
```diff
@@ -51,8 +52,6 @@ export function LinkedEntityBlock(props: LinkedEntityBlockProps) {
     Transforms.removeNodes(editor, { at: pathToElement });
   };
 
-  // TODO: fixme -- props not available on new editor
-  const isDisabled = false; // editor.props.readOnly || editor.props.actionsDisabled;
```
z0al, line 55: Out of curiosity, What did we use `actionsDisabled` for in the past?

packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const title = React.useMemo(
+    () =>
+      getEntryTitle({
+        entry,
+        contentType,
+        localeCode: props.sdk.field.locale,
+        defaultLocaleCode: props.sdk.locales.default,
+        defaultTitle: 'Untitled',
+      }),
+    [entry, contentType, props.sdk]
```
z0al, line 54: Nit: we could put `props.sdk.field.locale` and `props.sdk.locales.default` as deps instead of `props.sdk` here.

packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+  );
+
+  React.useEffect(() => {
+    if (!props.entryId) return;
+    getOrLoadEntry(props.entryId);
+  }, [props.entryId]); // eslint-disable-line
```
z0al, line 60: Can you leave a code comment on why it's not a good idea to add `getOrLoadEntry` to the list of deps?

z0al (commented):
LGTM. Left small comments

z0al (approved):
I realized all of my comments are small-ish so approving. I'd really like us to reduce the use of `cy.wait` if possible though.

packages/rich-text/src/plugins/EmbeddedEntityBlock/LinkedEntityBlock.tsx
cribbles: We use it for the `ReadOnlyRichTextEditor` in the UI diff view: https://github.com/contentful/user_interface/blob/09590e62f7ccf4058b10efad175573ceb1077532/src/javascripts/app/widgets/RichText/ReadOnlyRichTextEditor.js#L14-L25

But that's in coordination with `readOnly` and I can't remember if there was a practical rather than semantic reason for separating between these two.

cypress/integration/RichTextEditor.spec.ts
cribbles: Probably, but I think this is a larger issue we should address in the rest of our test suite. I'll add a ticket for it.

PR #790, opened 2021-07-30
fix: revert prevent default

cezarsmpio:
Reverts part of #784 

Only:

- doesn't impose `event.preventDefault()` on click events when the element is not clickable

This is making the embedded blocks being clickable on the rich text field.

A better solution should be made, for example, I think we don't need the `isClickable` at all, we can simply not pass the `entryUrl` or `getAssetUrl` property, but more time to test would be needed for this change.

### DIFF
--- a/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
@@ -176,13 +176,15 @@ export function WrappedEntryCard(props: WrappedEntryCardProps) {
                   Move to bottom
                 </DropdownListItem>
               </DropdownList>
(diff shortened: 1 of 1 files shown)


packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx
```diff
@@ -176,13 +176,15 @@ export function WrappedEntryCard(props: WrappedEntryCardProps) {
                   Move to bottom
                 </DropdownListItem>
               </DropdownList>
-            ) : <React.Fragment />}
+            ) : (
+              <React.Fragment />
+            )}
```
cezarsmpio, line 181: Prettier 🤷

cribbles (approved):

PR #792, opened 2021-08-03
fix: 🐛 search on dropdown losing focus after change

massao:
Fix issue with the search inside the dropdown losing focus after each change.

### DIFF
--- a/packages/reference/src/components/CreateEntryLinkButton/CreateEntryMenuTrigger.tsx
+++ b/packages/reference/src/components/CreateEntryLinkButton/CreateEntryMenuTrigger.tsx
@@ -110,6 +110,11 @@ export const CreateEntryMenuTrigger = ({
   const closeMenu = () => setOpen(false);
 
   useEffect(() => {
+    /*
+      This is a temporary workaround to keep the search input focused after the value is changed.
+      When the TextInput component is updated to use forwardRef we should update to pass the ref directly to it.
+      Since the input was being rerendered rather than updated when the value changed, this ref makes sure that the search field will keep the focus.
+    */
     textField?.current?.querySelector('input')?.focus()
   }, [searchInput])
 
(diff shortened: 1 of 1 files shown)


packages/reference/src/components/CreateEntryLinkButton/CreateEntryMenuTrigger.tsx
```diff
@@ -108,6 +109,10 @@ export const CreateEntryMenuTrigger = ({
 
   const closeMenu = () => setOpen(false);
 
+  useEffect(() => {
+    textField?.current?.querySelector('input')?.focus()
```
massao, line 113: The `querySelector` is a workaround as the `TextInput` component currently is a functional component without `forwardRef`, so we couldn't attach the ref directly on it. 😞

suevalov: wouldn't it focus and select all text every time? why this bug occurs at the first place? 

could you also leave a comment that this is a temporary workaround?

massao: It seems like whenever the `onChange` happens, the input field is rendered again as a new component and loses focus.
As for the `focus`, it keeps the last state of the element, so it should even keep the caret on the same place, it would select the entire text if it was using `select()`

Will also add this on the comment, to keep it clear why this useEffect is there too.

suevalov: can we use the autofocus property on the input element instead?

massao: I did try it with the `autofocus`, but it didn't work and also gives an eslint error, as autofocus is not recommended because of a11y

suevalov (approved):

djagya (approved):

Issue #491, opened 2020-11-23
Character count on the multiline long text field editor


amandahogan:
I'd like to request a feature for a character count to be included on the multiline long text field editor similar to the one on the markdown editor and short text field editor.

suevalov:
Hi @amandahogan, thanks for the feature request. We'll put it in our backlog. If you want it to happen faster, feel free to open a PR :)

amandahogan:
@suevalov just checking if this has already been worked on? If not I'll open a PR for it.

suevalov:
Hi @amandahogan, this is not being worked on at the moment. Feel free to open a PR 👍🏼

PR #795, opened 2021-08-05
feat: 🎸 add char counter

anho:
Showing the character count for multiline fields with opt-out possibility through additional prop.

✅ Closes: #491

### DIFF
--- a/packages/multiple-line/src/MultipleLineEditor.tsx
+++ b/packages/multiple-line/src/MultipleLineEditor.tsx
@@ -1,6 +1,13 @@
 import * as React from 'react';
 import { Textarea } from '@contentful/forma-36-react-components';
-import { FieldAPI, LocalesAPI, FieldConnector } from '@contentful/field-editor-shared';
+import {
+  FieldAPI,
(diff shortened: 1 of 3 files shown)


suevalov (approved):

packages/multiple-line/src/MultipleLineEditor.tsx
```diff
@@ -41,6 +55,17 @@ export function MultipleLineEditor(props: MultipleLineEditorProps) {
...
             />
+            {withCharValidation && (
+              <div className={styles.validationRow}>
+                <CharCounter value={value || ''} checkConstraint={checkConstraint} />
+                <CharValidation constraints={constraints} />
+              </div>
+            )}
+            {!withCharValidation && (
+              <div className={styles.validationRow}>
+                <CharCounter value={value || ''} checkConstraint={() => true} />
```
mshaaban0, line 66: These wrapper divs need some a11y love, `aria-live="polite"` this will tell screen readers to announce the content when it changes 🙂 even better if we do it inside the CharCount component itself

PR #796, opened 2021-08-05
feat: add suport for deserialization

cezarsmpio:
### Links
image

### Lists
image

### Tables
image

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
@@ -4,6 +4,7 @@ import {
   getRenderElement,
   getSlatePluginTypes,
   useStoreEditor,
+  getSlatePluginOptions,
 } from '@udecode/slate-plugins-core';
 import { Transforms, Element } from 'slate';
 import { INLINES } from '@contentful/rich-text-types';
@@ -61,7 +62,7 @@ function EmbeddedEntityInline(props: EmbeddedEntityInlineProps) {
   }
 
   return (
-    <span {...props.attributes} className={styles.root}>
+    <span {...props.attributes} className={styles.root} data-emebedded-entity-inline-id={entryId}>
       <span contentEditable={false}>
         <FetchingWrappedInlineEntryCard
           sdk={sdk}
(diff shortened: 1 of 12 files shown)


packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
```diff
@@ -61,7 +62,7 @@ function EmbeddedEntityInline(props: EmbeddedEntityInlineProps) {
   }
 
   return (
-    <span {...props.attributes} className={styles.root}>
+    <span {...props.attributes} className={styles.root} data-emebedded-entity-inline-id={entryId}>
```
cribbles, line 65: typo: should be `embedded` (also below)

packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -63,6 +64,79 @@ export function createHyperlinkPlugin(sdk: FieldExtensionSDK): SlatePlugin {
...
     onKeyDown: buildHyperlinkEventHandler(sdk),
+    deserialize: (editor) => {
+      const hyperlinkOptions = getSlatePluginOptions(editor, INLINES.HYPERLINK);
+      const entryHyperlinkOptions = getSlatePluginOptions(editor, INLINES.ENTRY_HYPERLINK);
+      const assetHyperlinkOptions = getSlatePluginOptions(editor, INLINES.ASSET_HYPERLINK);
+
+      const isAnchor = (element) =>
+        element.nodeName === 'A' &&
+        !!element.getAttribute('href') &&
+        element.getAttribute('href') !== '#';
```
cribbles, line 75: What's wrong with that? Is this something that comes up a lot in gdocs `href`s? (I get that the link wouldn't point anywhere -- even so...

I also wonder if we should think about js injection here via `javascript:`. But not something to capture in this PR and I don't think it's something we care about in the legacy editor anyhow.

cribbles (approved):
left a nit issue and a question, nothing blocking though, in general LGTM 👍

packages/rich-text/src/plugins/Hyperlink/index.tsx
cezarsmpio: Nothing is wrong to be honest... maybe we should validate if the `href` attribute is a valid URL?

PR #798, opened 2021-08-06
feat: disable disabled features

cribbles:
Wires toolbar buttons to the sdk-provided `enabledNodeTypes`, creating parity with the legacy editor behavior (also adding support for `table`). Doesn't disable keyboard shortcuts though.

### DIFF
--- a/packages/rich-text/src/Toolbar/index.tsx
+++ b/packages/rich-text/src/Toolbar/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { Fragment, useState } from 'react';
 
 import { css } from 'emotion';
 import { EditorToolbar, EditorToolbarDivider } from '@contentful/forma-36-react-components';
@@ -15,9 +15,12 @@ import { ToolbarTableButton } from '../plugins/Table';
 import { EmbeddedEntityDropdownButton } from '../plugins/EmbeddedEntity';
 import { ToolbarIcon as EmbeddedEntityBlockToolbarIcon } from '../plugins/EmbeddedEntityBlock';
 import { SPEditor, useStoreEditor } from '@udecode/slate-plugins-core';
-import { BLOCKS } from '@contentful/rich-text-types';
+import { BLOCKS, INLINES, MARKS } from '@contentful/rich-text-types';
 import { isNodeTypeSelected } from '../helpers/editor';
+import { isNodeTypeEnabled, isMarkEnabled } from '../helpers/validations';
 import { ToolbarEmbeddedEntityInlineButton } from '../plugins/EmbeddedEntityInline';
+import { useSdkContext } from '../SdkProvider';
+import { FieldExtensionSDK } from '@contentful/field-editor-reference/dist/types';
 
 type ToolbarProps = {
   isDisabled?: boolean;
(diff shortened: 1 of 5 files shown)


packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -42,61 +45,142 @@ const styles = {
...
+const EmbedAssetsWidget = ({ isDisabled }: ToolbarProps) => {
+  const sdk = useSdkContext();
   const [isEmbedDropdownOpen, setEmbedDropdownOpen] = useState(false);
   const onCloseEntityDropdown = () => setEmbedDropdownOpen(false);
   const onToggleEntityDropdown = () => setEmbedDropdownOpen(!isEmbedDropdownOpen);
 
+  const [canAccessAssets, setCanAccessAssets] = useState(false);
+  React.useEffect(() => {
+    sdk.access.can('read', 'Asset').then(setCanAccessAssets);
+  }, [sdk]);
```
cezarsmpio, line 57: Would it be better to be a `useMemo`?

packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -42,61 +45,142 @@ const styles = {
...
+        isDisabled={isDisabled}
+        onClose={onCloseEntityDropdown}
+        onToggle={onToggleEntityDropdown}
+        isOpen={isEmbedDropdownOpen}>
+        {icons}
+      </EmbeddedEntityDropdownButton>
+    );
+  } else {
+    return icons;
+  }
```
cezarsmpio, line 108: No need for the `else`.

packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -42,61 +45,142 @@ const styles = {
...
-        <EditorToolbarDivider />
-
-        <ToolbarBoldButton isDisabled={isDisabled} />
-        <ToolbarItalicButton isDisabled={isDisabled} />
-        <ToolbarUnderlineButton isDisabled={isDisabled} />
-        <ToolbarCodeButton isDisabled={isDisabled} />
+        {validationInfo.isAnyMarkEnabled && <EditorToolbarDivider />}
 
-        <EditorToolbarDivider />
+        {isMarkEnabled(sdk.field, MARKS.BOLD) && <ToolbarBoldButton isDisabled={isDisabled} />}
```
cezarsmpio, line 124: What's the difference between being `enabled` and `disabled` here? Can it be enabled and disabled at the same time? 😆 

I'm asking that because the `isDisabled` prop seems not necessary.

cezarsmpio (approved):
Great PR! Left some suggestions and a question only :)

packages/rich-text/src/Toolbar/index.tsx
cribbles: The `isDisabled` prop represents the button as disabled when the editor as a whole is in a disabled state, for example in the UI's read-only version comparison (diff) mode, or when an entry- or field-level error prevents the editor from being accessible. In cases like these, we still want buttons for valid marks and nodes to be represented, but displayed as disabled. By contrast, when a mark or node is disabled by the content model, we don't want to display the buttons to begin with.

cribbles:
@cezarsmpio reassigning you for review because:
- I responded to your suggestions in https://github.com/contentful/field-editors/pull/798/commits/9781b007e04898a473191490a6ce04be39f12a43
- I applied this logic to the hyperlink modal in https://github.com/contentful/field-editors/pull/798/commits/080c370e05c4a3bec4a6e2f1247083e313ce5533
- I applied this logic to the heading dropdown in https://github.com/contentful/field-editors/pull/798/commits/b191c0b5808775001dad36440a1f5752f541abe6

The last two points are not something we've discussed before but I think they're important for this PR as we will need to do them at some point in the future anyway.

cezarsmpio:
I can't approve it, I'm getting a 500 error.

Anyway, LGTM! ✅

PR #802, opened 2021-08-11
feat: add table tracking / remove keyboard shortcuts

cribbles:
Adds table tracking for insert and removal events, including `tablePaste`.

Does not include cypress tests, though I've confirmed this all works locally using the new `window.events` property in the demo environment.

Removes keyboard shortcuts (I'd add this in a separate PR but it would be too cumbersome to accomplish separately, given the code path overlap).


packages/rich-text/src/plugins/Table/index.tsx
```diff
@@ -85,24 +86,55 @@ export const withTableOptions: CustomSlatePluginOptions = {
...
+function hasTables(nodes: CustomElement[]) {
+  const toCheck = ([] as CustomElement[]).concat(nodes);
+  while (toCheck.length) {
+    const { type, children } = toCheck.shift() as CustomElement;
+    if (type === BLOCKS.TABLE) {
+      return true;
+    } else if (children) {
+      // Ideally we'd simply spread here, but we'd need to
+      // upgrade TS first.
+      Array.prototype.push.apply(toCheck, children);
```
z0al, line 119: What error do we get here?

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -41,21 +41,22 @@ import {
   withEmbeddedEntityInlineOptions,
 } from './plugins/EmbeddedEntityInline';
 import { SdkProvider } from './SdkProvider';
+import { TrackingProvider, useTrackingContext } from './TrackingProvider';
 import { sanitizeIncomingSlateDoc, sanitizeSlateDoc } from './helpers/sanitizeSlateDoc';
 import { TextOrCustomElement } from './types';
 
 type ConnectedProps = {
   sdk: FieldExtensionSDK;
+  onAction: (name: string, data: Record<string, unknown>) => unknown;
```
z0al, line 50: Nit: I think having more specific typing for `name` would be ideal

z0al (approved):
LGTM. Left small comments/questions

packages/rich-text/src/plugins/Table/index.tsx
cribbles:
```
This syntax requires an imported helper named '__spreadArray' with 3
parameters, which is not compatible with the one in 'tslib'. Consider
upgrading your version of 'tslib'.ts
```

PR #803, opened 2021-08-13
fix: FieldConnector setValue race conditions

andipaetzold:
The `FieldConnector` currently only works correctly when the `onValueChanged` is triggered immediately after calling `setValue`. If another callback triggers a rerender (e.g. `onSchemaErrorsChanged`) the child component is rerendered with an outdated `value`. This can be potentially caused by `setValue` triggering a longer async function (e.g. a network connection). This gives plenty of time for another callback to fire. Also, throttling the `setValue` may also cause race conditions here.

This PR resolves this problem by immediately storing the value of `setValue` in the `FieldConnector`'s state. That way all future rerenders triggered by any callback will always pass the correct value to the component's `children`. The `setValue` call of the `FieldAPI` stays throttled to not break any existing behaviour.

This PR includes a test that demonstrates the issue and fails on `master`.

### DIFF
--- a/packages/_shared/src/FieldConnector.ts
+++ b/packages/_shared/src/FieldConnector.ts
@@ -1,6 +1,6 @@
 import React from 'react';
-import throttle from 'lodash/throttle';
 import isEqual from 'lodash/isEqual';
+import throttle from 'lodash/throttle';
 import { FieldAPI } from '@contentful/app-sdk';
 
 type Nullable = null | undefined;
@@ -18,7 +18,6 @@ export interface FieldConnectorChildProps<ValueType> {
 interface FieldConnectorState<ValueType> {
   isLocalValueChange: boolean;
   externalReset: number;
-  lastSetValue: ValueType | Nullable;
   lastRemoteValue: ValueType | Nullable;
   value: ValueType | Nullable;
   disabled: boolean;
@@ -60,7 +59,6 @@ export class FieldConnector<ValueType> extends React.Component<
       isLocalValueChange: false,
       externalReset: 0,
       value: initialValue,
-      lastSetValue: initialValue,
       lastRemoteValue: initialValue,
       disabled: props.isInitiallyDisabled,
       errors: [],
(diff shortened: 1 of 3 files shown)


DanweDE:
> The `FieldConnector` currently only works correctly when the `onValueChanged` is triggered immediately after calling `setValue`.

Wasn't this exactly how `onValueChanged` used to work? It would trigger even before the network call to persist the value, so you should never have this problem:

> This can be potentially caused by setValue triggering a longer async function (e.g. a network connection).

I am a bit worried that if we changed the `sdk` behavior and callback orders change, customers' custom field editors might also start behaving buggy for no good reason. If `onValueChanged` really used to fire immediately after `setValue` then perhaps it'd be best to continue with this behavior. It'd be best if the callback order is clearly defined and predictable exactly to avoid such regressions and race conditions whenever possible.

DanweDE (commented):
> The `FieldConnector` currently only works correctly when the `onValueChanged` is triggered immediately after calling `setValue`.

Wasn't this exactly how `onValueChanged` used to work? It would trigger even before the network call to persist the value, so you should never have this problem:

> This can be potentially caused by setValue triggering a longer async function (e.g. a network connection).

I am a bit worried that if we changed the `sdk` behavior and callback orders change, customers' custom field editors might also start behaving buggy for no good reason. If `onValueChanged` really used to fire immediately after `setValue` then perhaps it'd be best to continue with this behavior. Ideally, callback order would be clearly defined and predictable exactly to avoid such regressions and race conditions whenever possible.

packages/_shared/src/FieldConnector.ts
```diff
@@ -71,34 +69,30 @@ export class FieldConnector<ValueType> extends React.Component<
   unsubscribeDisabled: Function | null = null;
   unsubscribeValue: Function | null = null;
 
-  setValue = throttle(
+  setValue = async (value: ValueType | Nullable) => {
+    if (this.props.isEmptyValue(value === undefined ? null : value)) {
```
gui-santos, line 73:
```suggestion
    if (this.props.isEmptyValue(value ?? null)) {
```

What do you think about doing this ☝️ ?

packages/_shared/src/FieldConnector.ts
```diff
@@ -71,34 +69,30 @@ export class FieldConnector<ValueType> extends React.Component<
...
-              lastSetValue: value,
-            },
-            () => {
-              this.props.field.setValue(value).then(resolve).catch(reject);
-            }
-          );
-        });
-      }
+      return new Promise((resolve, reject) => {
+        if (this.props.isEmptyValue(value === undefined ? null : value)) {
```
gui-santos, line 85: Same suggestion here :)

gui-santos (approved):
LGTM! 

Although I don't have a lot of context about the problem you are fixing here, maybe someone else can have more valuable comments on that regard ✌️

packages/_shared/src/FieldConnector.ts
andipaetzold: Normally I am the person mentioning these fixes in reviews 🤔

DanweDE (approved):
Thank you for the elaborate summary @andipaetzold :bow: 

one question:
> `onSchemaErrorsChanged` (value A) is triggered

Since `setValue` (value B) was already triggered, would it make sense for the `sdk` to hold this callback back?

andipaetzold:
> Thank you for the elaborate summary @andipaetzold 🙇
> 
> one question:
> 
> > `onSchemaErrorsChanged` (value A) is triggered
> 
> Since `setValue` (value B) was already triggered, would it make sense for the `sdk` to hold this callback back? Not sure what the exact consequences for the UI would be, perhaps it doesn't make any sense.

@DanweDE 
If we held the callback back we might not receive intermediate errors that could be still valid (e.g. min/max length of a textfield). But it would also make sense to not emit these callbacks as they are based on outdated values.

Reality is that the callbacks are pretty certainly already queued up and are only delivered very slow. Therefore they would need to be filtered out by the App SDK. Unfortunately, there is no reference between the errors and the value-change which triggered them, so this is almost impossible.

PR #805, opened 2021-08-16
feat: Forma v4 components adoption

burakukula:
Adopts Forma v4 components in all field-editors and app examples.


packages/reference/package.json
```diff
@@ -24,6 +24,7 @@
     "@contentful/field-editor-shared": "^0.22.0",
     "@contentful/forma-36-react-components": "^3.93.4",
     "@contentful/forma-36-tokens": "^0.11.0",
+    "@contentful/f36-badge": "4.0.1-next-v4-6179.2211",
```
suevalov, line 27: I think we can just use `"@contentful/f36-badge": "next-v4"` to always install latest published one

packages/rich-text/src/plugins/Hyperlink/EntityHyperlinkTootip.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { hyperlinkTooltipStyles as styles } from './styles';
-import { Tag } from '@contentful/forma-36-react-components';
+import { EntityStatusBadge } from "@contentful/f36-badge";
```
gui-santos, line 4: This is kind of off-topic, but is there a reason for `EntityStatusBadge` not being documented in the website and storybook?

gui-santos (commented):
LGTM!

I guess we will wait all the other PR's to be merged here before approving this one, right?

suevalov:
> LGTM!
> 
> I guess we will wait all the other PR's to be merged here before approving this one, right?

Ideally, we would wait till Forma V4 is beta. At least.

packages/rich-text/src/plugins/Hyperlink/EntityHyperlinkTootip.js
burakukula: I don't really know why EntityStatusBadge is not documented. I will write this down as something that we need to impove.

z0al (commented):
LGTM 👍🏻 . When do we plan to merge this?

burakukula:
> LGTM 👍🏻 . When do we plan to merge this?
@z0al we need to migrate all components first to v4, so it will stay here for a while, open. it's hard for us to ETA right now.

suevalov (approved):

bgutsol (approved):

PR #807, opened 2021-08-18
refactor: upgrade to @udecode/plate-*

z0al:
Mostly find and replace but came across typing issues that I needed to skip for now.


packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
```diff
@@ -66,12 +67,14 @@ export const createEmbeddedAssetBlockPlugin = createEmbeddedEntityPlugin(BLOCKS.
 export const withEmbeddedEntryBlockOptions: CustomSlatePluginOptions = {
   [BLOCKS.EMBEDDED_ENTRY]: {
     type: BLOCKS.EMBEDDED_ENTRY,
+    // @ts-expect-error
```
cribbles, line 70: what's the issue with these references now?

cribbles (approved):
pretty much LGTM (I know this is still in draft state), but I wasn't clear on the new `// ts-expect-error`s, maybe throw an explanatory comment in somewhere?

cribbles:
@z0al we'll likely have problems embedding this in other applications that are not using react 16 :/

packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
z0al: The typing of components expects the props to be either `RenderElementProps` or `RenderLeafProps` however, for some components we expect only elements. I wasn't sure what to do in these cases do so decided to push the changes and see what others think.

z0al:
Yeah, true @cribbles . However, I checked all of our `plate-*` and `slate-react` packages and they all require React v16.8.0 or higher. So we should be Ok. Still can't figure why we get the error though.

PR #818, opened 2021-08-25
fix(rich-text): correct inline embeds insertion

DanweDE:
Ensures when inserting inline embeds, after calling focus stealing entity picker, the embed is inserted at the original text selection.

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
@@ -16,6 +16,7 @@ import { CustomSlatePluginOptions } from '../../types';
 import newEntitySelectorConfigFromRichTextField from '../../helpers/newEntitySelectorConfigFromRichTextField';
 import { useSdkContext } from '../../SdkProvider';
 import { FetchingWrappedInlineEntryCard } from './FetchingWrappedInlineEntryCard';
+import { createInlineEntryNode } from './Util'
 
 const styles = {
   icon: css({
@@ -23,7 +24,7 @@ const styles = {
   }),
 
   root: css({
-    margin: '0px 5px',
+    margin: '0 1px',
     fontSize: 'inherit',
     span: {
       webkitUserSelect: 'none',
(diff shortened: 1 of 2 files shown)


packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
```diff
@@ -89,24 +90,19 @@ async function selectEntityAndInsert(editor, sdk: FieldExtensionSDK) {
...
-          linkType: 'Entry',
-        },
-      },
-    },
-  };
+  const inlineEntryNode = createInlineEntryNode(entry.sys.id)
 
-  Transforms.insertNodes(editor, inlineEntryNode);
+  // Got to wait until focus is really back on the editor or setSelection() won't work.
+  setTimeout(() => {
```
z0al, line 102: Is it worth using something like [p-wait-for](https://www.npmjs.com/package/p-wait-for) to provide more stability? In other words, is `0` is always enough?

```javascript
waitFor(ReactEditor.isFocused).then(()=> {
    Transforms.setSelection(editor, selection)
    Transforms.insertNodes(editor, inlineEntryNode);
})
```


P.S. if you choose to use `p-wait-for` make sure to stick to `v3` as `v4` uses ESM modules.

z0al (approved):
LGTM. left a single suggestion.

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
DanweDE: I believe the problem is that `ReactEditor.isFocused` will be `true` immediately, but I'll double-check.

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
```diff
@@ -89,24 +90,19 @@ async function selectEntityAndInsert(editor, sdk: FieldExtensionSDK) {
     ...newEntitySelectorConfigFromRichTextField(sdk.field, INLINES.EMBEDDED_ENTRY),
     withCreate: true,
   };
+  const selection = editor.selection;
```
cezarsmpio, line 93: I wouldn't recommend doing this, I'd rather create a new object to keep the values where they are at the moment of the selection, `editor.selection` can change, therefore `selection` will also be different.

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
```diff
@@ -166,21 +162,10 @@ export function createEmbeddedEntityInlinePlugin(sdk): SlatePlugin {
...
-                      type: 'Link',
-                      linkType: 'Entry',
-                    },
-                  },
-                },
-              };
+              const entryId = element.getAttribute('data-embedded-entity-inline-id');
+              return entryId
+                ? createInlineEntryNode(entryId)
+                : undefined;
```
cezarsmpio, line 168: Returning `undefined` can break other deserializations in my experience with Slate. Please, test pasting of other elements as well. I think SlateJS sees `undefined` as a value.

cezarsmpio (approved):

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
DanweDE: `editor.selection` object reference can change, but the selection object props themselves are immutable (read only props).

If you'd want to work with the assumption that the object could be mutated for some reason, then you should use `cloneDeep(editor.selection)` as there's some nesting.

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
DanweDE: Yes, it's `true` immediately after calling `focus()` :roll_eyes: Also see my [Slack #Slate message](https://slate-js.slack.com/archives/C1RH7AXSS/p1629833275346700?thread_ts=1629790927.323000&cid=C1RH7AXSS).

I'll probably open an issue with them. If `ReactEditor.focus()` would return a promise and `ReactEditor.isFocused()` turns `true` only once the `Promise` resolves, that'd be great.

but there's also the issue of why do we need to call `focus()` in the first place for ` Transforms.setSelection(editor, selection)` to work.

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
DanweDE: The previous code
```
if (!embeddedEntityInlineId) return;
```
would also result in `undefined`, just that we're explicit about it now :thinking:

PR #824, opened 2021-08-27
fix: blockquote wrapping/unwrapping buggy

cezarsmpio:
It unwraps blockquote when trying to add one of those types inside a blockquote:

- Heading
- List (UL, OL)
- HR

### DIFF
--- a/cypress/integration/RichTextEditor.spec.ts
+++ b/cypress/integration/RichTextEditor.spec.ts
@@ -273,6 +273,28 @@ describe('Rich Text Editor', () => {
 
         expectRichTextFieldValue(expectedValue);
       });
+
+      it('should unwrap blockquote', () => {
+        editor().click().typeInSlate('some text');
+
+        getQuoteToolbarButton().click();
+
+        const expectedQuoteValue = doc(
+          block(BLOCKS.QUOTE, {}, block(BLOCKS.PARAGRAPH, {}, text('some text', []))),
+          block(BLOCKS.PARAGRAPH, {}, text('', []))
+        );
+        expectRichTextFieldValue(expectedQuoteValue);
+
+        getHrToolbarButton().click();
+
+        const expectedHeadingValue = doc(
+          block(BLOCKS.PARAGRAPH, {}, text('some text', [])),
+          block(BLOCKS.HR, {}),
+          block(BLOCKS.PARAGRAPH, {}, text('', [])),
+          block(BLOCKS.PARAGRAPH, {}, text('', []))
+        );
+        expectRichTextFieldValue(expectedHeadingValue);
+      });
     });
   });
 
--- a/cypress/integration/RichTextEditor.spec.ts
+++ b/cypress/integration/RichTextEditor.spec.ts
@@ -288,7 +288,7 @@ describe('Rich Text Editor', () => {
     ];
 
     headings.forEach(([type, label, shortcut]) => {
-      describe.only(label, () => {
+      describe(label, () => {
         it(`allows typing ${label} (${type})`, () => {
           editor().click().typeInSlate('some text');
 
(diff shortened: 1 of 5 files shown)


cypress/integration/RichTextEditor.spec.ts
```diff
@@ -269,6 +273,28 @@ describe('Rich Text Editor', () => {
...
+        );
+        expectRichTextFieldValue(expectedQuoteValue);
+
+        getHrToolbarButton().click();
+
+        const expectedHeadingValue = doc(
+          block(BLOCKS.PARAGRAPH, {}, text('some text', [])),
+          block(BLOCKS.HR, {}),
+          block(BLOCKS.PARAGRAPH, {}, text('', [])),
+          block(BLOCKS.PARAGRAPH, {}, text('', []))
```
DanweDE, line 294: Who do we have two empty paragraphs now? Seems odd but I guess out of scope here.

cypress/integration/RichTextEditor.spec.ts
```diff
@@ -269,6 +273,28 @@ describe('Rich Text Editor', () => {
...
+
+        const expectedQuoteValue = doc(
+          block(BLOCKS.QUOTE, {}, block(BLOCKS.PARAGRAPH, {}, text('some text', []))),
+          block(BLOCKS.PARAGRAPH, {}, text('', []))
+        );
+        expectRichTextFieldValue(expectedQuoteValue);
+
+        getHrToolbarButton().click();
+
+        const expectedHeadingValue = doc(
```
DanweDE, line 290: Heading value? Copy paste?

cypress/integration/RichTextEditor.spec.ts
```diff
@@ -315,6 +341,46 @@ describe('Rich Text Editor', () => {
 
           getDropdownToolbarButton().should('have.text', label);
         });
+
+        if (type !== BLOCKS.PARAGRAPH) {
```
DanweDE, line 345: Considering Cypress tests are slow (I didn't check how slow though in this case) would it make sense to just run this for one heading type or ideally try all of them in one test? A `TODO` would be sufficient if you agree, I don't think we've got time for this right now.

cypress/integration/RichTextEditor.spec.ts
```diff
@@ -470,6 +532,30 @@ describe('Rich Text Editor', () => {
...
+        it('should unwrap blockquote', () => {
+          editor().click().typeInSlate('some text');
+
+          getQuoteToolbarButton().click();
+
+          const expectedQuoteValue = doc(
+            block(BLOCKS.QUOTE, {}, block(BLOCKS.PARAGRAPH, {}, text('some text', []))),
+            block(BLOCKS.PARAGRAPH, {}, text('', []))
+          );
+          expectRichTextFieldValue(expectedQuoteValue);
```
DanweDE, line 545: Do we really need to assert all of this in each of these tests or could we at least put it into a helper `insertTextAsBlockQuote('some text')` ?

DanweDE (approved):
Approving code with some suggestions.

I didn't have time to test though.

cypress/integration/RichTextEditor.spec.ts
cezarsmpio: 1 empty paragraph when adding blockquote and another one when adding HR. We could be more surgical and manage that but that's also how the current editor looks like when doing the same action.

cypress/integration/RichTextEditor.spec.ts
cezarsmpio: 🙈

cypress/integration/RichTextEditor.spec.ts
cezarsmpio: We have 3 elements that cannot interact with blockquote: headings, lists and HR. I think it makes sense to have test for each element interaction. I'm ok with the helper idea, don't mind.

cypress/integration/RichTextEditor.spec.ts
DanweDE: Yes, we should definitely improve where it makes sense and where it's no effort. The old editor definitely wasn't perfect.

At least it seems this is just a behavior when a blockquote gets unwrapped, we're not spamming the editor with unnecessary paragraphs when just inserting a hr so I don't think it's a common thing to happen :sweat_smile:

DanweDE (approved):
:+1:

cypress/integration/RichTextEditor.spec.ts
cezarsmpio: Not sure what happened to my last comment but I said that the old editor behavior is not necessarily perfect but we are at least not inserting "new things", things will stay as is but again, it doesn't mean it's the right behavior. 

Inserting an empty paragraph for blockquote and HR below them is "necessary" otherwise the customer will not be able to move through rows. The problem is that we are not removing those empty paragraphs when unwrapping things, which is where we can do better.

PR #826, opened 2021-08-31
refactor: use createPlateOptions

DanweDE:
Calls `createPlateOptions()` on our own plugin options as mentioned on Plate's configuration docs: https://plate.udecode.io/docs/configuration/#factory


DanweDE:
This is probably a bad idea after all and it doesn't seem to help with any of the current issues anyhow (although it solves TS issues for `Paragraph`)

1. `createPlateOptions()` creates plugin options for all plugins they support, not just the ones we're using.
2. If there are default options that we do not overwrite, e.g. for hotkeys, then I imagine we'll end up with undesired hotkeys etc.

PR #828, opened 2021-09-02
feat: add contentful editor provider and remove editor key

cezarsmpio:
~The `key` property breaks the editor regardless of its value.~

EDIT

Sorry, many changes are not related to the real problem but I took advantage and fixed some type issues. I was getting some `module not found` issues.

EDIT 2

The `key` property is necessary for collaborative editing, so I returned it and it seems the issue was the missing id when calling `useStoreEditorRef` to get the editor instance. 

I tested it by linking my library locally to the webapp and confirmed that the error we are trying to solve is gone and collaborative editing is working fine by opening the same entry on two different tabs.


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -179,28 +178,28 @@ const RichTextEditor = (props: Props) => {
...
+            <FieldConnector
+              throttle={0}
+              field={sdk.field}
+              isInitiallyDisabled={isInitiallyDisabled}
+              isEmptyValue={isEmptyValue}
+              isEqualValues={deepEquals}>
+              {({ lastRemoteValue, disabled, setValue, externalReset }) => (
+                <ConnectedRichTextEditor
+                  {...otherProps}
+                  key={`rich-text-editor-${externalReset}`}
```
cezarsmpio, line 192: We still need `key` for the collaborative editing.

DanweDE (changes requested):
This bug still seems an issue with this change:

issue-back

It only happens in the 2nd instance, not in the 1st instance.

cezarsmpio:
@DanweDE did a few more tests and it seems to be working fine, even on different locales with multiple RTE fields:

image

image

DanweDE (approved):

PR #829, opened 2021-09-03
chore: bump rich text field package

cezarsmpio:


apps/rich-text-app/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@contentful/app-sdk": "^3.36.0",
-    "@contentful/field-editor-rich-text": "^0.21.7-next",
+    "@contentful/field-editor-rich-text": "^0.22.0-next",
```
DanweDE, line 7: doesn't really matter, but why not 0.21.8-next?

DanweDE (approved):

apps/rich-text-app/package.json
cezarsmpio: Ah, just following "conventional commits" here, the last commits had `feat:`, so this means a new minor.

z0al (approved):

PR #831, opened 2021-09-06
chore: create insert before first void block global plugin

cezarsmpio:
It works for:

- HR
- Embedded asset
- Embedded entry

Any other void block?

### DIFF
--- a/cypress/integration/RichTextEditor.spec.ts
+++ b/cypress/integration/RichTextEditor.spec.ts
@@ -302,6 +302,25 @@ describe('Rich Text Editor', () => {
         );
         expectRichTextFieldValue(expectedValue);
       });
+
+      // TODO: Cypress throwing after {enter}: Cannot resolve a Slate node from DOM node: [object HTMLSpanElement]
+      // Question: Can we fix it by upgrading the Cypress version? Is this one of those issues that happen because Cypress doesn't fire all the events Slate needs in its current version?
+      it.skip('should add line if HR is the first void block', () => {
+        editor().click();
+
+        getHrToolbarButton().click();
+
+        cy.findByTestId('rich-text-editor-hr').click();
+
+        editor().type('{enter}');
+
+        const expectedValue = doc(
+          block(BLOCKS.PARAGRAPH, {}, text('', [])),
+          block(BLOCKS.HR, {}),
+          block(BLOCKS.PARAGRAPH, {}, text('', []))
+        );
+        expectRichTextFieldValue(expectedValue);
+      });
     });
   });
 
--- a/packages/rich-text/src/plugins/InsertBeforeFirstVoidBlock/index.tsx
+++ b/packages/rich-text/src/plugins/InsertBeforeFirstVoidBlock/index.tsx
@@ -1,0 +1,40 @@
+import { KeyboardEvent } from 'react';
+import { PlatePlugin, SPEditor } from '@udecode/plate-core';
+import {
+  isFirstChild,
+  isVoid,
+  getElementFromCurrentSelection,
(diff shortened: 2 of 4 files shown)


packages/rich-text/src/plugins/InsertBeforeFirstVoidBlock/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+            type: BLOCKS.PARAGRAPH,
+            data: {},
+            children: [{ text: '' }],
+          };
+
+          // It adds two paragraphs outside a timeout and after moving the cursor to the previous line, not sure why...
+          setTimeout(() => {
+            Transforms.insertNodes(editor, paragraph, { at: currentPath as Path });
+            moveToThePreviousLine(editor);
+          }, 0);
```
cezarsmpio, line 31: Without the timeout:

- Editor -> HR + Paragraph -> Enter **becomes** Editor -> Paragraph + Paragraph (selection) + HR + Paragraph

With the timeout:

- Editor -> HR + Paragraph -> Enter **becomes** Editor -> Paragraph (selection) + HR + Paragraph

cezarsmpio: 🤷

cypress/integration/RichTextEditor.spec.ts
```diff
@@ -302,6 +302,25 @@ describe('Rich Text Editor', () => {
         );
         expectRichTextFieldValue(expectedValue);
       });
+
+      // TODO: Cypress throwing after {enter}: Cannot resolve a Slate node from DOM node: [object HTMLSpanElement]
+      // Question: Can we fix it by upgrading the Cypress version? Is this one of those issues that happen because Cypress doesn't fire all the events Slate needs in its current version?
```
z0al, line 307: I'd really like us to upgrade to the latest Cypress sooner than later. We are waaay behind. But that's not something trivial.

z0al (commented):
Looks good 👍🏻  but let's try to ask about the timeout issue in Slack? we don't want this to be another source of bugs later 😕

z0al:
@cezarsmpio You may want to take a look at the [trailing-block](https://github.com/udecode/plate/blob/main/packages/trailing-block/src/createTrailingBlockPlugin.ts) plugin as it's very similar to what you're doing but inserts at the very end instead of at the beginning?

cezarsmpio:
@z0al they fix different things. This PR is for empty paragraphs above of first void blocks only.

cezarsmpio:
@z0al I also asked the Plate team about the timeout, you can follow the thread here if you wish: https://slate-js.slack.com/archives/C013QHXSCG1/p1631022429081900

packages/rich-text/src/plugins/InsertBeforeFirstVoidBlock/index.tsx
z0al: FYI @cezarsmpio I tried the suggestion from Slack (i.e. `Editor.withoutNormalizing`), it doesn't fix this issue.

cezarsmpio: Ok, thanks!

cypress/integration/RichTextEditor.spec.ts
cezarsmpio: We've tried to update it but all other tests from other packages started failing.

z0al (approved):
Approving in case you want to merge it as it's and follow up. But ideally, we want to figure out the `setTimeout` issue or add some working tests before merging this

PR #833, opened 2021-09-07
fix(rich-text): force trailing paragraph

z0al:
Used [@udecode/plate-trailing-block](https://github.com/udecode/plate/tree/main/packages/trailing-block) to keep an empty paragraph at the end of the document all the time. I think this behaviour matches the old editor IIRC.

Let me know if we have other custom logic somewhere that can be removed.

See [Forced the layout](https://plate.udecode.io/docs/plugins/forced-layout) docs. 

Let's wait for CI 😄


cribbles (approved):

cezarsmpio (approved):

z0al:
Sadly, it completely breaks the list plugin at Cypress and I see some weird behavior Tables too 😞

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -91,6 +92,9 @@ const getPlugins = (sdk: FieldExtensionSDK, tracking: TrackingProvider) => {
     createCodePlugin(),
     createItalicPlugin(),
     createUnderlinePlugin(),
+
+    // Other
+    createTrailingParagraphPlugin(),
```
cezarsmpio, line 97: The order matters here, shouldn't we add it to the top instead? So its rules will run before our plugins, but at the same time I'm afraid our plugins can override some of its rules 🤷‍♀️

z0al: 🤷🏻‍♂️

z0al:
Fixed all issues except the hyperlink tests. some of the tests fail on Cypress even though seem to work just fine in the actual Cypress UI.

z0al:
Wait it passed on Circle!!

what

cezarsmpio (approved):

PR #838, opened 2021-09-10
feat(rich-text): Table header support

z0al:
- [x] Render `ELEMENT_TH` as `<th>` with correct styling
- [x] Enable Table header by default 
- [x] Add options to enable/disable table header
- [x] Header-aware `addColumn` helpers
- [x] Never add headers in `addRow` helpers and prevent adding row above header cell
- [x] Update Cypress tests


packages/rich-text/src/plugins/Table/TableActions.tsx
```diff
@@ -36,26 +38,64 @@ export const TableActions = () => {
...
     Slate.ReactEditor.focus(editor);
-  };
+  }, [editor]);
+
+  React.useEffect(() => {
+    setHeaderEnabled(editor && isTableHeaderEnabled(editor));
+  }, [editor]);
+
+  const canInsertRowAbove = React.useMemo(() => {
+    if (!editor) {
```
cezarsmpio, line 58: We could use `editor.selection` if we want to know if the editor is focused. Sometimes we have `editor` but there's nothing selected.

cezarsmpio (approved):
Well done 👏

packages/rich-text/src/plugins/Table/TableActions.tsx
z0al: The `getAbove` call should use the `editor.selection` since we don't pass any path. 

But I guess in this particular case it doesn't really matter. We only use the `canInsertRowAbove` to enable/disable items on the actions menu which is only visible if the user focuses on a table cell

PR #839, opened 2021-09-10
refactor: use exit break plugin instead

cezarsmpio:
Using one of their solutions to replace `setTimeout`. We decided to use their own plugin. The whole conversation can be read here: https://slate-js.slack.com/archives/C013QHXSCG1/p1631208644108100?thread_ts=1631022429.081900&cid=C013QHXSCG1

The second solution would be just adding a `select: true` property to `Transforms.insertNodes` such as:

```
Transforms.insertNodes(editor, paragraph, { at: currentPath as Path, select: true }
```

### DIFF
--- a/packages/rich-text/package.json
+++ b/packages/rich-text/package.json
@@ -27,10 +27,11 @@
     "prepublishOnly": "yarn build"
   },
   "dependencies": {
-    "@contentful/forma-36-react-components": "^3.97.0",
     "@contentful/contentful-slatejs-adapter": "^14.2.0",
     "@contentful/field-editor-reference": "^2.20.6",
+    "@contentful/forma-36-react-components": "^3.97.0",
     "@contentful/rich-text-types": "^15.3.0",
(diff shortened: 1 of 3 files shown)


packages/rich-text/package.json
```diff
@@ -27,10 +27,11 @@
     "prepublishOnly": "yarn build"
   },
   "dependencies": {
-    "@contentful/forma-36-react-components": "^3.97.0",
     "@contentful/contentful-slatejs-adapter": "^14.2.0",
     "@contentful/field-editor-reference": "^2.20.6",
+    "@contentful/forma-36-react-components": "^3.97.0",
     "@contentful/rich-text-types": "^15.3.0",
+    "@udecode/plate-break": "^3.0.3",
```
cezarsmpio, line 34: https://plate.udecode.io/docs/plugins/exit-break/

z0al (approved):

z0al:
Note: the commit type should be `refactor` instead of `chore` :)

PR #843, opened 2021-09-15
chore: fix nested lists by bumping plate

djagya:


z0al:
The upgrade also fixes the table paste issue I'm working on

z0al:
The new undo/redo behavior is weird. when you write "some text" and press CTRL+Z it undos to "some tex". You have to press multiple times to clear the whole sentence. I'm not sure if it's Ok (we just adjust the test) or we should try to "fix" it.

cribbles (approved):

PR #845, opened 2021-09-16
chore: export reference utils

cezarsmpio:
I will need those utils to fix the embedded entry card.

### DIFF
--- a/packages/reference/src/index.tsx
+++ b/packages/reference/src/index.tsx
@@ -21,3 +21,5 @@ export type {
   MissingEntityCardProps,
   RenderCustomMissingEntityCard,
 } from './common/customCardTypes';
+export { isValidImage } from './utils/isValidImage';
+export { fromFieldValidations } from './utils/fromFieldValidations';
--- a/packages/reference/package.json
+++ b/packages/reference/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/field-editor-reference",
-  "version": "2.20.6",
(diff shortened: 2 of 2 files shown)


djagya (approved):

packages/reference/src/index.tsx
```diff
@@ -21,3 +21,5 @@ export type {
   MissingEntityCardProps,
   RenderCustomMissingEntityCard,
 } from './common/customCardTypes';
+export { isValidImage } from './utils/isValidImage';
+export { fromFieldValidations } from './utils/fromFieldValidations';
```
z0al, line 25: A similar function is also exported from `field-editor-shared` did you try it?

cezarsmpio: Let me check! :)

cezarsmpio: Hum, actually maybe we should move `isValidImage` to `field-editor-shared`?

z0al: It's only used in `isValidImage` but also doesn't seem tied to the reference logic. I'm fine with that.

The two implementations of `fromFieldValidations` seem to be slightly different. Not sure which one is more correct/complete.

cezarsmpio: I think I don't need `fromFieldValidations`, right now I'm only using `isValidImage`.

z0al: Let's not export it then

PR #846, opened 2021-09-16
feat: export shared util to validate images

cezarsmpio:
Next step: remove this util from the `reference` package and bump a minor version there, so current packages using it won't break.


packages/_shared/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/field-editor-shared",
-  "version": "0.22.1",
+  "version": "0.23.0",
```
z0al, line 3: The package should be released automatically. We don't need this version bump

z0al (approved):

PR #853, opened 2021-09-17
refactor: implement entry card for rich text

cezarsmpio:
It fixes and re-implement entry and asset cards. It mimics some of the behavior from the `referece` package but I removed all the unnecessary parts for the `rich-text`.


z0al (approved):
LGTM

PR #856, opened 2021-09-17
refactor: refactor HelpText to Text from Forma36 v4 [BAU-104]

bgutsol:
## Open question

I'm not sure if it's 100% correct decision to refactor the `HelpText` to simple `Text` in the places where it's actually used in pair with form fields. But from other side using `FormControl.HelpText` without `FormControl` itself will bring even more confusion. 🤷‍♂️ 
For example `FieldWrapper` component seems like a bit customised `FormControl` component from our forma36 components.


burakukula:
@bgutsol I agree with you, I think it might be useful to have a `HelpText` component, since we are using the same setup of styles as `FormControl.HelpText` but outside of the form context. Let's discuss how to solve this issue and maybe hold with the adoption for now? WDYT?

gui-santos (approved):

suevalov (approved):

PR #857, opened 2021-09-17
refactor: 💡 tooltip-v4

thebesson:
Run codemod and update forma-36 Tooltip component to v4
Some code had to be updated, as new tooltip API doesn't allow anything other than `string` as content


packages/rich-text/src/plugins/Hyperlink/EntityHyperlinkTootip.js
```diff
@@ -1,125 +0,0 @@
-import React from 'react';
```
z0al, line 1: Do we really want to remove this tooltip?

tooltip

thebesson: by the guidelines, the content of a tooltip should be string, we don't remove it, but update it, not the content inside will be just text.

suevalov (approved):

djagya (approved):

PR #861, opened 2021-09-21
chore: Icon v4 adoption

suevalov:
Adopts Icon v4 component


bgutsol (approved):
Great 👍
I've noticed that you also fixed some Button adoptions. 
Did we already do `Button` adoption? If so I guess it is worth to do it one more time since we could have miss it in some other files/projects as well.

suevalov:
@bgutsol Looks like, Button codemod wasn't run for Javascript files only. We'll do a final clean up before merging migration PR and make sure that all components are coming from V4.

PR #871, opened 2021-09-24
feat: 🎸 adopt IconButton component from forma V4

massao:


packages/rating/src/RatingRibbon.tsx
```diff
@@ -53,12 +56,17 @@ export class RatingRibbon extends React.Component<RatingRibbonProps, RatingRibbo
       <>
         {items.map((num) => (
           <IconButton
+            variant="transparent"
```
bgutsol, line 59: Shouldn't we remain the same logic for button as well? 
`{this.isSelected(num) ? 'primary' : 'transparent'}`

massao: No, otherwise the button would be blue and the icon would also be blue, to keep the same styling of the V3, we need the button as transparent, and the icon with the correct variant.
If both of them are rendered as primary, we would end up with this:
https://v4-f36-storybook.netlify.app/?path=/story/components-button-iconbutton--basic&args=variant:primary

bgutsol: Got you, thanks for clarifying!

bgutsol (approved):

thebesson (approved):

PR #872, opened 2021-09-28
fix: render invalid table children as text

cribbles:
Accomplishes this in the `normalizeNode` step definition for tables.

If an incoming child node is invalid, we:
- convert it to a contentful node
- gets the plain text value of the contentful node, as we would we with a document (using `@contentful/rich-text-plain-text-renderer`
- renders the plain text value inside a paragraph in place of the invalid node

This works well with insertion in empty table cells as well as within text selections from table cells.


z0al (approved):

z0al:
@cribbles I tested this on the preview URL ☝🏻 and it seems to paste `undefined` all the time. Can you check again?

cribbles:
@z0al discussed this with you separately, but posting here for transparency: This is due to an idiosyncrasy in our `rich-text` libraries that will be dealt with separately (basically our adapters need to know about the table types, and they don't right now, plus `'undefined'` should never be a valid output in most cases).

PR #874, opened 2021-09-29
feat: 🎸 Run Select v4 codemod

massao:


packages/date/src/TimezonePickerInput.tsx
```diff
@@ -18,15 +19,15 @@ export const TimezonepickerInput = ({
       aria-label="Select timezone"
       testId="timezone-input"
       value={value}
-      width="medium"
+      size="medium"
```
suevalov, line 22: As far as I know, `size` is a new property and we completely removed `width` property

massao: Oh ok, I thought it was replaced by the size, I'll remove it and update the codemod to drop the width

bgutsol (approved):
Looks good 👍

suevalov (approved):

PR #877, opened 2021-09-30
feat: enable block elements in list items

djagya:
Done by passing the list of all blocks to `validLiChildrenTypes` list plugin and additionally by patching it, the plugin doesn't change the element if it's in that list.


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -81,7 +82,9 @@ const getPlugins = (sdk: FieldExtensionSDK, tracking: TrackingProvider) => {
 
     // Block Elements
     createParagraphPlugin(),
-    createListPlugin(),
+    createListPlugin({
+      validLiChildrenTypes: Object.values(BLOCKS),
```
z0al, line 86: This should be [LIST_ITEM_BLOCKS](https://github.com/contentful/rich-text/blob/4e82b615d47b7730f5c791d0e274eba49ef27813/packages/rich-text-types/src/schemaConstraints.ts#L58)

Also a nit: to follow our convention let's export a custom `createListPlugin` from `/plugins/list/index.ts` and pass params there

cezarsmpio (approved):

PR #879, opened 2021-10-01
feat: v4 Form component adoption

burakukula:


packages/markdown/src/dialogs/EmdebExternalContentDialog.tsx
```diff
@@ -74,112 +67,110 @@ export const EmbedExternalContentModal = ({ onClose }: EmbedExternalContentModal
     }
   }, [mainInputRef]);
 
-  return (
```
gui-santos, line 77: It seems that you are only removing the `()` here
Why?

packages/markdown/src/dialogs/InsertLinkModalDialog.tsx
```diff
@@ -29,77 +29,75 @@ export const InsertLinkModal = ({ selectedText, onClose }: InsertLinkModalProps)
     }
   }, [mainInputRef]);
 
-  return (
```
gui-santos, line 32: same here

packages/markdown/src/dialogs/InsertTableModalDialog.tsx
```diff
@@ -25,70 +25,68 @@ export const InsertTableModal = ({ onClose }: InsertTableModalProps) => {
     }
   }, [mainInputRef]);
 
-  return (
```
gui-santos, line 28: ... and here

gui-santos (changes requested):
Looks good, but I didn't understand why the `()` in some components was removed

packages/markdown/src/dialogs/EmdebExternalContentDialog.tsx
burakukula: hmm, you are right. I have run the codemod, didn't remove it manually. Something is wrong with the codemod? hmmmm, gonna check it. thanks!

massao: try running the `prettier`, the default of the jscodeshift is not wrap single components in `()` in this case the single element being the `Fragment`.
But we have configured in prettier that it needs to, so prettier should fix it

massao (approved):

gui-santos (approved):

PR #881, opened 2021-10-05
feat: [EXT-3077] bump app sdk to v4

Jwhiles:


shikaan (approved):

z0al (approved):

packages/markdown/src/dialogs/openMarkdownDialog.tsx
```diff
@@ -19,64 +19,63 @@ import {
 } from './EmdebExternalContentDialog';
 import { ZenModeModalDialog, ZenModeResult } from './ZenModeModalDialog';
 
-export const openMarkdownDialog = (
```
Jwhiles, line 22: I made the mistake of running prettier

packages/validation-errors/src/ValidationErrors.tsx
```diff
@@ -59,9 +55,13 @@ function UniquenessError(props: UniquenessErrorProps) {
...
   );
 
+  let conflicting: any[] = []
+  if ('conflicting' in props.error) {
+    conflicting = props.error.conflicting
+  }
   React.useEffect(() => {
     const entryIds = state.entries.map((entry) => entry.id);
-    const conflictIds = props.error.conflicting?.map((entry) => entry.sys.id) || [];
+    const conflictIds = conflicting.map((entry) => entry.sys.id)
```
Jwhiles, line 64: I'm not sure if this is right - as it doesn't match the type we have in the SDK here https://github.com/contentful/ui-extensions-sdk/blob/0ec6a68cdf04992db31381f13254fe4e2d48d817/lib/types/validation-error.ts#L33

z0al (approved):

packages/validation-errors/src/ValidationErrors.tsx
z0al: The SDK types are probably wrong. I will provide you a link in Slack (public repo :) )

packages/validation-errors/src/ValidationErrors.tsx
```diff
@@ -59,9 +55,13 @@ function UniquenessError(props: UniquenessErrorProps) {
     [props.localeCode, props.defaultLocaleCode, contentTypesById]
   );
 
+  let conflicting: any[] = []
+  if ('conflicting' in props.error) {
+    conflicting = props.error.conflicting
+  }
```
z0al, line 61: Any reason for moving away from the one-liner `props.error.conflicting?.map((entry) => entry.sys.id) || [];`?

packages/reference/package.json
```diff
@@ -36,12 +36,12 @@
...
   },
   "devDependencies": {
-    "@contentful/app-sdk": "^3.36.0",
+    "@contentful/app-sdk": "^4.1.0",
     "@contentful/field-editor-test-utils": "^0.17.3",
     "@testing-library/react-hooks": "^5.0.3"
   },
   "peerDependencies": {
-    "@contentful/app-sdk": "^3.36.0",
+    "@contentful/app-sdk": "^4.1.",
```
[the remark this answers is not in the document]
Jwhiles: yep :)

packages/validation-errors/src/ValidationErrors.tsx
z0al: TIL

packages/reference/src/common/useAccessApi.ts
```diff
@@ -1,10 +1,11 @@
...
   AccessAPI,
-  ArchiveableAction,
-  CrudAction,
-  PublishableAction,
 } from '@contentful/app-sdk';
 
+
+type CrudAction = 'create' | 'read' | 'update' | 'delete'
+type PublishableAction = 'publish' | 'unpublish'
+type ArchiveableAction = 'archive' | 'unarchive'
```
andipaetzold, line 8: Did we accidentally remove that from App SDK?

Jwhiles: They're no longer exported - I think we probably should export them once more. WDYT?

andipaetzold: yes I think we should reexport that

PR #908, opened 2021-11-04
fix: pass a unique form-control id

suevalov:
Use sdk.field information to pass a unique field id


packages/default-field-editors/src/FieldWrapper.tsx
```diff
@@ -45,8 +45,11 @@ export const FieldWrapper: React.FC<FieldWrapperProps> = function (props: FieldW
     });
   }, [field]);
 
+  const fieldControlId = [field.id, field.locale].filter((item) => item).join('-');
```
bgutsol, line 48:
```suggestion
  const fieldControlId = [field.id, field.locale].filter(Boolean).join('-');
```
💅

bgutsol (approved):

PR #911, opened 2021-11-08
fix: embedded block between paragraphs []

cezarsmpio:
It adds a paragraph between blocks if siblings. 

The current behavior selects the next embedded block if they are siblings.

### DIFF
--- a/cypress/integration/rich-text/RichTextEditor.spec.ts
+++ b/cypress/integration/rich-text/RichTextEditor.spec.ts
@@ -9,6 +9,7 @@ import { expectRichTextFieldValue } from './utils';
 
 // the sticky toolbar gets in the way of some of the tests, therefore
 // we increase the viewport height to fit the whole page on the screen
+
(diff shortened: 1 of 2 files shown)


cypress/integration/RichTextEditor.spec.ts
```diff
@@ -1139,6 +1206,73 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
 
     for (const [triggerMethod, triggerEmbeddedAsset] of methods) {
       describe(triggerMethod, () => {
+        it('adds paragraph before the block when pressing enter if the block is first child', () => {
```
DanweDE, line 1209: "is first child" or "is "first document node" ? Should be the latter.

There's a slight difference, a block can actually be the first child e.g. within a list item containing multiple blocks so let's be more accurate.

DanweDE (approved):
LGTM overall, but some nit comments :+1:

DanweDE (approved):

PR #912, opened 2021-11-08
chore: bump rte next to 2.0.0-next.0

DanweDE:
Since in the `master` branch we're on `1-0-0` now, we'll be going for version `2-0-0` with the editor rewrite.

Subsequent `next` versions should only use `npm version prerelease`, e.g. from `2.0.0-next.0` to `2.0.0-next.1`


cezarsmpio (approved):

packages/rich-text/src/plugins/Heading/index.tsx
```diff
@@ -202,7 +202,11 @@ export function ToolbarHeadingButton(props: ToolbarHeadingButtonProps) {
       }
 
       toggleNodeType(editor, { activeType: type, inactiveType: type });
-      Slate.ReactEditor.focus(editor);
+
+      // TODO: Figure out why focus only works with timeout here.
+      setTimeout(() => {
+        Slate.ReactEditor.focus(editor);
```
cezarsmpio, line 208: I think we can use `Transform.select` before. I asked the Plate team once about it, this might be helpful for you: https://slate-js.slack.com/archives/C013QHXSCG1/p1631022429081900

PR #915, opened 2021-11-09
fix: deserialize different lines as multiple paragraphs

IBratWurst:
[SHE-561](https://contentful.atlassian.net/browse/SHE-561)

### DIFF
--- a/packages/rich-text/src/plugins/Paragraph/index.tsx
+++ b/packages/rich-text/src/plugins/Paragraph/index.tsx
@@ -8,6 +8,7 @@ import { PlatePlugin, getRenderElement, SPEditor } from '@udecode/plate-core';
 import { getToggleElementOnKeyDown } from '@udecode/plate-common';
 import { CustomSlatePluginOptions } from '../../types';
 import { deserializeElement } from '../../helpers/deserializer';
+import { Element, Node, Transforms } from 'slate';
 
 const styles = {
   [BLOCKS.PARAGRAPH]: css`
@@ -31,8 +32,8 @@ function isEmbed(element: HTMLElement) {
   );
 }
 
-function isParagraphDiv(element: HTMLElement) {
-  return element.nodeName === 'DIV' && !isEmbed(element);
+function isEmpty(element: HTMLElement) {
+  return element.textContent === '';
 }
 
 export function createParagraphPlugin(): PlatePlugin {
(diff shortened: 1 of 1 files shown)


packages/rich-text/src/plugins/Paragraph/index.tsx
```diff
@@ -33,7 +33,7 @@ export function createParagraphPlugin(): PlatePlugin {
     onKeyDown: getToggleElementOnKeyDown(BLOCKS.PARAGRAPH),
     deserialize: deserializeElement(BLOCKS.PARAGRAPH, [
       {
-        nodeNames: ['P'],
+        nodeNames: ['P', 'DIV'],
```
cezarsmpio, line 36: We need to be careful when adding `DIV` here, sometimes div is a container with no text at all, and this will result in a lot of empty paragraphs for each div.

Look at my first try of adding `div` as a valid node: https://github.com/contentful/field-editors/blame/4efe00650c34261f90bbd73e5fa430f34e3ab438/packages/rich-text/src/plugins/Paragraph/index.tsx#L36-L56

Did you try copying and pasting embedded blocks with this approach?

cezarsmpio: It doesn't work when copying embedded blocks because they contain div, and according to this logic, divs should be considered as paragraphs, which is not true in some cases: https://recordit.co/MW6qrjCRn6

cezarsmpio: When copying some texts that contain a lot of containers as div, each div will result in an empty paragraph as I commented previously:

image

Imagine this scenario:

`div > div > div > section > div > p + img + p`

Each div here will be considered as a paragraph. I'm not saying this is wrong, it might be acceptable but it's worth discussing this with Malin maybe?

IBratWurst: @cezarsmpio thanks for the explanation. I overlooked that as I was focused at getting a div into a paragraph.

I came up with a new version looking at yours, just not as "plain". So, your previous attempt was missing the nesting control, as mine now, right? Do we need the check for emptiness?

Kapture 2021-11-10 at 08 36 07

cezarsmpio: It looks much better! :) 

> Do we need the check for emptiness?

I think we should... Editors won't need to keep deleting those meaningless empty paragraphs. What do you think?

IBratWurst: A case with some nested html:
Kapture 2021-11-10 at 09 13 09

packages/rich-text/src/plugins/Paragraph/index.tsx
```diff
@@ -44,9 +55,29 @@ export function createParagraphPlugin(): PlatePlugin {
...
+                break;
+              }
+              if (isParagraphDiv(currentParent)) {
+                hasParagraphAnchestor = true;
+              }
+              currentParent = currentParent.parentElement;
+            }
+            if (!hasEmbedAnchestor && hasParagraphAnchestor) {
+              return;
+            }
```
cezarsmpio, line 79: Can't we solve this by using `editor.normalizeNodes`? I think we can use the rule of `['P', 'DIV']` and adding something like this to ensure only inline elements inside a paragraph:

```js
editor.normalizeNode = entry => {
    const [node, path] = entry
    // If the element is a paragraph, ensure its children are valid.
    if (Element.isElement(node) && node.type === BLOCKS.PARAGRAPH) {
      for (const [child, childPath] of Node.children(editor, path)) {
        if (Element.isElement(child) && !editor.isInline(child)) {
          Transforms.unwrapNodes(editor, { at: childPath })
          return
        }
      }
    }
    // Fall back to the original `normalizeNode` to enforce other constraints.
    normalizeNode(entry)
  }
```

packages/rich-text/src/plugins/Paragraph/index.tsx
```diff
@@ -44,9 +55,29 @@ export function createParagraphPlugin(): PlatePlugin {
         element: element?.map((deserializeNode) => ({
           ...deserializeNode,
           deserialize: (el: HTMLElement) => {
-            if (el.nodeName === 'DIV' && el.hasAttribute('data-entity-type')) {
+            if (el.textContent === '' || isEmbed(el)) {
```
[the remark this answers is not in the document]
IBratWurst: Done!

packages/rich-text/src/plugins/Paragraph/index.tsx
IBratWurst: Sorry @cezarsmpio, I could get back at this only today. Yes normalization does the trick in a nicer way, so I used it.
I tried to play with the available options and method but non covered all the scenarios. Anyway the simple unwrap solves the bug while keeping the previous behaviors intact or improved.

cezarsmpio (approved):

PR #916, opened 2021-11-09
feat: make embedded blocks draggable []

cezarsmpio:
Any HTML element become interactive  when adding `draggable="true"`. I found that by adding this attribute the whole block becomes draggable and Slate kind of take care of the rest.

I can drag:

- Embedded blocks
- Embedded inlines

Where:

- After, before and in the middle of paragraphs
- After and before other embedded blocks
- Inside blockquote elements
- Inside tables (only embedded inline elements)

### DIFF
--- a/packages/rich-text/src/plugins/DragAndDrop/index.tsx
+++ b/packages/rich-text/src/plugins/DragAndDrop/index.tsx
@@ -1,0 +1,67 @@
+import { Node as SlateNode, Transforms } from 'slate';
+import { PlatePlugin } from '@udecode/plate-core';
+import { CustomElement } from '../../types';
+import { BLOCKS, CONTAINERS } from '@contentful/rich-text-types';
+
+export function createDragAndDropPlugin(): PlatePlugin {
+  // Elements that don't allow other elements to be dragged into them and which callback should be used
+  const DND_BLOCKED_ELEMENTS = {
+    [BLOCKS.TABLE]: Transforms.removeNodes,
+    [BLOCKS.QUOTE]: Transforms.liftNodes,
+  };
+
+  // DND_BLOCKED_ELEMENTS callbacks will run on those elements only
+  const DRAGGABLE_TYPES: string[] = [BLOCKS.EMBEDDED_ENTRY, BLOCKS.EMBEDDED_ASSET];
+
+  // HTML node names where dropping should be disabled, usually when using `Transforms.removeNodes` callback
+  const ON_DROP_BLOCKED_TYPES = ['TABLE'];
+
+  return {
+    withOverrides: (editor) => {
+      const { normalizeNode } = editor;
+
+      editor.normalizeNode = (entry) => {
+        const [node, path] = entry;
+
+        Object.keys(DND_BLOCKED_ELEMENTS).forEach((blockedElementType) => {
+          const nodeType = (node as CustomElement).type;
+
+          if (SlateNode.isNode(node) && nodeType === blockedElementType) {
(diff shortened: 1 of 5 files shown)


z0al (approved):

DanweDE:
One very nice side-effect of this seems to be that we're also getting the inline embed highlighting back on cursor interactions:

packages/rich-text/src/plugins/DragAndDrop/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+      const [draggingBlock] = Array.from(
+        getNodes(editor, {
+          match: (node) => Editor.isBlock(editor, node) && DRAGGABLE_TYPES.includes(node.type),
+        })
+      );
+      if (!draggingBlock) return false;
+
+      const [draggingNode] = draggingBlock;
+
+      // TODO: looking up for html nodes is not the best solution and it won't scale, we need to find a way to know the dropping target slate element
```
DanweDE, line 76: I don't fully understand this comment as I don't see html here. So you mean "slate nodes" ?

packages/rich-text/src/plugins/DragAndDrop/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        });
+
+        normalizeNode(entry);
+      };
+
+      return editor;
+    },
+
+    // If true, the next handlers will be skipped.
+    onDrop: (editor) => (event) => {
```
DanweDE, line 66:
```suggestion
    onDrop: (editor) => (event: any) => {
```

and get rid of `// @ts-expect-error` below. Or, ideally, can we use the right type here which I assume should be some sort of Slate drag event, or where does `.path` come from?

DanweDE:
Now we're getting into nice UX (nicer than what we need right now) but do you have any idea how difficult it'd be to show a different cursor while hovering over the table when dragging the embed? That way we could just indicate that the user can't drop it there :sweat_smile:

DanweDE (approved):
Great job :+1: 

Did you look into Cypress by any chance and would it be difficult to add tests for drag & drop there? Could be a nice follow-up for testing week.

packages/rich-text/src/plugins/DragAndDrop/index.tsx
cezarsmpio: We are using `TABLE` from `ON_DROP_ALLOWED_TYPES` where `TABLE` is a HTML node name.

packages/rich-text/src/plugins/DragAndDrop/index.tsx
cezarsmpio: I couldn't find an ideal type for this event. It would be `DragEvent` but it doesn't contain the property `path`.

djagya (approved):
👏

DanweDE:
There were some failing tests
failed-tests

Fixed by

```
          addEmbeddedEntry();
          selectAndPressEnter(); // Inserts paragraph before embed because it's in the first line.
          addEmbeddedEntry();
          selectAndPressEnter(); // inserts paragraph in-between embeds.
```

instead of 

```
          addEmbeddedEntry();
          addEmbeddedEntry();
          selectAndPressEnter();
          selectAndPressEnter();
```

The effect should be the same, but it seems that the editor click (bottom) in `addEmbeddedEntry` does not fully move the editor focus away from the embed onto the paragraph below in time before the second embed is inserted, thus the second embed is replacing the first embed.

No idea TBH why this didn't previously fail though. Perhaps @cezarsmpio has an idea, but at least we should be able to merge this now.

PR #919, opened 2021-11-11
fix: pasting text content inside lists

z0al:
### Summary

It pastes anything inside lists as a text for now. A follow-up will be done next week to properly handle the pasting of other nodes.

Upgraded Plate dependencies along the way.


packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -139,7 +150,52 @@ export const withListOptions: CustomSlatePluginOptions = {
...
     validLiChildrenTypes: LIST_ITEM_BLOCKS,
-  });
+  };
+
+  const plugin = createPlateListPlugin(options);
+
+  plugin.withOverrides = withCustomList(options);
+
+  return plugin;
+};
```
DanweDE, line 201: Couldn't we instead use what we had and make minimal adjustments by some minimal overwrites?

```ts
export const createListPlugin = () =>
  createPlateListPlugin({
    validLiChildrenTypes: LIST_ITEM_BLOCKS,
    withOverrides: (options: WithListOptions) => (editor) => {
      withList(options)(editor); // This is the default for the `withOverrides` option.

      const origInsertFragments = editor.insertFragment

      editor.insertFragments = (editor: PlateEditor) => (fragment: TDescendant[]) => {
        const newFragment = fragmentToPlaintextFragment(fragment)
        origInsertFragments(newFragment)
      }
    }
  })
});
```

This should make forking `getListInsertFragment.ts` irrelevant or am I missing anything?

djagya (approved):
Looks good apart from the Daniel's question which would maybe make unnecessary copying the whole plugin?

packages/rich-text/src/plugins/List/index.tsx
z0al: That assumes we can pass a custom `withOverrides` but it looks like the list plugin [doesn't care](https://github.com/udecode/plate/blob/1406095fd388e53fcc3d7940f4b9a70125df5e36/packages/elements/list/src/createListPlugin.ts#L16) about what we pass there. cc @djagya

z0al: Ok so it's a little bit more complicated. Returning only fragments from `fragmentToPlaintextFragment` won't solve the issue as we still need to insert these correctly (the upstream implementation doesn't handle text correctly either).

I pushed a commit attempting to reduce the amount of copied code for `withCustomList` but I don't think we can't do much about the `getListInsertFragment` for now :/

z0al:
> Looks good apart from the Daniel's question which would maybe make unnecessary copying the whole plugin?

We didn't really copy the whole plugin (sorry for my bad English in the code comment), it's just the `getListInsertFragment` implementation as I couldn't figure out another way to override its logic.

packages/rich-text/src/plugins/List/index.tsx
DanweDE: Alright, let's just get it out then :shrug: 
I was also thinking, in the worst case let's just reduce code the way you already did.

PR #920, opened 2021-11-11
ci: integration tests via matrix

anho:
This runs the integration tests on more browsers.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,25 @@
 version: 2.1
 
+executors:
+  linux-node:
+    docker:
+      - image: cimg/node:14.17
+    environment:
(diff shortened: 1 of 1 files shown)


cezarsmpio (approved):

.circleci/config.yml
```diff
@@ -13,58 +33,59 @@ commands:
...
       - run: yarn test:ci
+
   integration-tests:
-    resource_class: medium+
+    parameters:
+      executor:
+        type: executor
+      browser:
+        type: enum
+        enum: [chrome, firefox, edge]
```
cezarsmpio, line 66: It seems that `edge` for linux is still in Beta. Is this a problem? Can we rely on it?

anho: I will just keep it optional for now.

PR #922, opened 2021-11-11
fix: no cell deletion via keyboard

cribbles:
This solves an issue in which deleting via the backspace or delete keys within or before a table would delete the current or forthcoming table cell.

### DIFF
--- a/packages/rich-text/src/plugins/Table/index.tsx
+++ b/packages/rich-text/src/plugins/Table/index.tsx
@@ -24,6 +24,10 @@ import { CustomElement, CustomSlatePluginOptions } from '../../types';
 import { isTableActive, insertTableAndFocusFirstCell } from './helpers';
 import { TrackingProvider, useTrackingContext } from '../../TrackingProvider';
 import { Element, Node, Text, Transforms } from 'slate';
+import {
+  currentSelectionPrecedesTableCell,
+  currentSelectionStartsTableCell,
+} from '../../helpers/editor';
 
 const styles = {
   [BLOCKS.TABLE]: css`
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/plugins/Table/index.tsx
```diff
@@ -199,11 +207,66 @@ function hasHeadersOutsideFirstRow(nodes: CustomElement[]) {
...
+    BLOCKS.TABLE_HEADER_CELL,
+  ]);
+  return (tableCellNode && !getText(editor, path)) || editor.selection?.focus.offset === 0;
+}
+
+/**
+ * This traversal strategy is unfortunately necessary because Slate doesn't
+ * expose something like Node.next(editor).
+ */
+function getNextNode(editor: SPEditor): CustomElement | null {
```
DanweDE, line 225: Can we move this and perhaps `isAtEndOfTextSelection` to a utils/helper file? They seem pretty generic and a bit distracting from the tables logic in here.

packages/rich-text/src/plugins/Table/index.tsx
```diff
@@ -199,11 +207,66 @@ function hasHeadersOutsideFirstRow(nodes: CustomElement[]) {
...
+  const [tableCellNode, path] = getNodeEntryFromSelection(editor, [
+    BLOCKS.TABLE_CELL,
+    BLOCKS.TABLE_HEADER_CELL,
+  ]);
+  return (tableCellNode && !getText(editor, path)) || editor.selection?.focus.offset === 0;
+}
+
+/**
+ * This traversal strategy is unfortunately necessary because Slate doesn't
+ * expose something like Node.next(editor).
```
DanweDE, line 223: :scream: there's no helpers for that provided anywhere?

DanweDE (commented):
Looking pretty good :+1: Waiting for the link to the preview to test it out.

cezarsmpio (approved):

packages/rich-text/src/plugins/Table/index.tsx
cribbles: `Path.next` "does not ensure that the returned `Path` is valid in the document" [[1](https://docs.slatejs.org/api/locations/path#path.next-path-path-greater-than-path)]. In practice this means it's programmatically +1s the last number in the path given and is really only useful for placing new nodes ahead of the current one within a path.

DanweDE:
@cribbles I found a regression:



I also noticed that you can't even delete an empty paragraph before the table anymore.

Backspace generally doesn't work anymore if the cursor is placed right before the table, so you can't e.g. delete a whole selected blob of text or an empty paragraph before the table. Example:

DanweDE:
Also, I still found a way how to delete cells.
My intention was to delete the cell's text, but the last cell is deleted again...

cribbles:
> @cribbles I found a regression:
> 
> 
> 
> I also noticed that you can't even delete an empty paragraph before the table anymore.
> 
> Backspace generally doesn't work anymore if the cursor is placed right before the table, so you can't e.g. delete a whole selected blob of text or an empty paragraph before the table. Example:
> 
> 

Fixed in https://github.com/contentful/field-editors/pull/922/commits/8e4ec79c099a5af38d6843946377465c02022a66.

cribbles:
> Also, I still found a way how to delete cells. My intention was to delete the cell's text, but the last cell is deleted again... 
> 
> If you hit BACKSPACE it does nothing, ideally it would still delete all the text in this case (without deleting any cells of course)
> 
> 🤔 I am wondering... couldn't we add a normalization rule that ensures that there are always enough (equal number of) cells in each row? With that, we would shield against any such bugs in the future.

As we discussed, I think this should be handled separately by normalizing tables to always maintain a uniform cell count.

PR #924, opened 2021-11-12
fix: {wip} table markup paste

cribbles:


z0al:
@cribbles wanna kill this one? 😉

PR #928, opened 2021-11-17
fix: internal copy and paste for firefox, chromium and safari []

cezarsmpio:
It seems Slate has compatibility issues with some browsers. Specially the ones that don't support the `beforeInput` event and this seems to affect the `contentEditable` attribute.

Their codebase has a lot of calls to browser detection, they even have a file with some [browser detection helpers](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/utils/environment.ts).

I copied some from there and I tested on the following browsers on Mac OS:

- Latest Chrome
- Latest Opera
- Latest Safari
- Latest Edge
- Latest Firefox

They all worked fine as expected, even the command to cut & copy.

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedEntityBlock/LinkedEntityBlock.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedEntityBlock/LinkedEntityBlock.tsx
@@ -7,6 +7,7 @@ import { CustomRenderElementProps } from '../../types';
 import { ReactEditor, useSelected, useReadOnly } from 'slate-react';
 import { Transforms } from 'slate';
 import { useContentfulEditor } from '../../ContentfulEditorProvider';
+import { HAS_BEFORE_INPUT_SUPPORT } from '../../helpers/environment';
 
 const styles = {
   root: css({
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/plugins/EmbeddedEntityBlock/LinkedEntityBlock.tsx
```diff
@@ -49,9 +50,13 @@ export function LinkedEntityBlock(props: LinkedEntityBlockProps) {
       className={styles.root}
       data-entity-type={entityType}
       data-entity-id={entityId}
-      draggable={true}
-      contentEditable={false}>
-      <div>
+      // COMPAT: This makes copy & paste work for Firefox
+      contentEditable={!HAS_BEFORE_INPUT_SUPPORT ? false : undefined}
+      draggable={!HAS_BEFORE_INPUT_SUPPORT ? true : undefined}>
```
djagya, line 55: So does it mean that the block won't be draggable if `beforeInput` is supported?

djagya (approved):

packages/rich-text/src/plugins/EmbeddedEntityBlock/LinkedEntityBlock.tsx
cezarsmpio: Dragging is supported whether `beforeInput` is supported or not. This "hack" is for copy & paste solution only. Interestingly the `draggable` attribute plays a role here as well. I've tested both drag & drop and copy & paste on the browsers I've mentioned above.

PR #930, opened 2021-11-17
fix: fix list field preventing commas after forma36 v4

djagya:
The list field stopped accepting typed commas, that happened after the Forma36 update to version 4, where `TextInput` lost its internal state. When converting from array of values and back to the string, we lose the trailing commas.

This PR adds the `ListEditorInternal` component which implements the internal state allowing us to preserve the trailing commas.

### DIFF
--- a/packages/list/src/ListEditor.tsx
+++ b/packages/list/src/ListEditor.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { FieldAPI, FieldConnector, LocalesAPI } from '@contentful/field-editor-shared';
 import * as styles from './styles';
+import isEqual from 'lodash/isEqual';
 
 import { TextInput } from '@contentful/f36-components';
 import { FieldConnectorChildProps } from '@contentful/field-editor-shared/dist/FieldConnector';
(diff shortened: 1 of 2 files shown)


packages/list/src/ListEditor.tsx
```diff
@@ -38,30 +40,45 @@ export function ListEditor(props: ListEditorProps) {
...
-        };
+function ListEditorInternal({
+  setValue,
+  value,
+  errors,
+  disabled,
+  direction,
+  isRequired,
+}: FieldConnectorChildProps<ListValue> & { direction: 'rtl' | 'ltr'; isRequired: boolean }) {
+  const [valueState, setValueState] = React.useState(() => (value || []).join(', '));
```
bgutsol, line 58: minor 💅
```suggestion
  const [valueState, setValueState] = React.useState((value || []).join(', '));
```

bgutsol (approved):
👍

packages/list/src/ListEditor.tsx
djagya: Theoretically it makes sense to compute the initial state just once, but ofc in this specific case it won't make a big difference in performance

bgutsol: I'm not sure if I get your point right.
But just to clarify, there is no difference if you pass initial state as a function or as a value It will be always computed only once on the initial render, isn’t it?

But sure, it's a minor thing and it will not affect performance in any way

djagya: Not really, if you specify the initial value not wrapped in a function, then it will be computed on every component render. So if we would be doing some expensive calculations to get the initial state, it must be wrapped in the function, then `setState` calls it just once.

bgutsol: Ah, I got it now. Thanks @djagya for explaining and sorry for ranting 🙇‍♂️

PR #933, opened 2021-11-19
fix: h5 & h6 are incorrectly marked as h1

z0al:


cezarsmpio (approved):

cezarsmpio:
Wondering why our tests didn't catch that, maybe the HTML representation was incorrect but the JSON structure output was correct and being sent to the API correctly.

PR #934, opened 2021-11-19
feat: replace empty paragraphs with tables when inserted []

cezarsmpio:


z0al (approved):
LGTM

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -658,6 +658,47 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+            table(row(emptyHeader(), emptyHeader()), row(emptyCell(), emptyCell())),
+            emptyParagraph()
+          )
+        );
+      });
+
+      describe('Toolbar', () => {
+        const buttonsToDisableTable = ['quote', 'ul', 'ol'];
+
+        buttonsToDisableTable.forEach((button) => {
```
DanweDE, line 690: I think we can be more pragmatic and test all of these buttons in one test to have a positive impact on their runtime.

```ts
it('disables toolbar buttons for elements unsupported within tables')
```

There's also `hr` which is not allowed within a table and you can also assert that the embeds button turns into "Embed inline entry" only.

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -658,6 +658,47 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+          it(`should disable table button on toolbar if ${button} is selected`, () => {
```
DanweDE, line 691: I think you meant
```suggestion
          it(`should disable ${button} button on toolbar if a table is focused`, () => {
```

cypress/integration/rich-text/RichTextEditor.spec.ts
cezarsmpio: We have tests for the second case you've mentioned: https://github.com/contentful/field-editors/blob/next/cypress/integration/rich-text/RichTextEditor.spec.ts#L622

I really don't have a strong opinion against having that in one test or not.

cypress/integration/rich-text/RichTextEditor.spec.ts
cezarsmpio: No, what I meant is right. That's what I wanted :) I don't want to add a table inside a blockquote or a list, so the table button should be disabled in those cases.

DanweDE: Sorry, I actually got this test a bit wrong. With that in mind, my suggestion above also didn't really make sense. The `${button}` variable naming also threw me a bit off track here as I was thinking "button" while we're referring to the actual element being selected, resulting in the table button being disabled.

PR #937, opened 2021-11-23
feat: remove empty paragraphs when pressing delete []

cezarsmpio:


packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
```diff
@@ -108,18 +106,7 @@ export function getWithEmbeddedEntityEvents(
...
-
-      if (pathToSelectedElement) {
-        if (isEnter) {
-          event.preventDefault();
-          moveToTheNextLine(editor);
-        } else if (event.key === 'Backspace') {
-          event.preventDefault();
-          Transforms.removeNodes(editor, { at: pathToSelectedElement });
-        }
-      } else if (
```
cezarsmpio, line 122: We don't need this anymore since we are using the `InsertBeforeFirstVoidBlock` plugin.

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -359,6 +359,63 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+          doc(
+            block(BLOCKS.EMBEDDED_ENTRY, {
+              target: {
+                sys: {
+                  id: 'example-entity-id',
+                  type: 'Link',
+                  linkType: 'Entry',
+                },
+              },
+            }),
```
cezarsmpio, line 390: Should we add tests for other blocks such as lists, blockquotes, hr, etc or that's enought?

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -360,6 +360,67 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+                    type: 'Link',
+                    linkType: 'Entry',
+                  },
+                },
+              }),
+              block(BLOCKS.PARAGRAPH, {}, text(''))
+            )
+          );
+
+          richText.editor.click('bottom').type('{uparrow}').type('{uparrow}').type('{del}');
```
cezarsmpio, line 391: I've tried `.type('{uparrow}{uparrow}')` but I think it triggers it too fast that the browser doesn't catch it sometimes and result in flakiness.

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -360,6 +360,73 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+            )
+          );
+
+          richText.editor
+            .click('bottom')
+            .type('{uparrow}')
+            .type('{uparrow}')
+            .type('{uparrow}')
+            .type('{del}')
+            .type('{del}');
```
cezarsmpio, line 397: Need to do this instead of `.type('{uparrow}{uparrow}{uparrow}{del}{del}')` because I think Cypress triggers it too fast and the editor doesn't get it, resulting in flakiness.

DanweDE: Could you try it as one command with a `{ delay:` option instead to make it a bit more explicit? (a comment would also help)

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -360,6 +360,73 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+            doc(
+              block(type, {}, text('x')),
+              block(BLOCKS.EMBEDDED_ENTRY, {
+                target: {
+                  sys: {
+                    id: 'example-entity-id',
+                    type: 'Link',
+                    linkType: 'Entry',
+                  },
+                },
```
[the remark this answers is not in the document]
cezarsmpio: Will do!

DanweDE (approved):
LGTM :+1: 

Behavior isn't implemented for tables yet: Cursor placed in empty paragraph before a table -> hitting DELETE has no effect.

We can get this merged and address it in a follow-up PR.

PR #941, opened 2021-11-25
feat: list node normalization

cribbles:
Implements list node normalization via `normalizeNode`.


packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -157,12 +161,67 @@ const withCustomList = (options): WithOverride => {
...
 
-    // Use our custom getListInsertFragment
+    // // Use our custom getListInsertFragment
     editor.insertFragment = getListInsertFragment(editor);
 
+    const { normalizeNode } = editor;
+    editor.normalizeNode = (entry) => {
+      const [node, path] = entry;
+
+      // If the element is a list, ensure its children are list nodes.
```
DanweDE, line 174: What's our strategy to ideally account for our A/C's
> Try to encapsulate the code in a reusable way so that it can be used in other normalization callbacks
?

Can we e.g. use `rich-text-types` constants instead of hardcoding certain things here and extract it into a reusable function?

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -149,6 +152,60 @@ export const withListOptions: CustomSlatePluginOptions = {
...
+
+/**
+ * Ensures each list item follows the list schema.
+ */
+const normalizeList = (editor: Editor, path: Path) => {
+  for (const [child, childPath] of Node.children(editor, path)) {
+    if (Element.isElement(child)) {
+      unwrapInvalidGrandchildren(editor, childPath);
+      if (!isListItem(child)) {
+        Transforms.wrapNodes(editor, emptyNodeOfType(BLOCKS.LIST_ITEM), { at: childPath });
```
z0al, line 188: I believe we should exit the loop early here, otherwise, this code will likely crash when more than one list item is invalid. The same applies to other normalize* functions

Read about [Multi-pass normalization](https://docs.slatejs.org/concepts/11-normalizing#multi-pass-normalizing) here

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -149,6 +152,60 @@ export const withListOptions: CustomSlatePluginOptions = {
...
+
+const isValidInsideList = (node: CustomElement) =>
+  (LIST_ITEM_BLOCKS as Array<TopLevelBlockEnum | INLINES>)
+    .concat(Object.values(INLINES))
+    .includes(node.type as TopLevelBlockEnum);
+
+const unwrapInvalidGrandchildren = (editor: Editor, path: Path) => {
+  for (const [grandchild, grandchildPath] of Node.children(editor, path)) {
+    if (!isValidInsideList(grandchild as CustomElement)) {
+      Transforms.unwrapNodes(editor, { at: grandchildPath });
```
z0al, line 175: I don't understand the expected behavior in this case. I thought we are supposed to convert the invalid nodes to text?

Or are you waiting for the helper to be available?

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
```diff
@@ -54,8 +59,155 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+                )
+              )
+            )
+          ),
+          block(BLOCKS.PARAGRAPH, {}, text(''))
+        )
+      );
+    });
+
+    // TODO: Update when https://contentful.atlassian.net/browse/SHE-605 is completed
```
z0al, line 208: You probably don't want to reference a Jira ticket in a public repo

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
```diff
@@ -54,8 +59,155 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
... draggable="true"><article class="css-122osjo" data-test-id="embedded-entry-inline"><div class="css-1sz1u6f" data-card-part="wrapper"><button type="button" aria-label="Actions" aria-haspopup="menu" aria-expanded="false" aria-controls="menu_8727" data-test-id="cf-ui-card-actions" class="css-2ulqgl"><span class="css-k008qs"><svg class="css-1jrff5x" data-test-id="cf-ui-icon" viewBox="0 0 24 24" aria-hidden="true"><path d="M0 0h24v24H0z" fill="none"></path><path d="M6 10c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm12 0c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm-6 0c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2z"></path></svg></span></button><div class="css-123a08z" data-card-part="content"><span class="css-1eadhne"><span>Example Content Type The best article ever</span></span></div></div></article></span><span data-slate-spacer="true" style="height: 0px; color: transparent; outline: none; position: absolute;"><span data-slate-node="text"><span data-slate-leaf="true"><span data-slate-zero-width="z" data-slate-length="0"></span></span></span></span></span><span data-slate-node="text"><span data-slate-leaf="true"><span data-slate-zero-width="z" data-slate-length="0"></span></span></span>',
```
z0al, line 68: Nit (not specific to this PR):  Paste tastes look very cryptic to me. I'd like us to use them only when necessary.

cribbles: I'm split up about it because in a sense we want this code to as closely as possible resemble the expected paste buffer from our own markup. But this comes with the tradeoff of unreadable tests. Let's discuss more today

packages/rich-text/src/plugins/List/index.tsx
cribbles: Yep, ideally we'd be calling the to-text helper here. Should have thrown a TODO in there

packages/rich-text/src/plugins/List/index.tsx
z0al: You may find the `withNormalizer` helper from [this PR](https://github.com/contentful/field-editors/pull/944) useful

cezarsmpio:
Should we address that known "bug" we saw when we were pairing? When adding multiple paragraphs on a list then press enter, the one next to where the cursor is, is dragged to the next LI.

image

Here, I pressed `enter` after the number `1`, so the number `2` was placed on the new LI, which is not I would expect.

I'd expect that number 2, 3 and 4 would be moved to the next LI.

cezarsmpio (approved):
Besides my comment, everything worked great! 🥇

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -163,6 +236,24 @@ const withCustomList = (options): WithOverride => {
     // Use our custom getListInsertFragment
     editor.insertFragment = getListInsertFragment(editor);
 
+    const { normalizeNode } = editor;
+    editor.normalizeNode = (entry) => {
+      const [node, path] = entry;
+
+      if (isList(node as CustomElement)) {
+        normalizeList(editor, path);
+        return;
```
z0al, line 245: We should only return here if `normalizeList` did change the editor structure (by executing the code at line `196`). Otherwise, we will be skipping the call to `normalizeNode(entry)` at the bottom for no reason

When integrated with the other normalization code we have, it will result in some normalizers never being called.

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -163,6 +236,24 @@ const withCustomList = (options): WithOverride => {
...
+      } else if (isListItem(node as CustomElement)) {
+        if (!hasListAsDirectParent(editor, path)) normalizeOrphanedListItem(editor, path);
+        return;
```
z0al, line 248: ditto

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -163,6 +236,24 @@ const withCustomList = (options): WithOverride => {
...
+      } else if (hasListAncestor(editor, path) && !isValidInsideList(node as CustomElement)) {
+        replaceInvalidListItemWithText(editor, path);
+        return;
```
z0al, line 251: ditto

z0al (changes requested):
Sorry for blocking this but I think there are some changes need to be addressed

packages/rich-text/src/plugins/List/index.tsx
cribbles: Fixed in f571378

z0al (approved):
👏🏻 . Let's create a separate issue for the bug Cezar mentioned.

PR #944, opened 2021-11-29
fix: normalize table nodes

djagya:


packages/rich-text/src/plugins/DragAndDrop/index.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { PlatePlugin, getNodes } from '@udecode/plate';
-import { CustomElement } from '../../types';
 import { BLOCKS, CONTAINERS, INLINES } from '@contentful/rich-text-types';
 
+import { CustomElement } from '../../types';
+
 export function createDragAndDropPlugin(): PlatePlugin {
   // Elements that don't allow other elements to be dragged into them and which callback should be used
   const DND_BLOCKED_ELEMENTS = {
-    [BLOCKS.TABLE]: Transforms.removeNodes,
```
z0al, line 9: @cezarsmpio I think that's all we need to do to remove the table normalization logic from here, correct?

cezarsmpio: Correct!

packages/rich-text/src/helpers/editor.ts
```diff
@@ -286,10 +288,15 @@ export function currentSelectionPrecedesTableCell(editor: PlateEditor): boolean
...
+export const replaceNode = (editor: Editor, path: Location, replacement: Node | Node[]) => {
+  Transforms.removeNodes(editor, { at: path });
+  Transforms.insertNodes(editor, replacement, { at: path });
+};
+
 /**
  * It filters out all paragraphs and headings from a path and convert them into paragraphs.
  */
-export function slateNodeEntryToText(editor: PlateEditor, path: Path): CustomElement[] {
+export function getTextAt(editor: PlateEditor, path: Path): CustomElement[] {
```
cezarsmpio, line 299: We need to be careful with the lists PR, it's currently using this helper as the previous name, so it will depend which one gets merged first.

packages/rich-text/src/plugins/Table/normalizers.ts
```diff
@@ -0,0 +1,74 @@
...
+const isTable = (node: CustomElement) => {
+  return node.type === BLOCKS.TABLE;
+};
+
+const isTableCell = (node: CustomElement) => {
+  return node.type === BLOCKS.TABLE_CELL || node.type === BLOCKS.TABLE_HEADER_CELL;
+};
+
+/**
+ * Normalizes TABLE_CELL & TABLE_HEADER_CELL nodes
```
cezarsmpio, line 18: 👍

packages/rich-text/src/plugins/Table/normalizers.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  return true;
+};
+
+export const addTableNormalizers = (editor: PlateEditor) => {
+  withNormalizer(editor, normalizeTable);
+  withNormalizer(editor, normalizeTableCell);
+};
```
cezarsmpio, line 74: Neat 🔥

packages/rich-text/src/test-utils/jsx.ts
```diff
@@ -0,0 +1,33 @@
...
+import { createText } from '@udecode/plate-test-utils';
+import { BLOCKS, INLINES } from '@contentful/rich-text-types';
+
+/**
+ * Mapping for JSX => Slate Node types
+ *
+ * Add items as needed.
+ */
+export const jsx = createHyperscript({
+  elements: {
```
cezarsmpio, line 11: Are we missing HR?

cezarsmpio: What about marks? Not necessary?

packages/rich-text/src/plugins/Table/normalizers.ts
cezarsmpio: Are we forgetting to normalize table rows?

z0al: That's a different ticket

packages/rich-text/src/test-utils/jsx.ts
z0al: Yeah, some items are missing and I'm still not sure how to add marks. But I'm currently working on that to add more tests

packages/rich-text/src/helpers/editor.ts
z0al: It will be a small conflict I think. Anyway, I'm not very happy with the name `getTextAt`, it's better than `slateNodeEntryToText` but doesn't quite describe what the function does

packages/rich-text/src/test-utils/jsx.ts
z0al: I added support for marks with the `htext` tag. You can do `<htext bold> my text </htext>`. I had to add `h` prefix to all tags to prevent colliding with HTML tags for which TypeScripts complains about non-standard attributes

z0al: I will introduce the other items later when integrating the `slate-test-utils`.

packages/rich-text/src/helpers/editor.ts
DanweDE: What about `toParagraphs` ? It's `Paragraph`s, not `Text` nodes that we're returning so I'd opt for more accurate naming. No strong opinion on the rest of the naming.

z0al: Renamed it to `extractParagraphsAt`

packages/rich-text/src/plugins/List/index.tsx
```diff
@@ -175,16 +176,15 @@ const hasListAsDirectParent = (editor: Editor, path: Path) => {
...
   Text.isText(node as TextElement) ||
   (LIST_ITEM_BLOCKS as Array<TopLevelBlockEnum | INLINES>)
     .concat(Object.values(INLINES))
     .includes((node as CustomElement).type as TopLevelBlockEnum);
 
 const replaceInvalidListItemWithText = (editor: PlateEditor, path: Path) => {
-  const textFromEntry = slateNodeEntryToText(editor, path);
-  Transforms.removeNodes(editor, { at: path });
-  Transforms.insertNodes(editor, textFromEntry, { at: path });
+  const textFromEntry = extractParagraphsAt(editor, path);
```
DanweDE, line 186: Nit: Seems like whenever we use `extractParagraphsAt` it's followed by a `replaceNode`, wondering if it would be more slightly more efficient if the function would already do that (assuming there's a slate helper that can be used instead of doing the current `Editor.nodes` call).

DanweDE (changes requested):
Looking good overall, just one or two concerns and some suggestions that might go beyond this PR.

DanweDE:
Just did some testing of pasting all kind of content into table cells and it was all handled well :+1:

packages/rich-text/src/plugins/List/__tests__/normalization.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    // editor.normalizeNode([input, []]) doesn't work
+    Editor.withoutNormalizing(editor, () => {
+      editor.insertText('X');
+      editor.deleteBackward('character');
+    });
+
+    expect(editor.children).toEqual(expected.children);
+  };
+
+  it('wraps orphaned list items in a list', () => {
```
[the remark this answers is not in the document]
z0al: I did the table tests that way first but it got annoying real quick when I had to `.only` some of them. I don't have strong opinion here though

packages/rich-text/src/helpers/normalizers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Injects a custom element normalization handler.
+ *
+ * Handlers must explicity return "true" to indicate a pass.
+ * A pass is when a handler didn't encounter any normalization
+ * paths (i.e. all nodes were valid)
+ *
+ * Important read:
+ * https://docs.slatejs.org/concepts/11-normalizing#multi-pass-normalizing
```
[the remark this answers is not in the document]
z0al: I don't get it. How exactly will be gain some performance boost?

The `return true` in this case is not for performance benefit but this:

> You might at first think this is odd, because with the return there, the original normalizeNodes will never be called, and the built-in constraints won't get a chance to run their own normalization.
>
> But, there's a slight "trick" to normalizing.
> 
> When you do call Editor.unwrapNodes, you're actually changing the content of the node that is currently being normalized. So even though you're ending the current normalization pass, by making a change to the node you're kicking off a new normalization pass. This results in a sort of recursive normalizing.
This multi-pass characteristic makes it much easier to write normalizations, because you only ever have to worry about fixing a single issue at once, and not fixing every possible issue that could be putting a node in an invalid state.

packages/rich-text/src/plugins/Table/index.tsx
```diff
@@ -228,10 +193,12 @@ export const createTablePlugin = (tracking: TrackingProvider) => ({
   onKeyDown: createTableOnKeyDown(),
   withOverrides: (editor) => {
     addTableTrackingEvents(editor, tracking);
-    addTableNormalization(editor);
+    addTableNormalizers(editor);
 
     const { insertFragment } = editor;
 
+    addTableNormalizers(editor);
```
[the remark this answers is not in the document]
z0al: Git merge errors. Will fix it

packages/rich-text/src/plugins/Table/normalizers.ts
```diff
@@ -0,0 +1,74 @@
...
+  // All direct children must be of type TABLE_ROW
+  for (const [child, childPath] of Node.children(editor, path)) {
+    if (!Element.isElement(child)) {
+      continue;
+    }
+
+    const isValidTableChild = CONTAINERS[BLOCKS.TABLE].includes(child.type as BLOCKS);
+
+    if (!isValidTableChild) {
+      Transforms.removeNodes(editor, { at: childPath });
```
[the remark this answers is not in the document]
z0al: We already do remove these nodes (if it ever happened) when cleaning up pasted HTML content. This code maintains that same behavior.

packages/rich-text/src/plugins/List/__tests__/normalization.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const input = (
+      <editor>
+        <hli>
+          <hp>Item</hp>
+          <cursor />
+        </hli>
+        <hp />
+      </editor>
+    );
```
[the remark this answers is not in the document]
z0al: That case was not handled in the list implementation actually. fixed

packages/rich-text/src/plugins/List/index.tsx
z0al: I couldn't find a better way

packages/rich-text/src/plugins/List/__tests__/normalization.test.tsx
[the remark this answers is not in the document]
z0al: I will re-consider that in the `slate-test-utils` work.

packages/rich-text/src/helpers/normalizers.ts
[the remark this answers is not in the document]
z0al: Btw, we don't "control" when a normalization pass kicks in. Even if we return `true` and continue recursively calling `normalizeNode`, slate will still kick in a new normalization because the changes we did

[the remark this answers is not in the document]
DanweDE: As I understand, `return true` results in another full normalization cycle (that's why it's called multi-pass).

If you could prevent this and just finish the current cycle (or "pass") instead, you'd reduce the total number of node checks.
But I guess even if you do this there would be another pass anyways, as you touched a node? No idea, this goes way too deep into Slate internals.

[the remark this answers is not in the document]
z0al: > But I guess even if you do this there would be another pass anyways, as you touched a node?

Yes. Exactly. Slate controls when kick a new pass (when a node is touched) 

We control when to finish the current pass. So it could be better to prevent bugs & performance-wise to exit early. Does that make sense now?

packages/rich-text/src/plugins/List/__tests__/normalization.test.tsx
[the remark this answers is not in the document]
DanweDE: yeah, `only` is pretty convenient vs. commenting out test cases :thinking:

packages/rich-text/src/plugins/List/__tests__/normalization.test.tsx
```diff
@@ -44,6 +44,44 @@ describe('List normalizers', () => {
     assertOutput(input, expected);
   });
 
+  it('adds empty paragraph to empty list items', () => {
```
DanweDE, line 47: :+1:

DanweDE (approved):

PR #945, opened 2021-11-30
feat: create helper to filter out text as paragraphs from slate entries

cezarsmpio:
We don't need to pass the node or a list of nodes as intended previously if we use the helpers provided by Slate/Plate. We only need the editor + path, in this case, I'm receiving the entry (node + path) but I'm not using the `node` at all. 

Question:

- For the developer experience, should we pass only `editor + entry` or `editor + path` as parameters to the helper?

That's how it's supposed to be used:

```ts
function CustomPlugin(): PlatePlugin {
  return {
    withOverrides: (editor) => {
      const { normalizeNode } = editor;

      editor.normalizeNode = (entry) => {
        const [node, path] = entry; // we use only path from entry

        if (isInvalidCase()) {
          const paragraphs = slateNodeEntryToText(editor, path);
          
          Transforms.insertNodes(editor, paragraphs);
        }

        normalizeNode(entry);
      };

      return editor;
    },
  };
}
```

### DIFF
--- a/packages/rich-text/src/helpers/editor.spec.ts
+++ b/packages/rich-text/src/helpers/editor.spec.ts
@@ -1,0 +1,269 @@
+import { createEditor as createSlateEditor } from '@udecode/plate-test-utils';
+import { PlateEditor } from '@udecode/plate';
+import { NodeEntry } from 'slate';
+import { BLOCKS, INLINES } from '@contentful/rich-text-types';
+import { slateNodeEntryToText } from './editor';
+import { CustomElement } from '../types';
+
+const createEditor = (children: CustomElement[]) =>
+  createSlateEditor('test-editor', {}, children) as PlateEditor;
+
+type Text = {
+  text: string;
+  bold?: boolean;
+  italic?: boolean;
+  underline?: boolean;
+  code?: boolean;
+  data?: Record<string, unknown>;
+};
+type Inline = {
+  children: Text[];
+  type: INLINES;
+  data: Record<string, unknown>;
+};
+type TextOrInline = Text | Inline;
+const buildParagraph = (children: TextOrInline[] = []) => ({
+  type: BLOCKS.PARAGRAPH,
+  data: {},
+  isVoid: false,
+  children: children.map((child) => ({ data: {}, ...child })),
+});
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/helpers/editor.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    expect(slateNodeEntryToText(editor, entry)).toEqual([
+      paragraph('text 1'),
+      paragraph('text 2'),
+      paragraph('text 3'),
+    ]);
+  });
+
+  it('list - OL', () => {
+    const ol: CustomElement = {
```
DanweDE, line 155: Almost seems like an exact copy of the one above.
Why not create a function that returns your mock like

```ts
const newList(listType: BLOCKS.UL_LIST|BLOCKS.OL_LIST) => ({
  data: {},
  isVoid: false,
  type: listType,
  children: [
    { ...
```

or perhaps even just combine the two cases and test with a nested list that contains both types.

packages/rich-text/src/helpers/editor.ts
```diff
@@ -285,3 +294,26 @@ export function currentSelectionPrecedesTableCell(editor: PlateEditor): boolean
     !!nextNode && TABLE_BLOCKS.includes(nextNode.type as BLOCKS) && isAtEndOfTextSelection(editor)
   );
 }
+
+/**
+ * It filters out all paragraphs and headings from a node entry (node + path) based on its path and convert them into paragraphs.
+ */
+export function slateNodeEntryToText(
+  editor: PlateEditor,
+  nodeEntry: SlateNodeEntry
```
DanweDE, line 303: Why require an entire `NodeEntry` here if all we use is its `path`?

cezarsmpio: That's what my question on the description is about.

packages/rich-text/src/helpers/editor.spec.ts
cezarsmpio: Sure :)

z0al (approved):
LGTM. Great work

djagya (commented):
👏 👏

PR #947, opened 2021-12-01
fix: separate dnd action in a separate history batch

djagya:
Addresses the bug, where Drag&Drop actions are not put in a separate history batch, so undoing a dnd event undoes the whole batch - leading to content deletion for example. The demonstration as a bug report in Slate: https://github.com/ianstormtaylor/slate/issues/4694

- [x] tests simulating a drop event (follow https://github.com/contentful/field-editors/pull/944 tests)


cezarsmpio:
Not sure if this covers all the situation, look at this example: http://recordit.co/kAJpqQ0Hx3

I'd expect the message to come back to `hello world`, not `jell`. What you think?

z0al (approved):

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -103,21 +103,57 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+      richText.expectValue(
+        doc(entryBlock(), block(BLOCKS.PARAGRAPH, {}, text('some text.')), emptyParagraph())
+      );
+
+      // undo
+      // Ensures that drag&drop was recoreded in a separate history batch,
+      // undoing should not delete the entry block.
+      // See the Slate bug report: https://github.com/ianstormtaylor/slate/issues/4694
+      richText.editor.click().type(`{${mod}}z`).click();
+      richText.expectValue(doc(paragraph, entryBlock(), emptyParagraph()));
```
DanweDE, line 155:
```suggestion
      richText.expectValue(docBeforeDragAndDrop)
```

or something like that if you want to be super expressive :laughing:

DanweDE (approved):

PR #948, opened 2021-12-03
fix: disable hyperlink text when editing []

cezarsmpio:


packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
```diff
@@ -119,15 +119,17 @@ export function HyperlinkModal(props: HyperlinkModalProps) {
...
-            <FormControl id="link-text" isRequired>
-              <FormControl.Label>Link text</FormControl.Label>
-              <TextInput
-                testId="link-text-input"
-                name="link-text"
-                value={linkText}
-                onChange={(event) => setLinkText(event.target.value)}
-              />
-            </FormControl>
+            {!props.linkType && (
```
z0al, line 122: Are we sure that's the only way to detect "editing" mode?

z0al (approved):

packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx
cezarsmpio: Not sure it's the only way but it's enough https://github.com/contentful/field-editors/blob/2f75688b640efb39506ba395041fd333098b8077/packages/rich-text/src/plugins/Hyperlink/HyperlinkModal.tsx#L239

PR #950, opened 2021-12-03
fix: replace @contentful/hostname-transformer

cezarsmpio:


packages/markdown/src/utils/insertAssetLinks.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { replaceAssetDomain } from './insertAssetLinks';
```
cezarsmpio, line 1: We also have Cypress test to reinforce this behavior.

z0al:
FYI: changed the type to "fix:" to force publishing a new release

z0al (approved):
Assuming we can't make the `@contentful/hostname-transformer` public, this is a good alternative 👍🏻

PR #963, opened 2021-12-16
fix: disable edit/remove actions on readonly mode

z0al:


packages/rich-text/src/plugins/Table/TableActions.tsx
```diff
@@ -94,6 +95,10 @@ export const TableActions = () => {
     [editor, isHeaderEnabled, close, onViewportAction]
   );
 
+  if (isDisabled) {
```
z0al, line 98: we were unable to add a unit or cypress test for this 😞

djagya (approved):
🏆

PR #966, opened 2021-12-21
fix: download image no longer opens a new tab

Cyberxon:


packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
```diff
@@ -30,8 +30,24 @@ interface FetchingWrappedAssetCardProps {
...
+    method: 'GET',
+    headers: {},
+  })
+    .then((response) => {
+      response.arrayBuffer().then(function (buffer) {
+        const url = window.URL.createObjectURL(new Blob([buffer]));
+        const link = document.createElement('a');
+        link.href = url;
+        link.setAttribute('download', fileName); //or any other extension
+        document.body.appendChild(link);
```
z0al, line 44: don't we want to remove this new `a` element afterward?

packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
```diff
@@ -30,8 +30,24 @@ interface FetchingWrappedAssetCardProps {
   sdk: FieldExtensionSDK;
 }
 
-function downloadAsset(url: string) {
-  window.open(url, '_blank', 'noopener,noreferrer');
+function downloadAsset(url: string, fileName: string) {
+  fetch(url, {
```
z0al, line 34: I think this won't work when the target has [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) disabled, right?

packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
```diff
@@ -30,8 +30,24 @@ interface FetchingWrappedAssetCardProps {
...
+        const url = window.URL.createObjectURL(new Blob([buffer]));
+        const link = document.createElement('a');
+        link.href = url;
+        link.setAttribute('download', fileName); //or any other extension
+        document.body.appendChild(link);
+        link.click();
+      });
+    })
+    .catch((err) => {
+      console.log(err);
```
djagya, line 49: Maybe show a Froam36 notification like "Failed to download the asset" AND re-throw the error so the app can catch it and report to Sentry. Worth checking if throwing an error crashes only the asset card or the whole app.

packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
```diff
@@ -30,8 +30,27 @@ interface FetchingWrappedAssetCardProps {
...
+        const link = document.createElement('a');
+        link.href = url;
+        link.setAttribute('download', fileName); //or any other extension
+        document.body.appendChild(link);
+        link.click();
+        document.body.removeChild(link);
+      });
+    })
+    .catch((err) => {
+      Notification.error('Failed to dowmload asset');
```
z0al, line 51:
```suggestion
      Notification.error('Failed to download asset');
```

z0al (approved):

PR #979, opened 2022-01-05
fix: normalize empty links

cezarsmpio:


packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -285,5 +287,28 @@ export const createHyperlinkPlugin = (sdk: FieldExtensionSDK): RichTextPlugin =>
         },
       },
     ],
+    normalizer: [
+      {
+        match: {
+          type: INLINES.HYPERLINK,
+        },
+        validNode: hasText,
+        transform: transformRemove,
```
z0al, line 296: * You can drop the `transformRemove` part because it's the default transformer 😉
* Another option would be to have a single `normalizer` rule similar on the top level like this:

```javascript
normalizer: [
  {
    match: {
      type: [INLINES.HYPERLINK, INLINES.ENTRY_HYPERLINK, INLINES.ASSET_HYPERLINK]
    },
    validNode: hasText,
  },
],
```

z0al (commented):
LGTM. Just a small comment

packages/rich-text/src/plugins/Hyperlink/index.tsx
cezarsmpio: - I think it's good to be explicit so it matches my expectations when looking at the code to know what's going on.
- I was wondering if adding an array with multiple types would work, thanks

z0al: > I think it's good to be explicit so it matches my expectations when looking at the code to know what's going on.

Fair enough. In my head, I think of it like: "Every invalid element should be removed, tell me when there is an exception"

packages/rich-text/src/plugins/Hyperlink/__tests__/createHyperlinkPlugin.test.tsx
```diff
@@ -0,0 +1,39 @@
+/** @jsx jsx */
+import { jsx, assertOutput } from '../../../test-utils';
+
+describe('normalization', () => {
+  it('removes empty links from the document structure', () => {
+    const input = (
+      <editor>
+        <hp>
+          <htext>link</htext>
+          <hlink uri="https://link.com" />
```
z0al, line 10: Can we add a link here with a value of just whitespaces, explicitly? e.g. ` <hlink uri="https://link.com">{'    '}</hlink>`. I not sure if these will be trimmed automatically on comparison in the `isAncestorEmpty` helper.

cezarsmpio: Sure, I will try. Right now I can't add a link when selecting a text.

PR #981, opened 2022-01-06
feat: add docx deserialize plugin

cezarsmpio:
Since I use Mac, I don't have a Microsoft Word app installed so I've tested it by exporting a .docx document from Google Docs and opening on my mac using the default app for this kind of document and creating a new document using the Microsoft Word online.

Both ways worked the same with and without the plugin. @z0al any suggestion?

Word online
image

RTE with deserialization
image


z0al:
> Both ways worked the same with and without the plugin

https://github.com/udecode/plate/issues/1001

cezarsmpio:
> @cezarsmpio you can sign up for [Office Online](https://office.live.com/start/Word.aspx) for free. That's what I did when testing MS Excel because I'm on Linux 🤷🏻‍♂️

yes, I did this too.

z0al (approved):
LGTM. but please add a snapshot test for a word file with marks, lists ..etc

z0al (approved):

PR #982, opened 2022-01-06
fix: select embedded blocks & HR on backspace

z0al:
Uses [@udecode/plate-select](https://github.com/udecode/plate/tree/main/packages/editor/select)


cezarsmpio (approved):
👍

PR #983, opened 2022-01-06
fix: allow LIST_ITEM_BLOCKS inside list items

z0al:
A regression caused by the Plate upgrade. We were incorrectly passing the `validLiChildrenTypes` option which caused Plate's list plugin to remove valid items e.g. blockquotes.

### DIFF
--- a/packages/rich-text/src/plugins/Quote/utils.ts
+++ b/packages/rich-text/src/plugins/Quote/utils.ts
@@ -7,51 +7,53 @@ import { CustomElement } from '../../types';
 
 const LIST_TYPES: BLOCKS[] = [BLOCKS.OL_LIST, BLOCKS.UL_LIST];
 
-export function toggleBlock(editor, type: string): void {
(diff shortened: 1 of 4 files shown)


packages/rich-text/src/plugins/Quote/utils.ts
```diff
@@ -7,51 +7,53 @@ import { CustomElement } from '../../types';
...
-    match: (node) => {
-      if (Editor.isEditor(node) || !Element.isElement(node)) {
-        return false;
-      }
-
-      // Lists
-      if (isList && LIST_TYPES.includes((node as CustomElement).type as BLOCKS)) {
-        return true;
-      }
+  Editor.withoutNormalizing(editor, () => {
```
z0al, line 15: This change is not relevant to this PR specifically, I just noticed it when trying to figure out the issue and thought it makes sense.

cribbles (approved):

PR #986, opened 2022-01-07
fix: pressing enter on multi-item list

cribbles:
Fixes a bug in which pressing enter on a list item with multiple children behaves unintuitively.

Logic works fine, still needs tests and cleanup (naming, etc -- this required some fairly hairy imperative node traversal).


z0al (commented):
You may find some of the helper [queries](https://github.com/udecode/plate/tree/main/packages/nodes/list/src/queries) & [transforms](https://github.com/udecode/plate/tree/main/packages/nodes/list/src/transforms) defined in the `@udecode/plate-list` helpful.

PR #991, opened 2022-01-17
fix(rich-text): rewrite insertListItem to handle multiple <li> children

z0al:
Taking [insertListItem](https://github.com/udecode/plate/blob/main/packages/nodes/list/src/transforms/insertListItem.ts) as a starting point. This PR rewrites the helper to handle `<li>` with multiple children. Previously, the code assumed a single paragraph.


packages/rich-text/src/plugins/List/__tests__/createListPlugin.test.tsx
```diff
@@ -105,3 +105,228 @@ describe('normalization', () => {
     assertOutput({ input, expected });
   });
 });
+
+describe('insertBreak', () => {
```
DanweDE, line 109: I really like this, lots of detailed case tested here without the overhead of Cypress :+1:

**Nit:** Wouldn't it more sense to make this more human-readable rather than referring to the function we use to simulate a paragraph? E.g.

```suggestion
describe('insert line break', () => {
```

packages/rich-text/src/plugins/List/getListInsertBreak.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Credit: Modified version of Plate's list plugin
```
DanweDE, line 2: Would be really useful to read a sentence about why we're doing this.

DanweDE:
Seems to work fine for paragraphs, just not in case you're hitting enter after a heading:
enter-in-list-with-heading

DanweDE (approved):
Looking good actually, just some nits and the fact that the the same problem still exists for headings.

cypress/integration/rich-text/RichTextLists.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+      it('should add a new list', () => {
+        richText.editor.click();
+
+        test.getList().click();
+        // TODO: Find a way to test deeper lists
+        /*
+					Having issues with `.type('{enter})` to break lines.
+					The error is:
+					Cannot resolve a Slate node from DOM node: [object HTMLSpanElement]
+				*/
```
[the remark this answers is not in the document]
z0al: The formatting is fine. Again, this is a new file. I cut & pasted from the main RT spec

z0al:
> Seems to work fine for paragraphs, just not in case you're hitting enter after a heading

Good catch. I think I know why. Though, it's an existing bug

packages/rich-text/src/plugins/List/__tests__/createListPlugin.test.tsx
z0al: I usually use the function name in unit tests to easily find it later. No strong opinion here. Will change it

z0al:
> Seems to work fine for paragraphs, just not in case you're hitting enter after a heading

Fixed in the latest commit

PR #992, opened 2022-01-18
feat: entry fetched callback in links

IBratWurst:
Restore tracking of linked entities rendering in the new rich text editor

### DIFF
--- a/cypress/integration/rich-text/RichTextEditor.spec.ts
+++ b/cypress/integration/rich-text/RichTextEditor.spec.ts
@@ -8,6 +8,7 @@ import {
   inline,
   text,
 } from '../../../packages/rich-text/src/helpers/nodeFactory';
+import documentWithLinks from './documentWithLinks';
 import { RichTextPage } from './RichTextPage';
 
 // the sticky toolbar gets in the way of some of the tests, therefore
@@ -1663,5 +1664,24 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
         });
       });
     }
+  });
+
+  describe('on action callback', () => {
+    it('is invoked callback when rendering links', () => {
+      cy.setInitialValue(documentWithLinks);
+      cy.editorActions().should('be.empty');
+      // Necessary for reading the correct LocalStorage values as we do
+      // the initial page load on the beforeEach hook
+      cy.reload();
+      cy.editorActions().should(
+        'deep.equal',
+        new Array(5).fill([
+          'linkRendered',
+          {
+            origin: 'viewport-interaction',
+          },
+        ])
+      );
+    });
   });
 });
--- a/packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
@@ -6,9 +6,11 @@ import { KeyboardHandler, HotkeyPlugin } from '@udecode/plate-core';
 import isHotkey from 'is-hotkey';
 import noop from 'lodash/noop';
 import { Transforms } from 'slate';
+import { TrackingProvider } from 'TrackingProvider';
 
 import { getNodeEntryFromSelection } from '../../helpers/editor';
 import { RichTextPlugin, CustomElement } from '../../types';
+import { withLinkTracking } from '../links-tracking';
 import { LinkedEntityBlock } from './LinkedEntityBlock';
 import { selectEntityAndInsert } from './Util';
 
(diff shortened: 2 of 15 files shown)


packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
```diff
@@ -45,14 +47,27 @@ function getWithEmbeddedEntityEvents(
     };
 }
 
+function withLinkRenderedTracking(tracking: TrackingProvider) {
+  return function LinkedEntityBlockWithTracking(props: ComponentProps<typeof LinkedEntityBlock>) {
+    return (
+      <LinkedEntityBlock
+        onEntityFetchComplete={() => {
+          tracking.onViewportAction('linkRendered');
```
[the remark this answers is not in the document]
IBratWurst: I noticed that as well but I did not find a place where that info is read upon. In the web app we only count how many 'linkRendered' actions are emitted against the number of links in the field. Once the numbers match the tracking data are sent and don't contain any data from the single actions.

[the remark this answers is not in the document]
z0al: Let's ping @DanweDE to double-check. As for me, it looks fine

z0al (approved):

DanweDE (changes requested):

packages/rich-text/src/plugins/EmbeddedEntityBlock/index.tsx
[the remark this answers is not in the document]
DanweDE: Yes, I don't think we're using it. What is `key` even in this context? `${entityType}:${entityId}`? It would make sense to send this info eventually as it would help us to address this TODO [here](https://github.com/contentful/user_interface/blob/7be464e71f2996773a78116220700cfbf6341591/src/javascripts/app/entity_editor/LoadEventTracker.js#L119) but not necessary right now.

DanweDE:
How do you feel about a meaningful Cypress test in to test this on an entity of each type?
This would require us to load the RTE with a RT doc upfront in `sdk.field.getValue()` :thinking: That's the same idea we'll be facing in [this ticket](https://contentful.atlassian.net/browse/SHE-653) actually.

packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -101,7 +104,10 @@ function EntityHyperlink(props: HyperlinkElementProps) {
   const isReadOnly = useReadOnly();
   const sdk: FieldExtensionSDK = useSdkContext();
   const { target } = props.element.data;
-
+  const { onEntityFetchComplete } = props;
+  React.useEffect(() => {
+    onEntityFetchComplete?.();
```
DanweDE, line 109: Maybe leave a comment here that we do this immediately as we lazy load the entity and therefore consider this ready with the first render.

packages/rich-text/src/plugins/shared/FetchingWrappedEntryCard.tsx
```diff
@@ -67,6 +69,13 @@ export function FetchingWrappedEntryCard(props: FetchingWrappedEntryCardProps) {
     getOrLoadEntry(props.entryId);
   }, [props.entryId]); // eslint-disable-line
 
+  React.useEffect(() => {
+    if (!entry) {
```
DanweDE, line 73: Just wondering now... what happens if the link to the entry can't be resolved (missing permissions or the entry was deleted) ? I guess `entry === 'failed` in this case? That seems a bit confusing but ok if it works :shrug:

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -139,6 +140,23 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
     richText.toolbar.embedDropdown.should('be.disabled');
   });
 
+  it('invoke onAction callback when rendering links', () => {
+    cy.setInitialValue(documentWithLinks);
+
+    // Necessary for reading the correct LocalStorage values as we do
+    // the initial page load on the beforeEach hook
+    cy.reload();
+    cy.editorActions().should(
```
DanweDE, line 149: Ideally, we would assert that this only fires once the links were actually fetched as the purpose is to measure how long that takes. So if these were to be fired early, the test should fail. Would be great if it's not too complex, otherwise let's focus on other fixes.

DanweDE (approved):

packages/rich-text/src/plugins/shared/FetchingWrappedEntryCard.tsx
IBratWurst: yes when fetching fails `entry === 'failed'`. It is "complete" to some extent 🤔

PR #994, opened 2022-01-18
perf(rich-text): avoid using Editor.nodes

z0al:
Replaced any use of `Editor.nodes` with `getAbove` or `getNode` ..etc. Basically, `Editor.nodes` iterates over all nodes in the document. We don't need/want that.


z0al:
To be followed up with different PRs

PR #999, opened 2022-01-20
fix: move cursor to previous point after heading choice

IBratWurst:


packages/rich-text/src/plugins/Heading/components/ToolbarHeadingButton.tsx
```diff
@@ -97,12 +96,16 @@ export function ToolbarHeadingButton(props: ToolbarHeadingButtonProps) {
         unwrapFromRoot(editor);
       }
 
-      toggleNodeType(editor, { activeType: type, inactiveType: type });
-
-      // TODO: Figure out why focus only works with timeout here.
-      setTimeout(() => {
+      const prevOnChange = editor.onChange;
+      /* Wait for the change event before focusing
```
z0al, line 100: what change event are we waiting for here?

IBratWurst: The one produced by `toggleNodeType` below (If I am not missing something  😃)

z0al: Do you know why do we need to wait for the change to happen before focusing on the editor? I can't make sense of it :)

z0al (approved):
Approving as it works but would be a good idea to add a comment or two explaining the "why"

packages/rich-text/src/plugins/Heading/components/ToolbarHeadingButton.tsx
IBratWurst: I don't, but a possibility is that while the value is changing the browser might find the field into as apparently empty  (safari) or without the text line being changed (chrome).
I got this impression while working on the ticket: I tried to place the cursor at a specific position of the text after the node type switch. If I did that immediately after, the cursor stayed at the beginning of the text (as per bug description) like if there was no text at all, but I waited an arbitrary little bit then I could get the cursor at the right position.
This suggested me that the focus was requested at the "wrong" time

PR #1001, opened 2022-01-21
fix: normalize initial rte value

IBratWurst:
With initial value normalization we can avoid the rte to crash on invalid document structures


cezarsmpio (approved):

cezarsmpio (approved):

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -1527,4 +1528,17 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+
+  describe('invalid document structure', () => {
+    describe('normalizable errors', () => {
+      it('does not throw', () => {
+        cy.setInitialValue(invalidDocumentNormalizable);
+        cy.reload();
+
+        //check that editors content is what we expect (not a thrown error)
+        richText.editor.findByText('unordered list first item').should('exist');
+        richText.editor.findByText('unordered list second item').should('exist');
```
z0al, line 1540: We should also check there are no validation errors

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -57,11 +57,31 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
...
+    Editor.normalize(editor, { force: true });
+    // We set the value so that the rendering can take over from here
+    setValue(editor.children);
+  }, [props.value, editor]);
+
   return (
     <div className={styles.root} data-test-id="rich-text-editor">
       <Plate
         id={getContentfulEditorId(props.sdk)}
-        initialValue={value}
```
z0al, line 64: Curious, what is the difference between using `initialValue` or `value` in this case?

IBratWurst: Probably I just let it over there after trying different things. It does not seem to make a difference.
Anyway I'd use it now that the first value we pass is an empty document and we change it in the effect, they call it [controlled value](https://plate.udecode.io/docs/Plate#value)

z0al:
@IBratWurst Did you try if using `normalizeinitialvalue` prop [would trigger an `onChange` event](https://github.com/contentful/field-editors/blob/4411939b489d1133388c3fa8485ef25069bf1f4c/packages/rich-text/src/RichTextEditor.jsx#L79-L80)? it it doesn't then it would be a cleaner solution

cypress/integration/rich-text/RichTextEditor.spec.ts
IBratWurst: > @IBratWurst Did you try if using `normalizeinitialvalue` prop [would trigger an `onChange` event](https://github.com/contentful/field-editors/blob/4411939b489d1133388c3fa8485ef25069bf1f4c/packages/rich-text/src/RichTextEditor.jsx#L79-L80)? it it doesn't then it would be a cleaner solution

@z0al I tried out `normalizeinitialvalue` (together with initial value) but unfortunately it didn't work. We get the same error as withoutany normalization

IBratWurst:
> @IBratWurst Did you try if using `normalizeinitialvalue` prop [would trigger an `onChange` event](https://github.com/contentful/field-editors/blob/4411939b489d1133388c3fa8485ef25069bf1f4c/packages/rich-text/src/RichTextEditor.jsx#L79-L80)? it it doesn't then it would be a cleaner solution

@z0al I tried out `normalizeinitialvalue` (together with initial value) but unfortunately it didn't work. We get the same error as without any normalization

PR #1002, opened 2022-01-21
fix: remove empty paragraph at rte start

Cyberxon:


z0al (commented):
Looks good overall. You can add unit tests for the `deleteBackward` & `deleteForward` similar to the `insertBreak` tests [here](https://github.com/contentful/field-editors/blob/4c24368ddd78caccd8bda70c38168ca5f2a6f396/packages/rich-text/src/plugins/List/__tests__/createListPlugin.test.tsx#L109-L412)

cezarsmpio:
Nothing more to add other than what Ahmed has already said. 👍

packages/rich-text/src/plugins/Text/__tests__/createTextPlugin.test.tsx
```diff
@@ -0,0 +1,294 @@
...
+        <hquote>
+          <hp>
+            <cursor />
+            p1
+          </hp>
+        </hquote>
+      ),
+    },
+    {
+      title: 'deletes the empty paragraph at the beginning of the RTE followed by a table',
```
z0al, line 239: This case doesn't actually work when you try it yourself 🤔

z0al (commented):
LGTM overall. Check the case with the table with the team.

packages/rich-text/src/plugins/Text/__tests__/createTextPlugin.test.tsx
Cyberxon: I tested it locally now, it works manually.

z0al: Try it [here](https://deploy-preview-1002--contentful-field-editors.netlify.app/rich-text) (the preview link for this PR). Pressing backspace works but delete doesn't

Cyberxon: seen with the team in PH, they approved it works and that deleting the empty paragraph should be working.

cezarsmpio: I think what @z0al meant is that pressing the key `delete` doesn't delete the empty paragraph. Only the `backspace` key. You can simulate the `delete` key by pressing `fn + backspace`.

cezarsmpio: Should we support delete key as well?

cezarsmpio: Maybe Slate is not triggering when pressing delete in this case?

z0al: Cezar is right. I meant the `delete` key, not backspace. 

Tbh, I'm more concerned about the false positive result of the test. We don't want a test that doesn't cover the real usage as it will be just misleading and bugs will slip through. Are we sure Slate always calls `deleteForward(character)` when pressing the `DELETE` key?

Cyberxon: I see, I understand the case now. I just tested whether that function is being called with a simple console.log in that specific case and it's actually not, which explains why it's not working in the manual test. I need to investigate more to understand which function is actually overriding that one in this case.

z0al (approved):

PR #1004, opened 2022-01-24
fix: remove leading li when pasting inside a list

z0al:


packages/rich-text/src/plugins/List/insertListFragment.ts
```diff
@@ -0,0 +1,121 @@
...
+
+const unwrapTextContainerAtStart = <T extends TDescendant>(nodes: T[]): T[] => {
+  const node = nodes[0];
+
+  if (TEXT_CONTAINERS.includes(node.type)) {
+    return [...node.children, ...nodes.slice(1)];
+  }
+
+  return nodes;
+};
```
z0al, line 72: The actual fix. Everything else is mostly refactoring

packages/rich-text/src/plugins/List/__tests__/insertListFragment.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+    },
+    {
+      title: 'single li with only asset card',
+      input: (
+        <hul>
+          <hli>
+            <hp>
+              hello
+              <cursor />
+              world
```
DanweDE, line 66: I assume this is equal to `hello<cursor />world` without any soft line breaks, right?

packages/rich-text/src/plugins/List/__tests__/insertListFragment.test.tsx
```diff
@@ -0,0 +1,200 @@
+/** @jsx jsx */
+import { jsx, assertOutput, createTestEditor } from '../../../test-utils';
+
+describe('insert fragment', () => {
```
DanweDE, line 4: Amazing set of new tests :+1:

DanweDE (approved):

packages/rich-text/src/plugins/List/__tests__/insertListFragment.test.tsx
z0al: Yes. Prettier decided to format it this way

packages/rich-text/src/plugins/List/components/List.tsx
```diff
@@ -8,10 +8,15 @@ import * as Slate from 'slate-react';
...
   padding: 0;
   margin: 0 0 1.25rem 1.25rem;
+
   div:first-child {
     margin: 0;
     line-height: ${tokens.lineHeightDefault};
   }
+
+  li + li {
+    margin-top: ${tokens.spacingXs};
```
DanweDE, line 18: What's the motivation for this one?

I actually noticed some odd spacing behavior playing around with this:
spacing-issues

DanweDE (changes requested):
There seems to be a spacing issue actually (see comment above)

packages/rich-text/src/plugins/List/components/List.tsx
z0al: Good catch. It was supposed to provide consistent spacing between li elements. I took it from the old editor code but looks like I broke something else. will check 👍🏻

z0al: pushed a fix

DanweDE (approved):

PR #1006, opened 2022-01-25
fix: scroll focus issue on safari

cezarsmpio:
The commit 07b10359007aebce4758c8875b773ae50506d00e tries to fix the issue by restoring it to original position, it's an attempt I think of on how to improve the UX for the user and fix the issue at the same time.

If it turns out not to be a good solution coding-wise and also in a perspective of a product own, we can drop only this commit.

Thoughts? @DanweDE @z0al @djagya @Cyberxon

### DIFF
--- a/packages/rich-text/src/helpers/editor.ts
+++ b/packages/rich-text/src/helpers/editor.ts
@@ -2,8 +2,10 @@ import { Link } from '@contentful/field-editor-reference/dist/types';
 import { BLOCKS, HEADINGS, INLINES, TABLE_BLOCKS } from '@contentful/rich-text-types';
 import { getText, PlateEditor } from '@udecode/plate-core';
 import { Text, Editor, Element, Transforms, Path, Range, Node } from 'slate';
+import { ReactEditor } from 'slate-react';
 
 import { CustomElement } from '../types';
+import { IS_SAFARI } from './environment';
 
 export const LINK_TYPES: INLINES[] = [
   INLINES.HYPERLINK,
@@ -230,4 +232,11 @@ export const INLINE_TYPES = Object.values(INLINES) as string[];
 export const isInlineOrText = (node: Node) => {
   // either text or inline elements
   return Text.isText(node) || (Element.isElement(node) && INLINE_TYPES.includes(node.type));
+};
+
+export const maybeFocus = (editor: PlateEditor) => {
+  // Safari has issues with `editor.focus({ preventScroll: true })`, it ignores the option `preventScroll`
+  if (IS_SAFARI) return;
+
+  ReactEditor.focus(editor);
 };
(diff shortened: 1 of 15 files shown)


packages/rich-text/src/helpers/editor.ts
```diff
@@ -231,3 +233,17 @@ export const isInlineOrText = (node: Node) => {
...
+
+  ReactEditor.focus(editor);
+
+  // Safari has issues with `editor.focus({ preventScroll: true })`, it ignores the option `preventScroll`
+  if (IS_SAFARI) {
+    setTimeout(function () {
+      window.scrollTo(x, y); // restore position
+    }, 0);
+  }
+};
```
cezarsmpio, line 249: The first commit simply ignores Safari

 https://github.com/contentful/field-editors/pull/1006/commits/185b0da497195eae03206b95d107e1d0c0d9a5a6#diff-4b9fbdb58a22376a73c74bceb917c863f1c56d9e02c096332f77a63cbae48358R237-R242

packages/rich-text/src/helpers/editor.ts
```diff
@@ -231,3 +233,17 @@ export const isInlineOrText = (node: Node) => {
...
 };
+
+export const focus = (editor: PlateEditor) => {
+  const x = window.scrollX;
+  const y = window.scrollY;
+
+  ReactEditor.focus(editor);
+
+  // Safari has issues with `editor.focus({ preventScroll: true })`, it ignores the option `preventScroll`
+  if (IS_SAFARI) {
```
z0al, line 244: I don't have access to Safari so I'm curious how it works in a real browser? does the browser scrolls to the top and then get forced (by the code above) to get back to the correct location? wouldn't that be more jarring UX-wise?

cezarsmpio: You can't really see the transition but there are a few tweaks when using table actions, in this case, yes, you can see some transition happening, but that was the only case, everything else works as fine as in other browsers.

packages/rich-text/src/plugins/Table/components/TableActions.tsx
```diff
@@ -41,13 +41,7 @@ export const TableActions = () => {
 
   const close = React.useCallback(() => {
     setOpen(false);
-
-    if (!editor) return;
-
-    // Makes sure we keep the editor in focus when clicking on/out
-    // the dropdown menu
-    Slate.ReactEditor.focus(editor);
```
IBratWurst, line 49: It is expected that we don't focus anymore?

cezarsmpio: Yes.

IBratWurst (approved):

PR #1007, opened 2022-01-25
fix: enter new line in list item after entity block

IBratWurst:
This fixes the behavior of RTE when a user has an entity block selected within a list. Before the block was removed and the cursor moved out of the list.
Instead we want to add a new paragraph to the same list item the block is in

### DIFF
--- a/packages/rich-text/src/RichTextEditor.styles.ts
+++ b/packages/rich-text/src/RichTextEditor.styles.ts
@@ -31,6 +31,7 @@ export const styles = {
     // We need to reset LIC style due to conflicts between PARAGRAPH styles
     'ul > li > div': {
       margin: 0,
+      marginBottom: '1.25rem !important',
     },
   }),
   hiddenToolbar: css({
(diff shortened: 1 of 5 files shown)


packages/rich-text/src/RichTextEditor.styles.ts
```diff
@@ -31,6 +31,7 @@ export const styles = {
     // We need to reset LIC style due to conflicts between PARAGRAPH styles
     'ul > li > div': {
       margin: 0,
+      marginBottom: '1.25rem !important',
```
z0al, line 34: * the margin here is a bit too much. half of that is probably enough.
* How about applying the margin to the void entity blocks directly instead?

packages/rich-text/src/plugins/Voids/createVoidsPlugin.ts
```diff
@@ -20,8 +20,11 @@ export const createVoidsPlugin = (): RichTextPlugin => ({
     {
       // Inserts a new paragraph on enter when a void element is focused
       hotkey: 'enter',
+      // exploit the internal use of Array.slice(0, level + 1) by the exitBreak plugin
+      // to stay in the parent element
+      level: -2,
```
z0al, line 25: Clever trick until we get upstream support 👍🏻

z0al (approved):

packages/rich-text/src/RichTextEditor.styles.ts
IBratWurst: Found a way to do it "better".
That's the same margin of a block at root level. Do we want less space in a list?

PR #1020, opened 2022-02-10
fix(number-editor): use "text" input that syncs numeric value when valid

denkristoffer:
Due to browser implementations, React is unable to fully control inputs of `type="number"`: https://github.com/facebook/react/issues/6556

This commit changes `NumberEditor` to `type="text"` to avoid these issues. We keep a local string input value that we then convert to a number when it's actually valid.

### DIFF
--- a/packages/number/src/NumberEditor.tsx
+++ b/packages/number/src/NumberEditor.tsx
@@ -41,20 +41,28 @@ type InnerNumberEditorProps = Pick<
   field: NumberEditorProps['field'];
 };
 
-function InnerNumberEditor({ disabled, errors, field, setValue, value }: InnerNumberEditorProps) {
-  const previousValue = React.useRef(valueToString(value));
-  const [stringValue, setStringValue] = React.useState(valueToString(value));
+function InnerNumberEditor({
+  disabled,
+  errors,
+  field,
+  setValue,
+  value: sdkValue,
+}: InnerNumberEditorProps) {
+  const previousValue = React.useRef(valueToString(sdkValue));
+  const [inputValue, setInputValue] = React.useState(valueToString(sdkValue));
   const range = getRangeFromField(field);
 
   React.useEffect(() => {
-    previousValue.current = valueToString(value);
-  }, [value]);
+    previousValue.current = valueToString(sdkValue);
+  }, [sdkValue]);
 
   React.useEffect(() => {
-    if (previousValue.current !== String(value) && String(value) !== stringValue) {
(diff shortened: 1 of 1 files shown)


packages/number/src/NumberEditor.tsx
```diff
@@ -26,37 +30,80 @@ function getRangeFromField(field: FieldAPI): RangeValidation {
...
 
+  React.useEffect(() => {
+    previousValue.current = valueToString(value);
+  }, [value]);
+
+  React.useEffect(() => {
+    if (previousValue.current !== String(value) && String(value) !== stringValue) {
+      setStringValue(valueToString(value));
+    }
+  }, [stringValue, value]);
```
bgutsol, line 57: Can you elaborate on this logic pls. Based on code it seems like previousValue.current will be always equal to String(value), because you are updating it on every value change in the useEffect above. Isn't it?

packages/number/src/parseNumber.ts
```diff
@@ -8,11 +8,11 @@ export function parseNumber(
...
-  // For values with chars in 'em, it gives
+  // For values with chars in them, it gives
   // us NaN unlike parseFloat
   const floatVal = +value;
-  const hasDot = /\./g.test(value);
-  const hasFractional = /\.\d+/g.test(value);
+  const hasDot = value.includes('.');
+  const hasFractional = /^(?:\+|-)?\d+\.\d+$/.test(value);
 
   if (isEmpty(value)) {
```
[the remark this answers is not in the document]
denkristoffer: `isEmpty` is old code and I just didn't want to rock the boat 🤷 I assume it might catch cases I didn't think of.

The return is to not mark empty values as invalid in the editor.

packages/number/src/NumberEditor.tsx
denkristoffer: I've updated with some hopefully more clear variable names and a comment which I hope helps.

The first effect is to set a reference that updates when the value converted to a string is changed. This is needed for the second effect, where we check if the value sent from field connector has changed. If it has we can assume that it's valid and then we should update the input value to show it. The comparison with the reference is necessary because the second effect also runs when the input value changes, and in those cases we don't want to trigger the call to update the input value.

denkristoffer: Writing it out like that I see that we could perhaps get the samme outcome from not listing the input value in the effect's dependency array, although that would probably trigger a lint warning.

bgutsol (approved):

PR #1026, opened 2022-02-15
fix: links within safari

IBratWurst:
Recognizes entity hyperlinks copied from safari, which don't have an `href` attribute

### DIFF
--- a/cypress/integration/LocationEditor.spec.ts
+++ b/cypress/integration/LocationEditor.spec.ts
@@ -103,6 +103,7 @@ describe('Location Editor', () => {
   });
 
   it('should set value after using search input', () => {
+    cy.mockGoogleMapsResponse(require('../fixtures/map-search-address-reply.json'));
     cy.editorEvents().should('deep.equal', []);
 
     selectors.getSearchInput().type(LOCATION_2.address);
--- a/packages/rich-text/src/plugins/Hyperlink/index.tsx
+++ b/packages/rich-text/src/plugins/Hyperlink/index.tsx
@@ -182,9 +182,7 @@ export function ToolbarHyperlinkButton(props: ToolbarHyperlinkButtonProps) {
 }
 
 const isAnchor = (element: HTMLElement) =>
-  element.nodeName === 'A' &&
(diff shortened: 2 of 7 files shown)


packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -182,9 +182,7 @@ export function ToolbarHyperlinkButton(props: ToolbarHyperlinkButtonProps) {
 }
 
 const isAnchor = (element: HTMLElement) =>
-  element.nodeName === 'A' &&
-  !!element.getAttribute('href') &&
-  element.getAttribute('href') !== '#';
+  element.nodeName === 'A' && element.getAttribute('href') !== '#';
```
z0al, line 185: This may break the URL hyperlink. Let's have a different `isAnchor` for entity links

IBratWurst: what about only checking for `element.getAttribute('data-link-type') === xxx`. It should be specific already, right?

z0al: It's safer to check the HTML tag as well for entity links but we should only check `href` in case of URL hyperlinks

IBratWurst: Updated!

cypress/integration/LocationEditor.spec.ts
```diff
@@ -103,6 +103,7 @@ describe('Location Editor', () => {
   });
 
   it('should set value after using search input', () => {
+    cy.mockGoogleMapsResponse(require('../fixtures/map-search-address-reply.json'));
```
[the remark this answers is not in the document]
IBratWurst: Ops! I merged at the same time. I thought it made more sense to address it once the test is unskipped tough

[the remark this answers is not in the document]
z0al: We can also do that when we update the next branch with the latest changes in master soon ™️

PR #1027, opened 2022-02-16
chore(tracking): add legacy tracking

cezarsmpio:
Missing Cypress tests. Will do.


packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
```diff
@@ -0,0 +1,45 @@
...
+    const { insertData } = editor;
+
+    editor.insertData = (data) => {
+      const isCopyAndPaste = data.types.length !== 0;
+      if (isCopyAndPaste) {
+        const characterCountSelection = window.getSelection()?.toString().length;
+        const characterCountBefore = getCharacterCount(editor);
+
+        setTimeout(() => {
+          const characterCountAfter = getCharacterCount(editor);
```
z0al, line 30: Why do we need to do `getCharacterCount` twice?

Also, what is the reasoning of `setTimeout`?

packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
```diff
@@ -0,0 +1,45 @@
...
+function getCharacterCount(editor: PlateEditor) {
+  const document = contentfulSlateJSAdapter.toContentfulDocument({
+    document: editor.children,
+    schema: Schema,
+  });
+
+  return documentToPlainTextString(document).length;
+}
+
+export const createTrackingPlugin = (tracking: TrackingProvider): RichTextPlugin => ({
```
z0al, line 18: Since we have a plugin anyway, how do you feel about getting rid of the `tracking` value we pass everywhere (+ the `TrackingProvider` ofc) and instead inject `tracking` property in the `editor` instance?

* Lots of plugins already do something similar so I don't think it's a bad practice
* We already have access to the `editor` instance everywhere in the codebase, whether it's a React component or a pure JS helper.

Something like this:

```javascript
editor.tracking?.logShortcutAction(...)
```

packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
cezarsmpio: I'm following the same logic as here: https://github.com/contentful/field-editors/blob/0ef4f2401a6b175673bab0cda805828f1a637f3a/packages/rich-text/src/plugins/Paste/index.js#L9-L22

> Why do we need to do getCharacterCount twice?

We need to count the number of characters before and after the pasting.

> Also, what is the reasoning behind the setTimeout?

The `setTimeout` delays the execution and it's called after the pasting is done.

cypress/integration/rich-text/RichTextPage.ts
```diff
@@ -120,6 +120,14 @@ export class RichTextPage {
       .should('be.empty')
       .as('validationErrors');
   }
+
+  expectTrackingvalue(expectedValue: any) {
+    cy.window().then((win) => {
+      setTimeout(() => {
+        expect(win.actions).to.deep.equal(expectedValue);
```
z0al, line 127: This will be a source of flakiness the way it's currently written as Cypress won't re-try assertion. [See this guide](https://docs.cypress.io/guides/core-concepts/retry-ability)

cezarsmpio: This has been improved here: [`07173ea` (#1027)](https://github.com/contentful/field-editors/pull/1027/commits/07173ea2807571c2554a8c5d31014ed4c868b9b0)

packages/rich-text/src/plugins/Heading/createHeadingPlugin.ts
```diff
@@ -1,11 +1,23 @@
...
+import { getAbove, toggleNodeType } from '@udecode/plate-core';
+import isHotkey from 'is-hotkey';
 
-import { isInlineOrText } from '../../helpers/editor';
+import { isBlockSelected, isInlineOrText } from '../../helpers/editor';
 import { transformLift, transformUnwrap } from '../../helpers/transformers';
 import { RichTextPlugin } from '../../types';
 import { HeadingComponents } from './components/Heading';
 
+const buildHeadingEventHandler = (type: BLOCKS) => (editor, plugin) => (event) => {
```
z0al, line 10: Please add param types. You may want to use `KeyboardHandler` from `@udecode/plate-core`

cypress/integration/rich-text/RichTextPage.ts
```diff
@@ -120,6 +120,10 @@ export class RichTextPage {
       .should('be.empty')
       .as('validationErrors');
   }
+
+  expectTrackingvalue(expectedValue: any) {
+    cy.window().its('actions', { timeout: 10000 }).should('deep.equal', expectedValue);
```
z0al, line 125: Any reason for the huge timeout? In other words, what is the difference between this assertion and existing ones e.g.  `expectNoValidationErrors`?

packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -72,12 +72,13 @@ function UrlHyperlink(props: HyperlinkElementProps) {
   const isReadOnly = useReadOnly();
   const sdk: FieldExtensionSDK = useSdkContext();
   const { uri } = props.element.data;
+  const tracking = useTrackingContext();
```
z0al, line 75: Do we still need the `TrackingProvider` component and the `useTrackingContext` when we have things available on `editor.tracking`?

packages/rich-text/src/plugins/Table/addTableTrackingEvents.ts
```diff
@@ -34,7 +29,7 @@ export function addTableTrackingEvents(
       setTimeout(() => {
         if (hasTables(markupBefore)) return;
         if (hasTables(markupAfter)) {
-          onViewportAction('paste', {
+          editor.tracking?.onViewportAction('paste', {
             tablePasted: true,
             hasHeadersOutsideFirstRow: hasHeadersOutsideFirstRow(markupAfter),
           });
```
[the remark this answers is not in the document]
DanweDE: No idea, this was led/implemented by Chris.

Maybe the intention back then could've been to just know how many docs use tables?
I'd be all for cleaning this up if you think it makes sense and if the effort is low.

cypress/integration/rich-text/RichTextPage.ts
cezarsmpio: Our CI is kind of slow, so I wanted to give it more time than the current `4000ms`. 

Regarding your second question, I don't see the point... `expectTrackingvalue` does something different, it accesses the `window.actions` property and asserts `expectedValue`. I did it following that link you've sent me.

packages/rich-text/src/plugins/Hyperlink/index.tsx
cezarsmpio: Yeah, I think so.

cypress/integration/rich-text/RichTextPage.ts
z0al: > Regarding your second question, I don't see the point.

Sorry I wasn't clear. I meant the difference in performance, not why we need it.. like why does the `expectTrackingvalue` need more time given that it's very similar to other `expectValue` & `expectNoValidationErrors` ..etc. They all access a value on the window object and assert the result. Why do we need to give the `expectTrackingvalue` more timeout?

packages/rich-text/src/plugins/Hyperlink/index.tsx
z0al: Cool. We can make the tracking plugin accept `onAction` and refactor the code accordingly

cezarsmpio: Yep.

cypress/integration/rich-text/RichTextEditor.Tracking.spec.ts
```diff
@@ -0,0 +1,627 @@
...
+
+      it('adds row above', () => {
+        doAction('Add row above');
+
+        richText.expectTrackingValue([
+          insertTableAction(),
+          insertTableRowAction({
+            tableSize: {
+              numColumns: 2,
+              numRows: 2,
```
z0al, line 234: This doesn't seem right, the default table when we insert is 2x2, after inserting another row I was expecting 3x2. Don't we count table headers?

z0al: The same applies to other tests below

cypress/integration/rich-text/RichTextEditor.Tracking.spec.ts
```diff
@@ -0,0 +1,627 @@
...
+        'shortcut',
+        () => {
+          richText.editor.type(`{${mod}}k`);
+        },
+      ],
+    ];
+
+    for (const [triggerMethod, origin, triggerLinkModal] of methods) {
+      describe(triggerMethod, () => {
+        it('tracks adds and removes hyperlinks', () => {
```
z0al, line 380: I think we are missing a test that opens the hyperlink modal but cancels without adding a link

packages/rich-text/src/ContentfulEditorProvider.ts
```diff
@@ -25,9 +27,19 @@ export function useContentfulEditorId() {
   return id;
 }
 
+// This hook re-render when value value changes
```
z0al, line 30:
```suggestion
// This hook re-renders when the value changes
```

packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
```diff
@@ -20,7 +26,7 @@ export async function selectEntityAndInsert(nodeType, sdk, editor, logAction) {
     }
     Transforms.select(editor, selection);
     insertBlock(editor, nodeType, entity);
-    logAction('insert', { nodeType });
+    logAction?.('insert', { nodeType });
```
z0al, line 29: Based on the function type the `logAction` can't be undefined?

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
```diff
@@ -94,24 +99,40 @@ interface ToolbarEmbeddedEntityInlineButtonProps {
...
+    }, 0);
+
+    logAction('insert', { nodeType: INLINES.EMBEDDED_ENTRY });
+  } catch (error) {
+    if (error) {
+      throw error;
+    } else {
+      logAction('cancelCreateEmbedDialog', { nodeType: INLINES.EMBEDDED_ENTRY });
+    }
+  }
```
z0al, line 135: How does this part work? What do we get when `sdk.dialogs.selectSingleEntry()` rejects?

packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
```diff
@@ -0,0 +1,114 @@
...
+  // Copy & Paste
+  | 'paste'
+  // Hyperlinks
+  | 'cancelCreateHyperlinkDialog'
+  | 'cancelEditHyperlinkDialog'
+  | 'linkRendered'
+  | 'openCreateHyperlinkDialog'
+  | 'openEditHyperlinkDialog'
+  | 'unlinkHyperlinks'
+  // Embeddeds
```
z0al, line 28:
```suggestion
  // Embeds
```

z0al (commented):
Looks great. A few small comments

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
z0al: Based on the [docs](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#dialogs), looks like it doesn't reject at all. Instead, it returns `null`

That means the logging of `cancelCreateEmbedDialog` should be done in the `if(!entry) return` above in line 118

cypress/integration/rich-text/RichTextEditor.Tracking.spec.ts
cezarsmpio: We get the table size before applying the action. I don't know why we do that, we can discuss it with the team.

packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
cezarsmpio: I think I forgot to remove `?.` in this place.

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
cezarsmpio: This method throws an error in reality. I guess the docs are wrong or we have a different version of the SDK.

PR #1028, opened 2022-02-16
test(location): revert changes of lat and lon after mocking

cezarsmpio:
Using values from here: https://github.com/contentful/field-editors/commit/fa51bfaaf1e390d05012ac6371fcda337ca8251c#diff-52e8c5b13cd65b1e1f4cbaf10f52792b8196463b8e13b393bb4d2944e0b47659R48-R51


cypress/integration/LocationEditor.spec.ts
```diff
@@ -4,7 +4,7 @@ const LOCATION_1 = {
 };
 const LOCATION_2 = {
   address: 'Max-Urich-Straße 1, 13355 Berlin, Germany',
-  value: { lon: 13.38372, lat: 52.53926 },
+  value: { lon: 13.3838105, lat: 52.5388562 },
```
z0al, line 7: If it fails try these 👇🏻 (copied from master. Google Maps JS does some weird rounding sometimes)

```suggestion
  value: { lon: 13.38381, lat: 52.53885 },
```

z0al (approved):

PR #1029, opened 2022-02-17
chore: bump rich-text-next version []

IBratWurst:


z0al (approved):
FYI: There is no automatic publishing on `next`. You need to run `npm publish` manually locally. **ONLY** for the `rich-text` package though

PR #1031, opened 2022-02-17
perf(rich-text): debounce value changes

z0al:


z0al:
@cezarsmpio I converted this to draft as I need to fix the tests first.

packages/rich-text/src/useOnValueChanged.ts
```diff
@@ -0,0 +1,46 @@
...
+    () =>
+      debounce((document: unknown) => {
+        handler?.(toContentfulDocument({ document, schema }));
+      }, 500),
+    [handler]
+  );
+
+  return useCallback(
+    (value: unknown) => {
+      const operations = editor.operations.filter(isRelevantOperation);
```
IBratWurst, line 38: Are `editor.operations` cleared after every change?

packages/rich-text/src/useOnValueChanged.ts
```diff
@@ -0,0 +1,46 @@
...
+  editor: PlateEditor;
+  handler?: (value: Document) => unknown;
+};
+
+export const useOnValueChanged = ({ editor, handler }: OnValueChangedProps) => {
+  const onChange = useMemo(
+    () =>
+      debounce((document: unknown) => {
+        handler?.(toContentfulDocument({ document, schema }));
+      }, 500),
```
IBratWurst, line 32: I'd prefer this delay to be configurable, to easily refine it later and/or tweak it in different apps. In future we might even think about making it dynamic to some extent, for example basing it on the content size or on the history of the user typing happening at the same time of the handler run

z0al: I think it's unlikely that we will make use of the configuration given that it has always been that way for years. The way I thought of this is basically: we can always make it configurable later easily but not the other way around (not without a major version bump)

> tweak it in different apps

Do you have a case in mind?

packages/rich-text/src/ContentfulEditorProvider.ts
```diff
@@ -9,15 +10,24 @@ export function getContentfulEditorId(sdk: FieldExtensionSDK) {
...
+
+  return id;
 }
 
-function useContentfulEditorHook({ sdk }: useContentfulEditorHookProps) {
-  const editorId = getContentfulEditorId(sdk);
-  const editor = usePlateEditorRef(editorId);
+export function useContentfulEditor() {
+  const editorId = useContentfulEditorId();
+  const editor = usePlateEditorState(editorId);
```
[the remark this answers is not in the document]
z0al: Wondering how did it work in the past? well, because `onChange` wasn't memoized that caused the creation of a new `editor` instance on every change :)

packages/rich-text/src/useOnValueChanged.ts
IBratWurst: Not any specific one, but, for example, I imagine that compose is more likely to face performance issue than the web app. Thus a more relaxed debounce there would be beneficial.
As you said, though, it's easy to implement later

IBratWurst (approved):

PR #1040, opened 2022-02-23
fix: safari paste

IBratWurst:
Ensure a slate fragment is identified even if `application/x-slate-fragment` is not present in clipboard, as it happens in safari


z0al (approved):

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
```diff
@@ -99,7 +99,8 @@ describe(
         richText.expectSnapshotValue();
       });
 
-      it('pastes only the text content of other blocks', () => {
+      // eslint-disable-next-line mocha/no-skipped-tests
+      it.skip('pastes only the text content of other blocks', () => {
```
z0al, line 103: Why are we skipping this?

IBratWurst: I updated the test with a note, basically this test is not realistic as it doesn't represent a real consistent behavior from a browser. Depending on what exactly is in the clipboard the table may or may not be inserted as text-only.
Different browsers and different positions of the table within the content result in different clipboard content, where `data-slate-fragment` is sometimes omitted.

PR #1041, opened 2022-02-24
fix(rich-text): handle list unwrapping

z0al:


packages/rich-text/src/plugins/List/transforms/toggleList.spec.tsx
```diff
@@ -0,0 +1,395 @@
...
+          </hli>
+        </hul>
+      </editor>
+    ) as any as PlateEditor;
+
+    const expected = (
+      <editor>
+        <hp>1</hp>
+        <hp>2</hp>
+        <hp>3</hp>
```
DanweDE, line 252: Is the focus lost in this case?

packages/rich-text/src/plugins/List/transforms/toggleList.spec.tsx
```diff
@@ -0,0 +1,395 @@
...
+          </hli>
+        </hul>
+      </editor>
+    ) as any as PlateEditor;
+
+    const expected = (
+      <editor>
+        <hol>
+          <hli>
+            <hp>1</hp>
```
DanweDE, line 284: Same question about focus here.

packages/rich-text/src/plugins/List/transforms/toggleList.spec.tsx
```diff
@@ -0,0 +1,395 @@
...
+            <hol>
+              <hli>
+                <hp>11</hp>
```
DanweDE, line 376: This one too, what happens to the focus?

packages/rich-text/src/plugins/List/transforms/toggleList.ts
```diff
@@ -0,0 +1,163 @@
...
+  unwrapNodes,
+  wrapNodes,
+} from '@udecode/plate-core';
+import { ELEMENT_LIC } from '@udecode/plate-list';
+import { getListItemEntry } from '@udecode/plate-list';
+import { Editor, Node, NodeEntry, Path, Range } from 'slate';
+
+const listTypes = [BLOCKS.UL_LIST, BLOCKS.OL_LIST] as string[];
+
+const unwrapList = (editor: PlateEditor, { at }: { at?: Path } = {}) => {
```
DanweDE, line 22: Is this mostly new logic or borrowed from Plate? If the latter, perhaps leave a note.

DanweDE (commented):
Impressive :clap: I've still got to do some testing before approval, some comments for now mainly about the cursor in tests.

packages/rich-text/src/plugins/List/transforms/toggleList.ts
z0al: The whole file is borrowed from Plate. I made some modifications here and there. I will add some comments.

z0al:
> Impressive clap I've still got to do some testing before approval, some comments for now mainly about the cursor in tests.

@DanweDE Keep in mind that I discovered two issues while working on this that I couldn't fix as part of this PR (will require rewriting the `toggleList` file in this PR, which is mostly borrowed from Plate as of now):

* If you select more than one list item children and press the toolbar toggle it won't do anything
* if you put the cursor in only one list item children and press the toolbar toggle it would unwrap only that child, not the full list item like in the old editor.

I will bring those to the team to discuss the priority.

packages/rich-text/src/plugins/List/transforms/toggleList.spec.tsx
z0al: fixed

DanweDE (approved):

PR #1049, opened 2022-03-01
fix: clear rte state on unmount

IBratWurst:
This allows the editor to correctly update once a new version of the entry is given, as the old plate state does not interfere with the new one

### DIFF
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -1,10 +1,10 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect } from 'react';
 
 import { FieldExtensionSDK } from '@contentful/app-sdk';
 import { EntityProvider } from '@contentful/field-editor-reference';
 import { FieldConnector } from '@contentful/field-editor-shared';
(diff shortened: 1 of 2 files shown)


z0al:
@IBratWurst this change had other undesirable side effects we had to [revert it](https://github.com/contentful/field-editors/pull/995).

IBratWurst:
> @IBratWurst this change had other undesirable side effects we had to [revert it](https://github.com/contentful/field-editors/pull/995).

@z0al the new commit clean the plate state when unmounting the editor. It should solve the issue in a different way

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -62,6 +62,11 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
     props.isToolbarHidden && styles.hiddenToolbar
   );
 
+  useEffect(() => {
+    getPlateActions(id).enabled(true);
+    return () => getPlateActions(id).enabled(false);
+  }, [id]);
```
z0al, line 68: We will probably forget why we have this magic here. Let's add a comment to remind us.

z0al (approved):

PR #1050, opened 2022-03-02
fix: correct selection after embedding an entity

IBratWurst:


cezarsmpio (approved):

packages/rich-text/src/plugins/EmbeddedEntityBlock/ToolbarIcon.tsx
```diff
@@ -31,11 +32,12 @@ export function EmbeddedEntityBlockToolbarIcon({
...
   const sdk: FieldExtensionSDK = useSdkContext();
 
-  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+  const handleClick = async (event: React.MouseEvent<HTMLButtonElement>) => {
     event.preventDefault();
 
     onClose();
-    selectEntityAndInsert(nodeType, sdk, editor, editor.tracking.onToolbarAction);
+    await selectEntityAndInsert(nodeType, sdk, editor, editor.tracking.onToolbarAction);
+    moveToTheNextLine(editor);
```
z0al, line 40: This won't work if the entity card is the last element in a list. The reason being the TrailingParagraph plugin ensures we have an extra p only at the root level.

https://user-images.githubusercontent.com/12673605/156346204-82d4118a-3259-4c8e-bded-5261cc5f1fcc.mp4

z0al (commented):
Looks good but there is an issue with lists (and other elements where the entity card won't have a paragraph below it). Also, adding cypress tests would be ideal

packages/rich-text/src/plugins/EmbeddedEntityBlock/ToolbarIcon.tsx
IBratWurst: @z0al I changed it to insert an empty paragraph instead, as in the old editor

z0al (approved):

PR #1051, opened 2022-03-02
fix(rte): reset data when adding headings and paragraphs

cezarsmpio:


packages/rich-text/src/helpers/editor.ts
```diff
@@ -240,3 +245,14 @@ export const focus = (editor: RichTextEditor) => {
...
+
+export function toggleElement(
+  editor: RichTextEditor,
+  options: ToggleNodeTypeOptions,
+  editorOptions?: Omit<EditorNodesOptions, 'match'>
+) {
+  toggleNodeType(editor, options, editorOptions);
+
+  // We must reset `data` from one element to another
+  Transforms.setNodes(editor, { data: {} });
```
z0al, line 257: I think this is unsafe but I could be wrong. @DanweDE this could break for users who used to rely on having custom `data` (for historical reasons) or ?

DanweDE: Correction: Yes, it seems under some circumstances custom data can be allowed. I'll send you some more info.

cezarsmpio: Old editor seems to keep the data property when doing the same action:

image

z0al (approved):

PR #1055, opened 2022-03-03
fix: reset empty list items & quotes on backspace

z0al:


cezarsmpio (approved):
I can't toggle headings into blockquote but this issue is also present on our current next version. Is it something we can fix in this PR?

PR #1061, opened 2022-03-07
fix(rte): pressing backspace on empty paragraphs

cezarsmpio:
This implementation should remove only the first paragraph when empty, it acts as pressing `delete`, other paragraph and headings deletion will be handled by the paragraph and heading original plugins when pressing `delete` and `backspace` because of their different behavior.


DanweDE (approved):
Looking good :+1:

Allowing to delete the first empty paragraph even with backspace probably makes a lot of sense, considering not everyone knows how to delete forward on a Mac :thinking:

PR #1064, opened 2022-03-14
fix: fetch hyperlink entities only once

z0al:


packages/rich-text/src/plugins/Hyperlink/index.tsx
```diff
@@ -1,316 +1,2 @@
...
-  const sdk: FieldExtensionSDK = useSdkContext();
-  const { target } = props.element.data;
-  const { onEntityFetchComplete } = props;
-
-  React.useEffect(() => {
-    // The real entity loading happens in the tooltip
-    // Since that is deferred the link is considered rendered as soon
-    // the component mounts (link text displayed)
-    onEntityFetchComplete?.();
-  }, [onEntityFetchComplete]);
```
z0al, line 118: The `onEntityFetchComplete` will now be called once the actual fetch is complete in the `useEntityInfo` hook. However, I wonder if we should be calling anything at all. I checked master and it seems we don't have the equivalent of `onEntityFetchComplete` for hyperlinks? cc @IBratWurst

DanweDE: I think that was a bug that we fixed in the new RTE compared to the old one.

All node types relying on an entry/asset should report once they received their entity (i.e. it got fetched)

z0al: But the old editor [doesn't report on entity hyperlinks](https://github.com/contentful/field-editors/search?q=linkRendered), only embeds?

packages/rich-text/src/plugins/Hyperlink/components/EntityHyperlink.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { target } = props.element.data;
+  const { onEntityFetchComplete } = props;
+
+  const tooltipContent = useEntityInfo({
+    target,
+    sdk,
+    onEntityFetchComplete,
+  });
+
+  if (!target) return null;
```
cezarsmpio, line 33: Should we return `props.children` for the text to be rendered?

z0al: * The check here is only to avoid TS complaining. It existed previously, I only moved logic around
* It's not possible to end up with a hyperlink node with no target based on our [schema rules](https://github.com/contentful/rich-text/blob/bca1072322e85ce9b3c9cc03d50e2ba232f252f8/packages/rich-text-types/src/nodeTypes.ts#L143-L157)
* If for whatever reason we did, then we also have to account for the possibility of not having `props.children at all and React won't like it when a component returns undefined.

PR #1065, opened 2022-03-15
fix: update rte content on external update

IBratWurst:
Fix for remote value updates not being displayed in a rich text editor

When dealing with the value updates we need to keep in mind 3 scenarios addressed by the refactoring:
- internal updates (local user changes)
- old snapshots being restored
- external updates (remote user changes or first value load)

The last 2 scenarios should not trigger an autosave action


packages/rich-text/src/prepareDocument.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+/**
+ * Converts a contenful rich text document to the corresponding slate editor
+ * value
+ */
+export const documentToEditorValue = (doc: any) => {
+  return toSlatejsDocument({
+    document: hasContent(doc) ? doc : EMPTY_DOCUMENT,
+    schema,
```
DanweDE, line 64: We should delete the schema, it's a remnant from the old RTE without any purpose beyond checking whether a certain node type is void [here](https://github.com/contentful/rich-text/blob/fa79626e4020d9640a920ca5d0ccb654e89cfa90/packages/contentful-slatejs-adapter/src/schema.ts#L52) (and that info should probably just be kept in that repo anyways)

Please leave a TODO or, if time permits, ideally just get rid of the schema.

DanweDE (commented):
Please also add a Cypress test that triggers a `sdk.field.onValueChanged` to simulate a remote change and assert how the editor handles it.

**Edit:** Just saw you've got that in your TODO anyways :sweat_smile:

IBratWurst:
> Please also add a Cypress test that triggers a `sdk.field.onValueChanged` to simulate a remote change and assert how the editor handles it.
> 
> **Edit:** Just saw you've got that in your TODO anyways 😅

@DanweDE it's there now :)

packages/rich-text/src/prepareDocument.ts
IBratWurst: Left  a todo

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -113,10 +133,10 @@ const RichTextEditor = (props: Props) => {
         isInitiallyDisabled={isInitiallyDisabled}
         isEmptyValue={isEmptyValue}
         isEqualValues={deepEquals}>
-        {({ lastRemoteValue, disabled, setValue, externalReset }) => (
+        {({ lastRemoteValue, disabled, setValue }) => (
```
z0al, line 136: Why remove the use of `externalReset`?

z0al (approved):

packages/_test/src/createFakeFieldAPI.ts
```diff
@@ -31,9 +31,13 @@ export function createFakeFieldAPI<T>(
         } else {
           fn = args[0];
         }
-        emitter.on('onValueChanged', fn as Handler);
+        const handler: Handler = (value) => {
+          _value = value;
```
DanweDE, line 35: :+1:

DanweDE (approved):

packages/rich-text/src/RichTextEditor.tsx
z0al: But I think we still need to support `externalReset` somehow, no?

z0al: Since the `externalReset changes [only when the remote value changes](https://github.com/contentful/field-editors/blob/48a50159c968dac5e4d623eda1a7f8a43b2ef292/packages/_shared/src/FieldConnector.ts#L98-L123), I think it could be helpful to fix the exact issue this PR is fixing. We [rely on the that in master](https://github.com/contentful/field-editors/blob/48a50159c968dac5e4d623eda1a7f8a43b2ef292/packages/rich-text/src/RichTextEditor.jsx#L269-L270).

The different ist that the old editor didn't have this "store" logic that Plate has. Did you try resetting/destorying Plate store when the `externalReset` changes? something like (pseudo-code):

```javascript
// RichTextEditor
const lastExternalReset = usePreviousValue(externalReset)

useEffect(()=> {
  cosnt oldId = `${getEditorIdFromSDK(sdk)}-${lastExternalReset}`
  clearPlateStore(oldId)
}, [externalReset, ...])

// in the ConnectedRichTextEditor use id= `${getEditorIdFromSDK(sdk)}-${externalReset}`
```

PR #1066, opened 2022-03-15
feat: show the disable state on fields

MohammedAl-Rowad:
There are "issues" when you disable a field using the `isInitiallyDisabled` prop

 - Some fields will be disabled but the UI does not reflect that ( JSON/Rich Text/ Markdown ).
 - The user can still edit them in a way (Map/Radio).

This PR is to fix that

### DIFF
--- a/packages/_shared/src/FieldConnector.ts
+++ b/packages/_shared/src/FieldConnector.ts
@@ -1,7 +1,8 @@
 import React from 'react';
+
+import { FieldAPI, ValidationError } from '@contentful/app-sdk';
 import isEqual from 'lodash/isEqual';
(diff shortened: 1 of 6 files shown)


z0al (approved):
LGTM 👏🏻

anho (changes requested):
Sorry for blocking, want to clarify something first.

anho:
@MohammedAl-Rowad would you mind turning this into a fix version instead of a feature.

MohammedAl-Rowad:
@anho I replaced this PR with this https://github.com/contentful/field-editors/pull/1071

PR #1067, opened 2022-03-17
fix(rich-text): moving list items down

z0al:
Moves list items to the correct position when pressing `TAB`.

#### Before:

tab-list-previously

#### After:

tab-list


packages/rich-text/src/plugins/List/transforms/moveListItemUp.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (!liParent) {
+      let toListPath;
+      try {
+        toListPath = Path.next(listPath);
+      } catch (err) {
+        return;
+      }
+
+      const condA = hasListChild(editor, liNode);
+      const condB = !isLastChild(list, liPath);
```
cezarsmpio, line 45: What does it mean? Do we have better names?

packages/rich-text/src/plugins/List/transforms/moveListItemUp.ts
```diff
@@ -0,0 +1,139 @@
...
+    const liParent = getAbove<TElement>(editor, {
+      at: listPath,
+      match: { type: getPluginType(editor, ELEMENT_LI) },
+    });
+    if (!liParent) {
+      let toListPath;
+      try {
+        toListPath = Path.next(listPath);
+      } catch (err) {
+        return;
```
cezarsmpio, line 41: Do we do any transformation before returning? I mean, if not, then I think exiting it is safe to do.

packages/rich-text/src/plugins/List/transforms/moveListItemUp.ts
z0al: This file was copied as is from Plate. I made changes on the `moveListItemDown`

packages/rich-text/src/plugins/List/transforms/moveListItemUp.ts
z0al: What do you mean by transformations here?

packages/rich-text/src/plugins/List/transforms/moveListItemUp.ts
z0al: I removed it completely and used the helper directly from Plate. I thought I'd need to modify it at first but not anymore.

packages/rich-text/src/plugins/List/transforms/moveListItemUp.ts
z0al: I got rid of this file

cezarsmpio (approved):

PR #1071, opened 2022-03-17
feat: show the disable state on fields

MohammedAl-Rowad:
There are "issues" when you disable a field using the `isInitiallyDisabled` prop

 - Some fields will be disabled but the UI does not reflect that ( JSON/Rich Text/ Markdown ).
 - The user can still edit them in a way (Map/Radio).

This PR is to fix that

### DIFF
--- a/packages/_shared/src/FieldConnector.ts
+++ b/packages/_shared/src/FieldConnector.ts
@@ -1,7 +1,8 @@
 import React from 'react';
+
+import { FieldAPI, ValidationError } from '@contentful/app-sdk';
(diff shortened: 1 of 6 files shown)


z0al:
@MohammedAl-Rowad I think this needs to be targeting `master` instead of `next`. In `next` we have the new (rewrite of) rich text package, everything else (e.g. JSON editor, Markdown ..etc) is untouched.

z0al:
You may have to recreate your branch based on master to avoid all the conflicts

PR #1075, opened 2022-03-21
feat: introduce isDisabled prop to the field editors

MohammedAl-Rowad:
introducing `isDisabled` prop to all the field editors and reflect the disabled state on them with styles or disabling the actions that could be performed to edit the values

### DIFF
--- a/packages/_shared/src/FieldConnector.ts
+++ b/packages/_shared/src/FieldConnector.ts
@@ -27,6 +27,7 @@ interface FieldConnectorState<ValueType> {
 interface FieldConnectorProps<ValueType> {
   field: FieldAPI;
   isInitiallyDisabled: boolean;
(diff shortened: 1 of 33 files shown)


packages/_shared/src/FieldConnector.ts
```diff
@@ -95,6 +97,13 @@ export class FieldConnector<ValueType> extends React.Component<
     }
   );
 
+  componentDidUpdate(prevProps: FieldConnectorProps<ValueType>) {
+    if (prevProps.isDisabled !== this.props.isDisabled) {
+      // eslint-disable-next-line react/no-did-update-set-state
```
denkristoffer, line 102: Is it necessary to disable the lint rule here? Can't this be done with `getDerivedStateFromProps`?

PR #1079, opened 2022-03-23
fix(rte): pasting tables with containers []

cezarsmpio:


packages/rich-text/src/plugins/Table/createTablePlugin.ts
```diff
@@ -66,10 +71,11 @@ export const createTablePlugin = (): RichTextPlugin =>
       editor.insertFragment = (fragments) => {
         // We need to make sure we have a new, empty and clean paragraph in order to paste tables as-is due to how Slate behaves
         // More info: https://github.com/ianstormtaylor/slate/pull/4489 and https://github.com/ianstormtaylor/slate/issues/4542
-        const fragmentHasTable = fragments.some(
-          (fragment) => (fragment as CustomElement).type === BLOCKS.TABLE
-        );
-        if (fragmentHasTable) {
+        const fragmentHasTable = fragments.some((fragment) => isTable(fragment as CustomElement));
+        const isFirstFragment =
```
cezarsmpio, line 75: We only need to add a paragraph in case table is the first element of a list of multiple other items, e.g. `table + ul + p`, but if we have `p + ul + table`, then we don't need to add a new and empty paragraph.

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
```diff
@@ -179,6 +179,14 @@ describe(
...
...white-space:pre;white-space:pre-wrap;">Cell 2</span></p></td></tr><tr style="height:0pt"><td style="border-left:solid #000000 1pt;border-right:solid #000000 1pt;border-bottom:solid #000000 1pt;border-top:solid #000000 1pt;vertical-align:top;padding:5pt 5pt 5pt 5pt;overflow:hidden;overflow-wrap:break-word;"><p dir="ltr" style="line-height:1.2;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Cell 3</span></p></td><td style="border-left:solid #000000 1pt;border-right:solid #000000 1pt;border-bottom:solid #000000 1pt;border-top:solid #000000 1pt;vertical-align:top;padding:5pt 5pt 5pt 5pt;overflow:hidden;overflow-wrap:break-word;"><p dir="ltr" style="line-height:1.2;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Cell 4</span></p></td></tr></tbody></table></div></b>`,
```
cezarsmpio, line 185: This markup comes from Google Docs.

cezarsmpio: Source: 
image

packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
```diff
@@ -20,6 +20,7 @@ export const sanitizeHTML = (html: string): string => {
     new DOMParser().parseFromString(html, 'text/html')
   );
 
-  // Remove whitespace between tags
-  return doc.body.innerHTML.replace(/>\s+</g, '><');
+  return doc.body.innerHTML
+    .replace(/>\s+</g, '><') // Remove whitespace between tags
+    .replace(/(.*)<div.*>(<table.*<\/table>)<\/div>(.*)/g, '$1$2$3'); // remove div containers from tables
```
cezarsmpio, line 25: Regex experts, feel free to improve it.

IBratWurst, line 25:
```suggestion
    .replace(/<div.*>(<table.*<\/table>)<\/div>/g, '$1'); // remove div containers from tables
```

cezarsmpio: I need what comes before and after the table container because it can include other elements such as lists, headings, paragraphs, etc.

packages/rich-text/src/plugins/Table/createTablePlugin.ts
```diff
@@ -66,10 +72,12 @@ export const createTablePlugin = (): RichTextPlugin =>
...
-        const fragmentHasTable = fragments.some(
-          (fragment) => (fragment as CustomElement).type === BLOCKS.TABLE
-        );
-        if (fragmentHasTable) {
+        const isInsertingTable = fragments.some((fragment) => isTable(fragment as CustomElement));
+        const isTableFirstFragment =
+          fragments.findIndex((fragment) => isTable(fragment as CustomElement)) === 0;
+        const currentLineHasText = getText(editor, editor.selection?.focus.path) !== '';
+
+        if (isInsertingTable && isTableFirstFragment && currentLineHasText) {
```
IBratWurst, line 80: It can't be `isTableFirstFragment: true` if `isInsertingTable: false` right?
```suggestion
        const isTableFirstFragment =
          fragments.findIndex((fragment) => isTable(fragment as CustomElement)) === 0;
        const currentLineHasText = getText(editor, editor.selection?.focus.path) !== '';

        if (isTableFirstFragment && currentLineHasText) {
```

packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
IBratWurst: I thinks js leaves those sections there, only replacing the match

IBratWurst (approved):
Only small comments

packages/rich-text/src/plugins/Table/createTablePlugin.ts
cezarsmpio: It can't be. The other way around is true, I can insert table, but it might not be the first element on the clipboard.

`table` = `isTableFirstFragment: true` and `isInsertingTable: true`
`table + p` = `isTableFirstFragment: true` and `isInsertingTable: true`
`p + ul + table` = `isTableFirstFragment: false` and `isInsertingTable: true`
`p` = `isTableFirstFragment: false` and `isInsertingTable: false`

PR #1082, opened 2022-03-28
fix: untoggling list

IBratWurst:
Fix behaviour of raising list items when switching a list off

### DIFF
--- a/packages/rich-text/src/plugins/List/transforms/unwrapList.ts
+++ b/packages/rich-text/src/plugins/List/transforms/unwrapList.ts
@@ -3,27 +3,32 @@
  * See: https://github.com/udecode/plate/blob/main/packages/nodes/list
  */
 import { BLOCKS } from '@contentful/rich-text-types';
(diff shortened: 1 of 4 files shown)


packages/rich-text/src/plugins/List/transforms/unwrapList.ts
```diff
@@ -3,27 +3,32 @@
...
 export const unwrapList = (editor: PlateEditor, { at }: { at?: Path } = {}) => {
   Editor.withoutNormalizing(editor, () => {
     do {
-      unwrapNodes(editor, {
+      // lift list items to the root level
+      Transforms.liftNodes(editor, {
         at,
-        match: { type: BLOCKS.LIST_ITEM },
-        split: true,
+        match: (node) => Element.isElement(node) && node.type === BLOCKS.LIST_ITEM,
```
z0al, line 22:
```suggestion
        match: match: { type: BLOCKS.LIST_ITEM },
```

z0al (approved):

PR #1086, opened 2022-03-30
fix(rte): pressing enter before a table

cezarsmpio:
This errors happens when putting the text cursor before a table, it doesn't block anything for the user but it's annoying, it also doesn't allow me to write Cypress tests for this feature.

Cypress will complain saying the editor thrown an error.

image

The error is also triggered for any key I press while the cursor is before table.

---

I needed to use `window.getSelection` in this case because Slate's `path` is the same when the cursor is one the first cell or before table, so I couldn't differentiate their positions based on what Slate offers.

The solution was to create a wrapper for the table, so I can identify if the selection is between the container and the table.

### DIFF
--- a/cypress/integration/rich-text/RichTextEditor.spec.ts
+++ b/cypress/integration/rich-text/RichTextEditor.spec.ts
@@ -767,34 +767,37 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
     };
     const expectTable = (...tableElements) => expectDocumentStructure(table(...tableElements));
 
-    it('inserts new line before table', () => {
-      // prevent expected error `Cannot resolve a Slate point from DOM point` from failing this test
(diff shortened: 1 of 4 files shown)


z0al (approved):
Can this be tested with Cypress?

z0al:
> it's testable but Cypress fails the test because the editor throws an error when putting the cursor before table.

@cezarsmpio have you tried [this](https://stackoverflow.com/questions/53845493/cypress-uncaught-assertion-error-despite-cy-onuncaughtexception)?

cezarsmpio:
@z0al cool, just added it [`e1606d9` (#1086)](https://github.com/contentful/field-editors/pull/1086/commits/e1606d905dff3a2b995b3635cfe1c0fa7336a5c1)

PR #1095, opened 2022-04-06
feat: tab key at the last table cell creates a row

z0al:


DanweDE (approved):
Works like a charm :+1:

PR #1104, opened 2022-04-13
chore(tracking): pasting source

cezarsmpio:


packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
```diff
@@ -111,5 +111,43 @@ export const createTrackingPlugin = (onAction: RichTextTrackingActionHandler): R
...
+
+            if (data.includes('docs-internal-guid')) {
+              source = 'Google Docs';
+            }
+
+            if (data.includes('<google-sheets-html-origin>') || data.includes('data-sheets-')) {
+              source = 'Google Spreadsheets';
+            }
+
+            if (data.includes('class="TextRun') || data.includes('class="OutlineElement')) {
```
z0al, line 129: This feels too generic. Is the only way we can tell it's MS Word Online? Also what about normal MS Word and Excel Sheets?

z0al, line 129: This feels too generic. Is the only way we can tell it's MS Word Online? Also what about normal MS Word and Excel Sheets?

z0al: I'd say none of that is reliable so maybe we can ignore the online version. The desktop MS Word however (I tested on Windows VM) adds a ` <meta name=ProgId content=Word.Document>` and ` <meta name=Generator content="Microsoft Word 15">`. Additionally, we also support the `TEXT/RTF` format specifically for MS Word

z0al: Similar to Word, MS Excel on desktop adds ` <meta name=ProgId content=Excel.Sheet> <meta name=Generator content="Microsoft Excel 15">`

cezarsmpio: Thanks for that! I will test if we can extract soem stuff from `TEXT/RTF` as well. I will bring up this question to our PH today. I still think it's safe to use one of those values. The context here is very specific, it happens only when pasting and it needs to be `text/html`, and values like those is not likely to be used  everywhere else.

z0al: MS Word online doesn't use the `text/rtf` format so it's not helpful for that and since the desktop apps add specific metadata  I think it's ok to rely only on `text/html`.

packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
```diff
@@ -111,5 +111,55 @@ export const createTrackingPlugin = (onAction: RichTextTrackingActionHandler): R
...
+
+            // TODO: MS Word Online doesn't give us specific tags, we might need to have a closer look at its tracking result since we are using generic values to identify it
+            if (
+              data.includes('Arial_MSFontService') &&
+              (data.includes('class="TextRun') || data.includes('class="OutlineElement'))
+            ) {
+              source = 'Microsoft Word Online';
+            }
+
+            if (data.includes('<meta name=Generator content="Microsoft Excel')) {
```
z0al, line 137: A more reliable way I think is to use `DomParser().parseFromString` then you could use [CSS selectors like what we do in the HTML plugin](https://github.com/contentful/field-editors/blob/84825818e6684c80cf9862f62d197f77ef54a8f8/packages/rich-text/src/plugins/PasteHTML/utils/sanitizeSheets.ts#L11-L12). Same goes for other `.includes()` calls.

packages/rich-text/src/plugins/Tracking/createTrackingPlugin.ts
```diff
@@ -111,5 +111,55 @@ export const createTrackingPlugin = (onAction: RichTextTrackingActionHandler): R
...
+            if (data.includes('<meta name="Generator" content="Cocoa HTML Writer">')) {
+              source = 'Apple Notes';
+            }
+
+            if (data.includes('Slack-Lato, Slack-Fractions')) {
+              source = 'Slack';
+            }
+
+            if (source) {
+              editor.tracking.onShortcutAction('paste', {
```
z0al, line 154: We have other places where we fire `paste` events e.g. in the table plugin, wouldn't that mean it would fire twice? making the data less real. Can we combine these calls somehow?

z0al (approved):

PR #1106, opened 2022-04-14
fix: better way to track invalid table cell paste

z0al:
The current way doesn't capture things like pasting a block entry in a table


packages/rich-text/src/plugins/Table/tableTracking.ts
```diff
@@ -0,0 +1,58 @@
...
+      insertData(data);
+    }
+  };
+}
+
+export const withInvalidCellChildrenTracking = (transformer: NodeTransformer): NodeTransformer => {
+  return (editor, childEntry) => {
+    const [node] = childEntry;
+
+    if (Element.isElement(node)) {
```
cezarsmpio, line 50: I guess `validChildren` will already filter out invalid elements, right?

cezarsmpio (approved):

packages/rich-text/src/plugins/Table/tableTracking.ts
z0al: Yes. But needed it to make TypeScript happy

PR #1117, opened 2022-04-22
fix(rich-text): prevent editor crash on triple clicks

z0al:


DanweDE (approved):

DanweDE:
Ok, I assume these are the broken Cypress tests you mentioned earlier :)

PR #1120, opened 2022-04-26
feat: reference hide actions prop [DANTE-488]

YvesRijckaert:
Expose two props `hasRemoveActions` and `hasMoveActions ` in `MultipleEntryReferenceEditor` and one prop `hasRemoveActions` in `SingleEntryReferenceEditor` to be able to hide the actions menu completely.
This is to fix a bug when a user without editing permissions could still open the actions menu in the web app.

### DIFF
--- a/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
@@ -141,13 +141,15 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
       onMoveBottom: props.onMoveBottom,
     };
 
-    const { hasCardEditActions } = props;
+    const { hasCardEditActions, hasMoveActions, hasRemoveActions } = props;
 
     function renderDefaultCard(props?: CustomEntityCardProps) {
       const builtinCardProps: WrappedEntryCardProps = {
         ...sharedCardProps,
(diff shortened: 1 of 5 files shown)


denkristoffer:
Hey @YvesRijckaert, can we avoid adding this complexity? Looking at the code it seems like the menu item shouldn't be rendered if `onRemove` is not passed to the editor, would that acheive the same result as this new flag?

YvesRijckaert:
@denkristoffer I made the changes that we discussed. Removed the general `hasActions` prop and exposed these two props instead: `hasRemoveActions` and `hasMoveActions`

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -141,13 +141,15 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
       onMoveBottom: props.onMoveBottom,
     };
 
-    const { hasCardEditActions } = props;
+    const { hasCardEditActions, hasMoveActions, hasRemoveActions } = props;
```
denkristoffer, line 144: I think we should stick to the `hasCard...Actions` naming for all of them here.

denkristoffer (changes requested):
Looks good but I think we should align the naming.

denkristoffer (approved):
Thank you 🙇

PR #1121, opened 2022-04-26
fix(EmbedAsset): always assume that assets are available for linking [HOMER-829]

Chaoste:
Same issue as [here](https://github.com/contentful/field-editors/blob/master/packages/reference/src/common/useEditorPermissions.ts#L49-L62): Tag-based permissions break the access evaluation. We assume that one can always link assets as fallback solution.

### DIFF
--- a/packages/rich-text/src/Toolbar/components/EmbedEntityWidget.tsx
+++ b/packages/rich-text/src/Toolbar/components/EmbedEntityWidget.tsx
@@ -23,19 +23,22 @@ export const EmbedEntityWidget = ({ isDisabled, canInsertBlocks }: EmbedEntityWi
   const onCloseEntityDropdown = () => setEmbedDropdownOpen(false);
   const onToggleEntityDropdown = () => setEmbedDropdownOpen(!isEmbedDropdownOpen);
 
-  const [canAccessAssets, setCanAccessAssets] = useState(false);
-  React.useEffect(() => {
-    let subscribed = true;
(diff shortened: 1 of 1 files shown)


z0al (changes requested):
Sorry for blocking this. Let's chat about it tomorrow

packages/rich-text/src/Toolbar/components/EmbedEntityWidget.tsx
```diff
@@ -23,14 +23,17 @@ export const EmbedEntityWidget = ({ isDisabled, canInsertBlocks }: EmbedEntityWi
   const onCloseEntityDropdown = () => setEmbedDropdownOpen(false);
   const onToggleEntityDropdown = () => setEmbedDropdownOpen(!isEmbedDropdownOpen);
 
-  const [canAccessAssets, setCanAccessAssets] = useState(false);
+  // Hardcoded `true` value following https://contentful.atlassian.net/browse/DANTE-486
+  // TODO: refine permissions check in order to account for tags in rules
+  const [canAccessAssets] = useState(true);
```
z0al, line 28: I think it's safe to get rid of the `canAccessAssets` and the `useEffect` below it.

Chaoste: @IBratWurst advised me to keep the query for monitoring reasons. As soon as the access API would correctly identify tag-based permissions, we want to readd this logic.

z0al (approved):

PR #1123, opened 2022-04-28
fix(rich-text): don't disable "Edit" actions on embeds

z0al:
Users should still be able to edit (or view) embedded entities when the RT editor is disabled.


DanweDE (approved):
Feel like this is one of our "favorite" regressions on all reference cards :laughing:

PR #1125, opened 2022-04-28
fix: reflect disabled state on fields

MohammedAl-Rowad:
Some field editors do not look disabled when you disable them (by setting `isInitiallyDisabled` to true)

- JSON ( looks the same but you can’t edit )
- Rich text ( the courser is not `not-allowed` )
- Mark down ( the courser is not `not-allowed` )

Even after passing `isInitiallyDisabled` as `true`  you can still edit/delete or add values to these fields 

- Map editor (you can still change the marker location (edit))
- Radio (you can click the clear button )
- Tags (you can still delete tags and the courser is not `not-allowed` )

## Related PRs

https://github.com/contentful/field-editors/pull/1128

### DIFF
--- a/packages/radio/src/RadioEditor.tsx
+++ b/packages/radio/src/RadioEditor.tsx
@@ -75,7 +75,11 @@ export function RadioEditor(props: RadioEditorProps) {
                     {item.label}
                   </Radio>
                   {checked && (
-                    <TextLink as="button" className={styles.clearBtn} onClick={clearOption}>
+                    <TextLink
+                      as="button"
+                      className={styles.clearBtn}
+                      onClick={clearOption}
+                      isDisabled={disabled}>
                       Clear
                     </TextLink>
                   )}
--- a/packages/radio/src/RadioEditor.tsx
+++ b/packages/radio/src/RadioEditor.tsx
@@ -74,12 +74,8 @@ export function RadioEditor(props: RadioEditorProps) {
                     }}>
                     {item.label}
                   </Radio>
-                  {checked && (
-                    <TextLink
-                      as="button"
-                      className={styles.clearBtn}
-                      onClick={clearOption}
-                      isDisabled={disabled}>
+                  {checked && !disabled && (
+                    <TextLink as="button" className={styles.clearBtn} onClick={clearOption}>
                       Clear
                     </TextLink>
                   )}
--- a/packages/rich-text/src/RichTextEditor.jsx
+++ b/packages/rich-text/src/RichTextEditor.jsx
@@ -64,6 +64,7 @@ const styles = {
   }),
   disabled: css({
     background: tokens.gray100,
+    cursor: 'not-allowed',
   }),
 };
 
--- a/packages/tags/src/TagsEditor.tsx
+++ b/packages/tags/src/TagsEditor.tsx
@@ -145,9 +145,12 @@ export function TagsEditor(props: TagsEditorProps) {
               label={item}
               index={index}
               key={item + index}
-              // disabled goes for SortableElement
               disabled={isDisabled}
-              // isDisabled goes for SortablePill
+              /**
+               * `isSortablePillDisabled` is needed as SortableElement
+               * from react-sortable-hoc doesn't pass down the disabled prop.
+               * See: https://github.com/clauderic/react-sortable-hoc/issues/612
+               */
               isSortablePillDisabled={isDisabled}
               onRemove={() => {
                 removeItem(index);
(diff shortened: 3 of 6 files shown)


packages/rich-text/src/RichTextEditor.jsx
```diff
@@ -64,6 +64,7 @@ const styles = {
   }),
   disabled: css({
     background: tokens.gray100,
+    cursor: 'not-allowed',
```
z0al, line 67: This change needs to be done in the `next` branch, otherwise, it won't show up in the web app until later when we merge the `next` branch back into `master`.

MohammedAl-Rowad: done:

https://github.com/contentful/field-editors/pull/1128

packages/tags/src/TagsEditor.tsx
```diff
@@ -137,7 +145,10 @@ export function TagsEditor(props: TagsEditorProps) {
               label={item}
               index={index}
               key={item + index}
+              // disabled goes for SortableElement
               disabled={isDisabled}
+              // isDisabled goes for SortablePill
+              isSortablePillDisabled={isDisabled}
```
z0al, line 151: Any reason why a single `disabled` prop won't work? I think we always want the `SortablePill` to be disabled & its children, no?

packages/radio/src/RadioEditor.tsx
```diff
@@ -75,7 +75,11 @@ export function RadioEditor(props: RadioEditorProps) {
                     {item.label}
                   </Radio>
                   {checked && (
-                    <TextLink as="button" className={styles.clearBtn} onClick={clearOption}>
+                    <TextLink
+                      as="button"
+                      className={styles.clearBtn}
+                      onClick={clearOption}
+                      isDisabled={disabled}>
```
z0al, line 82: Nit: I'd prefer to conditionally render the `clear` button here instead e.g. `!disabled && <TextLink .../>` to reduce visual noise

packages/tags/src/TagsEditor.tsx
```diff
@@ -47,31 +55,31 @@ const styles = {
   }),
 };
 
-const SortablePillHandle = SortableHandle(() => (
-  <div className={styles.handle}>
+const SortablePillHandle = SortableHandle((props: { isDisabled: boolean }) => (
+  <div className={`${styles.handle} ${props.isDisabled ? styles.pillDisabled : ''}`}>
```
z0al, line 59: You can use the [cx](https://5faaafd0bd0f3f0008469537--emotion.netlify.app/docs/emotion#cx) helper here instead

packages/tags/src/TagsEditor.tsx
```diff
@@ -47,31 +55,31 @@ const styles = {
...
-  disabled: boolean;
+  isSortablePillDisabled: boolean;
   index: number;
 }
 
 const SortablePill = SortableElement((props: SortablePillProps) => (
   <Pill
     testId="tag-editor-pill"
-    className={styles.pill}
+    className={`${styles.pill} ${props.isSortablePillDisabled ? styles.pillDisabled : ''}`}
```
z0al, line 74: same here. `cx` is a better fit for specifying multiple styles.

z0al (commented):
Looks great. Left small comments

packages/tags/src/TagsEditor.tsx
MohammedAl-Rowad: `<SortablePill />` is a component that uses [SortableElement](https://github.com/clauderic/react-sortable-hoc#sortableelement-hoc) HOC, which will not pass the disabled props down to the `<Pill />` component.


```jsx
const SortablePill = SortableElement((props: SortablePillProps) => (
 // props.disabled is not available here
  <Pill
    testId="tag-editor-pill"
    className={cx(styles.pill, { [styles.pillDisabled]: props.isSortablePillDisabled })}
    label={props.label}
    onClose={() => {
      if (!props.isSortablePillDisabled) {
        props.onRemove(props.index);
      }
    }}
    onDrag={noop}
    dragHandleComponent={<SortablePillHandle isDisabled={props.isSortablePillDisabled} />}
  />
));
```

z0al: Thanks for the explanation. That's unfortunate but it's Ok I guess since the `SortablePill` isn't exposed outside the package. I think it's worth noting that down though as a code comment as it's not very obvious at the moment.

z0al (approved):
Looks great 👍🏻 . Let's add the code comment I mentioned above and :shipit:

packages/tags/src/TagsEditor.tsx
MohammedAl-Rowad: done

PR #1134, opened 2022-05-05
fix: only show readable / creatable CTs in reference field editor [ZEND-2154]

FLoppix:
## Purpose

The `useCallback` prevented updating the creatable / readable CTs since we resolve them asynchronously. Actual readable / creatable CTs were never propagated to the UI.

### DIFF
--- a/packages/reference/src/components/LinkActions/LinkEntityActions.tsx
+++ b/packages/reference/src/components/LinkActions/LinkEntityActions.tsx
@@ -88,6 +88,7 @@ export function useLinkActionsProps(props: LinkEntityActionsProps): LinkActionsP
     [sdk, entityType, onCreated]
   );
 
+  // Wrapping these two with useCallback caused a bug [ZEND-2154] where CTs were not propagated to the UI at all
   const onLinkExisting = async (index?: number) => {
     const entity = await selectSingleEntity({
       sdk,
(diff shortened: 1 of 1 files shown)


packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -88,37 +88,31 @@ export function useLinkActionsProps(props: LinkEntityActionsProps): LinkActionsP
...
-    },
-    [sdk, entityType, onLinkedExisting]
-  );
+    if (!entities || entities.length === 0) {
+      return;
+    }
+    onLinkedExisting(entities, index);
+  };
 
   return useMemo(
```
Chaoste, line 117: Your changes look suspicious to me. Doing that means that this `useMemo` doesn't make sense anymore. If this is a hotfix, I would add a note. Something like `// Use raw callbacks to avoid any memoization`.

if it's urgent, how about a follow-up ticket to re-add `useCallback` and find out why this logic broke in the first place (which rerenders and recalculations are desired and which are not).

@IBratWurst this might be related to the hotfix you introduced recently for the other bug.

IBratWurst: when used correctly, memoized callbacks shouldn't case any issue or need to be removed. The reason for that should be that memoization causes 0 or negative performance gain.

Anyway I don't think this interferes with the hotfix for the disabled link button

Chaoste: Thanks for checking 👍 (I thought it related to an empty list of content types)

Chaoste (approved):

PR #1138, opened 2022-05-09
fix(rich-text): sanitize slate content

z0al:
We rely on initial normalization to prevent crashes when we find e.g. hyperlink without children, however, Slate still crashes in some cases (when attempting to do the normalization) and complains about empty nodes (nodes without text).

In this particular case, having an empty hyperlink is fine (i.e. handled by the normalization) but the editor crashes as soon as that link is followed by a list with exactly two text nodes 🤷🏻‍♂️ 

This PR brings back the old `santizeIncomingSlateDoc` helper 😞 as it seems the most reliable way we got to prevent such errors.

### DIFF
--- a/packages/rich-text/src/helpers/sanitizeIncomingSlateDoc.ts
+++ b/packages/rich-text/src/helpers/sanitizeIncomingSlateDoc.ts
@@ -1,0 +1,25 @@
+import { CustomText, TextOrCustomElement } from '../types';
+
+const isTextElement = (node: TextOrCustomElement): node is CustomText => 'text' in node;
+
+/**
+ * Ensures all nodes have a child leaf text element. This should be handled by
+ * Slate but its behavior has proven to be buggy and unpredictable.
+ */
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/helpers/sanitizeIncomingSlateDoc.ts
```diff
@@ -0,0 +1,25 @@
+import { CustomText, TextOrCustomElement } from '../types';
+
+const isTextElement = (node: TextOrCustomElement): node is CustomText => 'text' in node;
+
+/**
+ * Ensures all nodes have a child leaf text element. This should be handled by
+ * Slate but its behavior has proven to be buggy and unpredictable.
```
DanweDE, line 7: Is there any Slate issue we could link to?

DanweDE (approved):

packages/rich-text/src/helpers/sanitizeIncomingSlateDoc.ts
z0al: I couldn't find any

PR #1141, opened 2022-05-11
fix: made timezone picker match webapp, added guessing [SHE-831]

MayaGillilanContentful:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "mochawesome": "^6.1.1",
     "mochawesome-merge": "^4.0.3",
     "mochawesome-report-generator": "^5.1.0",
+    "moment-timezone": "^0.5.34",
     "patch-package": "^6.4.7",
     "prettier": "2.3.0",
     "react": ">=16.14.0",
--- a/packages/date/package.json
+++ b/packages/date/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/field-editor-date",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "main": "dist/index.js",
   "module": "dist/field-editor-date.esm.js",
   "typings": "dist/index.d.ts",
--- a/packages/date/package.json
+++ b/packages/date/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/field-editor-date",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "main": "dist/index.js",
   "module": "dist/field-editor-date.esm.js",
   "typings": "dist/index.d.ts",
--- a/packages/rich-text/src/Toolbar/components/EmbedEntityWidget.tsx
+++ b/packages/rich-text/src/Toolbar/components/EmbedEntityWidget.tsx
@@ -31,22 +31,6 @@ export const EmbedEntityWidget = ({ isDisabled, canInsertBlocks }: EmbedEntityWi
   const blockAssetEmbedEnabled =
     isNodeTypeEnabled(sdk.field, BLOCKS.EMBEDDED_ASSET) && canInsertBlocks;
 
-  const numEnabledEmbeds = [
-    inlineEntryEmbedEnabled,
-    blockEntryEmbedEnabled,
-    blockAssetEmbedEnabled,
-  ].filter(Boolean).length;
-
-  const shouldDisplayDropdown = numEnabledEmbeds > 1 || isDisabled;
-
-  // Avoids UI glitching when switching back and forth between
-  // different layouts
-  React.useEffect(() => {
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "mochawesome": "^6.1.1",
     "mochawesome-merge": "^4.0.3",
     "mochawesome-report-generator": "^5.1.0",
+    "moment-timezone": "^0.5.34",
```
denkristoffer, line 94: What's the size impact of adding this package? Bundlephobia seems to be down so I don't know of an easy way to check, but in my experience everything related to moment comes at a big size cost.

MayaGillilanContentful: That's a good question, would anyone else maybe know how I could check this?

packages/rich-text/src/Toolbar/components/EmbedEntityWidget.tsx
```diff
@@ -31,54 +31,31 @@ export const EmbedEntityWidget = ({ isDisabled, canInsertBlocks }: EmbedEntityWi
   const blockAssetEmbedEnabled =
     isNodeTypeEnabled(sdk.field, BLOCKS.EMBEDDED_ASSET) && canInsertBlocks;
 
-  const numEnabledEmbeds = [
```
z0al, line 34: These changes were already part of https://github.com/contentful/field-editors/pull/1133

packages/date/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/field-editor-date",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
z0al, line 3: The changes in the date package should be done in `master`. Use `next` only for rich-text changes

package.json
z0al: 100kb gzipped is a lot IMHO. Given that we want to phase away moment at some point I think we could start considering alternatives here. Here are a couple of alternatives you could try:

* [Luxon](https://moment.github.io/luxon/#/moment) - Recommended by Moment maintainers
* [date-fns](https://date-fns.org/) but I'm not sure if it supports timezones
* [dayjs + the timezone plugin](https://day.js.org/docs/en/plugin/timezone)

andipaetzold: It might be possible to do all this without any extra dependency at all:

* Full list of time zones: `Intl.supportedValuesOf("timeZone")`
* The user's current time zone: `Intl.DateTimeFormat().resolvedOptions().timeZone`
* Format a date in a specific time zone: `Intl.DateTimeFormat('en', { timeZone: 'Asia/Tokyo'}).format(new Date())`

-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

MayaGillilanContentful: Then we wouldn't have the guessing functionality which was mentioned as desired in the ticket, although we could just search if we're okay with having something a little less "smart"

andipaetzold: `moment.tz.guess()` is the same as `Intl.DateTimeFormat().resolvedOptions().timeZone`

see https://momentjs.com/timezone/docs/#/using-timezones/guessing-user-timezone/

Sorry for just randomly jumping in this PR. I don't know what the requirements for the ticket are

MayaGillilanContentful: Oh, that's very good to know! I will test that out and see if it works just to do that then. And no worries about jumping in, I appreciate the input

MayaGillilanContentful: After evaluating all the options:

Intl does not offer a convenient way to get offset from a timezone, meaning unnecessary work to write a conversion function that would likely result in too much overhead IMO. It also causes some odd typescript errors but this is a minor issue.

DayJS, while the lightest weight of all the packages, doesn't offer any functionality to list timezones.

Luxon also does not list timezones and is larger than DayJS, but has more functionality otherwise (although not sure that that's really relevant here).

Date-fns does support timezones but seems to have the least functionality of all the options (along with not supporting listing timezones) so I think we can exclude it.

So unless I'm missing something which is very possible, it looks like the main options are to 

- continue using Moment
- do a big rewrite using Intl to avoid any dependencies
- use Intl to list timezones and then DayJS for formatting and guessing (which is safer for old browsers than using only Intl)

andipaetzold: Another interesting package to use is https://www.npmjs.com/package/tzdata, but I also don't know whether that includes all you need.

The type issues you ran into probably happen because DateTimeFormat is fairly new and have not been added to TypeScript yet. Browser support is also not perfect here.

I don't know what's the best option. I would especially try to avoid `moment` as it's a big library and use another library that is already used in other places, so we have some deduping.

PR #1146, opened 2022-05-16
fix: made timezone picker match webapp, added guessing [SHE-831]

MayaGillilanContentful:
image
image


packages/date/src/utils/timezones.ts
```diff
@@ -0,0 +1,20 @@
...
+dayjs.extend(utc);
+dayjs.extend(timezone);
+
+export type Timezone = {
+  displayValue: string;
+  ianaName: string;
+};
+
+//@ts-expect-error
+const timezones = Intl.supportedValuesOf('timeZone').map((timezone: string) => ({
```
z0al, line 15: The support for `Intl.supportedValueOf` is almost non-existent 😞 

image

z0al (commented):
Looks good. We need to figure an alternative to `Intl.supportedValueOf` though

packages/date/src/utils/timezones.ts
z0al: Yeah, I think we don't have a better option. 25kb is better than 100kb 🤷🏻‍♂️ . I'm okay with using that package (a better option would be finding another regularly updated package that gives us only the names - nothing extra - that would save a few kbs)

But after you get the list of TZ names remember to check each if it's supported. We can borrow [some logic from this polyfill](https://github.com/tc39/proposal-intl-enumeration/blob/e09f83795e4dfc4c3d8ad2804f81f7396b7125ae/polyfill/timeZones.js#L100-L110) for that.

MayaGillilanContentful: How does it look to you now?

packages/date/src/utils/timezones.ts
```diff
@@ -0,0 +1,36 @@
...
+
+dayjs.extend(utc);
+dayjs.extend(timezone);
+
+export type Timezone = {
+  displayValue: string;
+  ianaName: string;
+};
+
+function isSupported(timeZone: string): boolean {
```
z0al, line 16:
```suggestion
// Borrowed from https://github.com/tc39/proposal-intl-enumeration/blob/master/polyfill/timeZones.js
function isSupported(timeZone: string): boolean {
```

z0al (approved):

denkristoffer (approved):

package.json
```diff
@@ -48,7 +48,7 @@
     "@commitlint/config-conventional": "^12.0.1",
     "@contentful/contentful-extension-scripts": "^0.20.4",
     "@contentful/eslint-config-extension": "0.4.2",
-    "@contentful/f36-components": "^4.9.0",
```
MohammedAl-Rowad: r u sure 🙄?

packages/date/package.json
```diff
@@ -23,7 +23,7 @@
     "test:ci": "tsdx test --env=jsdom --ci"
   },
   "dependencies": {
-    "@contentful/f36-components": "^4.0.27",
```
MohammedAl-Rowad, line 26: r u sure 🙄 ?

MohammedAl-Rowad (approved):

package.json
MayaGillilanContentful: This is a revert, I first updated f36 generally and then realized that it may be safer to update it only where I need the newest version. Would it be better instead to update it everywhere?

packages/date/package.json
MayaGillilanContentful: Yes, this is the newest version https://github.com/contentful/forma-36/releases?page=1

package.json
MohammedAl-Rowad: I agree it's safer 👍

PR #1147, opened 2022-05-17
chore: change shelley to tolkien

MohammedAl-Rowad:
Change team name


MohammedAl-Rowad:
already done here
https://github.com/contentful/field-editors/pull/1142

PR #1155, opened 2022-05-30
fix: re-add timezone picker, fix local timezone overriding user set timezone in date editor

MayaGillilanContentful:
Adding the new timezone picker back in now with bug fix. See original PR: https://github.com/contentful/field-editors/pull/1146

Relevant bug fix:
```js
export const TimezonepickerInput = ({
  disabled,
  onChange,
  value = defaultZoneOffset,
}: TimezonepickerProps) => {
  const defaultTimezone =
    allTimezones.find((timezone: Timezone) => timezone.ianaName === dayjs.tz.guess()) ||
    allTimezones.find(
      (timezone: Timezone) => dayjs.tz(undefined, timezone.ianaName).format('Z') === value
    );
	...
```
to =>
```js
export const TimezonepickerInput = ({
  disabled,
  onChange,
  value,
}: TimezonepickerProps) => {
  const localTimezone = dayjs.tz.guess();
  const defaultTimezone =
    value && value !== dayjs.tz(undefined, localTimezone).format('Z')
      ? allTimezones.find((timezone: Timezone) => dayjs.tz(undefined, timezone.ianaName).format('Z') === value)
      : allTimezones.find((timezone: Timezone) => timezone.ianaName === localTimezone);
	...
```


packages/date/src/TimezonePickerInput.tsx
```diff
@@ -1,32 +1,77 @@
...
+  const localTimezone = dayjs.tz.guess();
+  const defaultTimezone =
+    value && value !== dayjs.tz(undefined, localTimezone).format('Z')
+      ? allTimezones.find(
+          (timezone: Timezone) => dayjs.tz(undefined, timezone.ianaName).format('Z') === value
+        )
+      : allTimezones.find((timezone: Timezone) => timezone.ianaName === localTimezone);
+
+  const [filteredTimezones, setFilteredTimezones] = useState(allTimezones);
+  const [userInput, setUserInput] = useState(defaultTimezone as Timezone);
```
MohammedAl-Rowad, line 36:
```ts
const defaultTimezone =
```

will run on each re-render, but we only need it for the state so I would recommend doing this ( I added some [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to make the code shorter)

```suggestion
    const [filteredTimezones, setFilteredTimezones] = useState(allTimezones);
  const [userInput, setUserInput] = useState<Timezone>(() => {
    const localTimezone = dayjs.tz.guess();

    const timezone =
      value === dayjs.tz(undefined, localTimezone).format('Z')
        ? allTimezones.find(({ ianaName }: Timezone) => ianaName === localTimezone)
        : allTimezones.find(
            ({ ianaName }: Timezone) => dayjs.tz(undefined, ianaName).format('Z') === value
          );

    return timezone as Timezone;
  });
```
----

packages/date/src/TimezonePickerInput.tsx
```diff
@@ -1,32 +1,77 @@
...
 };
+
 export const TimezonepickerInput = ({
   disabled,
   onChange,
   value = defaultZoneOffset,
 }: TimezonepickerProps) => {
+  const localTimezone = dayjs.tz.guess();
+  const defaultTimezone =
+    value && value !== dayjs.tz(undefined, localTimezone).format('Z')
```
MohammedAl-Rowad, line 29: I don't think you need `value && value !==  string` since `value` uses `defaultZoneOffset` by default, and the only way to override this is by passing `null ... <TimezonepickerInput value={null}/>` which will convert the expression to `null && null !== string` which means that we don't need the first part `value &&`

packages/date/src/TimezonePickerInput.tsx
```diff
@@ -1,32 +1,77 @@
...
+  const [userInput, setUserInput] = useState(defaultTimezone as Timezone);
+
+  useEffect(() => {
+    if (!value) {
+      setUserInput({
+        displayValue: '',
+        ianaName: '',
+      } as Timezone);
+    }
+  }, [value]);
```
MohammedAl-Rowad, line 45:
```suggestion
  useEffect(() => {
    if (value) {
      return;
    }
    setUserInput({ displayValue: '', ianaName: '' });
  }, [value]);
```

packages/date/src/TimezonePickerInput.tsx
```diff
@@ -1,32 +1,77 @@
...
+    },
+    [onChange]
+  );
+
+  const handleChange = (val: string) => {
+    const newFilteredTimezones = allTimezones.filter((timezone: Timezone) =>
+      timezone.displayValue.toLowerCase().includes(val.toLowerCase())
+    );
+    setFilteredTimezones(newFilteredTimezones.length > 0 ? newFilteredTimezones : allTimezones);
+  };
```
MohammedAl-Rowad, line 60:
```suggestion
  const handleChange = useCallback((val: string) => {
    const newFilteredTimezones = allTimezones.filter(({ displayValue }: Timezone) =>
      displayValue.toLowerCase().includes(val.toLowerCase())
    );
    setFilteredTimezones(newFilteredTimezones.length > 0 ? newFilteredTimezones : allTimezones);
  }, []);
```

PR #1161, opened 2022-06-10
perf(rich-text): optimize reference cards rendering

z0al:
Brings back #1135 & optimizes the rendering along the way

It looks like there is an issue with mouse selection when interacting with entry/asset cards:
* clicking on a card doesn't allow copying/deleting/or pressing enter

Everything works when placing the cursor with the keyboard once. I'd like to merge this one and work on the mouse selection issue separately 

P.S. Netlify is having some f36 issues that shouldn't be relevant


packages/rich-text/src/plugins/shared/FetchingWrappedEntryCard.tsx
```diff
@@ -1,119 +1,34 @@
...
-            props.onRemove && props.onRemove();
-          }}>
-          Remove
-        </MenuItem>
-      ) : null,
-    ].filter((item) => item);
-  }
+const InternalEntryCard = React.memo((props: InternalEntryCard) => {
+  const { entry, sdk, loadEntityScheduledActions } = props;
+  console.log('re-rendering');
```
giotiskl, line 28:
```suggestion
```

YvesRijckaert (approved):
LGTM

PR #1166, opened 2022-06-16
fix: hide embed button when all actions are disabled

MohammedAl-Rowad:
Hiding the `+Embed` dropdown button when there are no actions under it,

> This is already implemented on the master branch  

https://github.com/contentful/field-editors/blob/58dedc65a2de20973775b48cbb95b54d81894d0a/packages/rich-text/src/Toolbar/index.js#L130-L158


z0al (approved):

z0al:
I will release the changes in `rebase-next` shortly

PR #1167, opened 2022-06-20
chore: require descriptions when ignoring eslint, upgrade exhaustive-deps rule to error

MayaGillilanContentful:


packages/.eslintrc.js
```diff
@@ -7,7 +7,21 @@ module.exports = {
     require.resolve('@contentful/eslint-config-extension/react'),
   ],
   rules: {
-    'react-hooks/exhaustive-deps': 'warn',
+    'react-hooks/exhaustive-deps': 'error',
+    'eslint-comments/require-description': [
```
z0al, line 11: Ironically, I think we need a short description for the require-description rule haha. It's unclear what exactly are we requiring when looking at the `ignore` list

z0al (approved):

PR #1169, opened 2022-06-23
feat: rich text commands [TOL-109]

YvesRijckaert:
Add rich text commands plugin

How does it work?
- When the user presses the slash key, the editor will show a command palette
- When the user presses a key, the command palette will show the command suggestions
- When the user presses enter, the command palette will execute the command
- When the user presses escape, the command palette will hide
- When the user presses a letter, number, or space, the command palette will show the command suggestions
- When the user presses backspace, the command palette will remove the last character from the command string

Tests and tracking will be added later, they are part of a different story.


packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const [selectedItem, setSelectedItem] = React.useState<string>(() => {
+    // select the first item on initial render
+    if ('group' in commandItems[0]) {
+      return commandItems[0].commands[0].id;
+    }
+    return commandItems[0].id;
+  });
+
+  React.useEffect(() => {
+    if (!container || !container.current) {
```
giotiskl, line 27: NIT: optional chaining syntax makes for better readability imo
```suggestion
    if (!container?.current) {
```

packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
```diff
@@ -0,0 +1,200 @@
...
+      return;
+    }
+    const buttons = Array.from(container.current.querySelectorAll('button')) as HTMLButtonElement[];
+    const currBtn = buttons.find((btn) => btn.id === selectedItem);
+    const currIndex = currBtn ? buttons.indexOf(currBtn) : 0;
+
+    if (!currBtn && buttons.length) {
```
giotiskl, line 34: NIT: I'd suggest extracting this into a bool variable and giving it a self-explanatory name to make it easy to understand what the condition represents at a glance: `!currBtn && buttons.length`

packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
```diff
@@ -0,0 +1,200 @@
...
+          <Popover.Trigger>
+            <span />
+          </Popover.Trigger>
+          <Popover.Content className={styles.menuContent}>
+            <header className={styles.menuHeader}>
+              <SectionHeading marginBottom="none">Richtext commands</SectionHeading>
+            </header>
+            <div className={styles.menuList}>
+              {commandItems.map((item) => {
+                return 'group' in item ? (
```
giotiskl, line 125: the nested ternary makes it hard to follow the rendering logic, can we use if-else here and give some semantic names to the conditions?

packages/rich-text/src/plugins/CommandPalette/useCommands.ts
```diff
@@ -0,0 +1,186 @@
...
+            });
+          },
+        },
+      ],
+    };
+    return [...contentTypeCommands, assetCommand];
+  });
+
+  return query
+    ? commands.reduce((list, nextItem) => {
```
giotiskl, line 170: can we leave an explanatory comment here of what this commands.reduce is doing? the logic is a bit hard to follow

packages/rich-text/src/plugins/CommandPalette/utils/fetchAssets.ts
```diff
@@ -0,0 +1,26 @@
...
+      localeCode: sdk.field.locale,
+      defaultLocaleCode: sdk.locales.default,
+      defaultTitle: 'Untitled',
+    });
+
+    return {
+      contentTypeName: 'Asset',
+      displayTitle,
+      id: asset.sys.id,
+      entry: asset,
```
giotiskl, line 19: I'd propose renaming this prop name either to `asset` or `entity` if the type is reused for entries and assets.

Naming it entry and assigning an asset as the value is a bit confusing.

YvesRijckaert:
> Are we planning to add any tests for all of these cases? Will this be done separately or in this ticket?

Brought that up yesterday during PH, I've created a ticket for the next sprint to add tests and tracking

packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
```diff
@@ -0,0 +1,163 @@
...
+}
+
+const CommandListItems = ({
+  commandItems,
+  selectedItem,
+}: {
+  commandItems: CommandItems;
+  selectedItem: string;
+}) => {
+  const group = (command) => {
```
giotiskl, line 24: `group`, `asset` and `item` are all React components essentially, declaring them inside the body of `CommandListItems` is inefficient as the references are being recreated in every `render`.

What's more, they should use Pascal case. I'd be happy to keep them in the same file, but I'd propose moving them outside of `CommandListItems` body.

packages/rich-text/src/plugins/CommandPalette/components/CommandPrompt.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { BLOCKS } from '@contentful/rich-text-types';
+import { PlateRenderLeafProps } from '@udecode/plate-core';
+import { RichTextEditor } from 'types';
+
+import { isNodeTypeSelected } from '../../../helpers/editor';
+import { trimLeadingSlash } from '../utils/trimLeadingSlash';
+import { CommandList } from './CommandList';
+
+export const CommandPrompt = (props: PlateRenderLeafProps) => {
+  const query = trimLeadingSlash(props.text.text);
```
giotiskl, line 12: We need to use `React.useMemo` on `query`, otherwise the referential inequality is going to make `CommandList` re-render on every render of `CommandPrompt`, when it might not need to.

giotiskl (changes requested):
Thanks for changes until now, looks better, couple of last comments.

packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
YvesRijckaert: 🤦‍♂️ I don't know where my head was yesterday
Thanks for spotting it!

PR #1172, opened 2022-07-04
test: add tests for rich-text-commands

MayaGillilanContentful:


z0al (approved):
Great test coverage. I'm missing one for checking that the internal mark is not preserved in the output though

mshaaban0 (approved):
Looks Good

PR #1174, opened 2022-07-05
feat: add tracking for rich-text-commands

MayaGillilanContentful:


packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
```diff
@@ -96,6 +96,15 @@ export const CommandList = ({ query, editor }: CommandListProps) => {
   const commandItems = useCommands(sdk, query, editor);
   const { selectedItem } = useCommandList(commandItems, container);
 
+  React.useEffect(() => {
+    editor.tracking.onCommandPaletteAction('openRichTextCommandPalette');
+
+    return () => {
+      editor.tracking.onCommandPaletteAction('closeRichTextCommandPalette');
```
z0al, line 103: This is `cancelRichTextCommandsPalette` not close. I think it should be called when pressing `esc` instead?

z0al (commented):
It would be nice to have some tests for this

packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
MayaGillilanContentful: Shouldn't `close` be called on any dismount, since that will cover both the palette closing due to the user not wanting it (cancel) and also after embedding? Or do we want to add a `cancel` event?

MayaGillilanContentful:
> It would be nice to have some tests for this

Should the tests just be jest tests where we expect the tracking functions to be called (and with a certain payload in the case of insert events)? I don't see any tests for the other plugins using this tracking so I just want to confirm

MayaGillilanContentful:
REMINDER (for myself): add tracking to clickoutside handler after merged, or vice versa

PR #1184, opened 2022-07-12
fix: return 'failed' on getResource failures [DANTE-584]

Bikappa:


grassator:
Would be great to have a description why and a test?

Bikappa:
Never mind, I don't necessarily need this on the caller side

PR #1185, opened 2022-07-13
fix: fix commands not working properly inside headings

MayaGillilanContentful:


z0al (commented):
* Is heading the only node type affected by this?
* A test bitte :)

z0al:
> Sorry, Voids is also using the same logic but I think we don't need to add any handling for this there, right?

Voids are different. Pressing enter on a void should add a new paragraph underneath. I can't see how they would interfere with RT commands

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -430,6 +430,28 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
           ).to.have.length(1);
         });
       });
+
+      //todo test should not delete adjacent non-command text
```
z0al, line 434:
```suggestion
```

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -430,6 +430,28 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
...
+        richText.editor.click().type('test/{downarrow}{enter}{enter}');
+        richText.expectSnapshotValue();
+      });
+
+      it('should work inside headings', () => {
+        richText.editor.click().type('Heading 1');
+        richText.toolbar.toggleHeading(BLOCKS.HEADING_1);
+        richText.editor.click().type('/{enter}{enter}');
+
+        //this is used instead of snapshot value because we have randomized entry IDs
```
z0al, line 446: This is getting more and more annoying. I will see if we can make these random ids more predictable (e.g. counter)

z0al (approved):

PR #1189, opened 2022-07-18
fix(rich-text): avoid unnecessary render of reference cards

z0al:
Adds `useStableCallback` hook to avoid the performance issues some users are experiencing caused by unnecessary re-render. Work is planned to address the issue in the reference package but until that lands, this can be considered a workaround.


anho (changes requested):
Sorry to block, but maybe its easier to integrate #1190

MayaGillilanContentful (approved):
LGTM

anho (approved):

PR #1192, opened 2022-07-20
fix: no embedded entries check rich text commands [TOL-273]

YvesRijckaert:
Rich Text Editor crashes anytime a customer types "/", when the rich text field does not allow embedded entries in validation

### DIFF
--- a/packages/rich-text/src/plugins/CommandPalette/hooks/useCommandList.tsx
+++ b/packages/rich-text/src/plugins/CommandPalette/hooks/useCommandList.tsx
@@ -5,6 +5,9 @@ import isHotkey from 'is-hotkey';
 export const useCommandList = (commandItems, container) => {
   const [selectedItem, setSelectedItem] = React.useState<string>(() => {
     // select the first item on initial render
+    if (!commandItems.length) {
+      return '';
+    }
     if ('group' in commandItems[0]) {
       return commandItems[0].commands[0].id;
     }


MayaGillilanContentful (approved):
LGTM, does this add any TS errors though? If so we can just move the check for `commandItems.length` in `CommandList` up above the `useCommandList` call

YvesRijckaert:
> LGTM, does this add any TS errors though? If so we can just move the check for `commandItems.length` in `CommandList` up above the `useCommandList` call

Unfortunately we can not do that.. hooks can not be called conditionally and must be called in the exact same order in every component render.
I don't see any typescript errors though

z0al:
> when the rich text field does not allow embedded entries in validation

@YvesRijckaert Do you mean neither inline nor blocks? Wouldn't it make more sense then to not add the mark in the first place on `/` ?

YvesRijckaert:
> > when the rich text field does not allow embedded entries in validation
> 
> @YvesRijckaert Do you mean neither inline nor blocks? Wouldn't it make more sense then to not add the mark in the first place on `/` ?

Yes that's what I mean. Your comment makes sense, but we do not have access to `commandItems` in the `onKeyDown` handler, so how would you check if there are any available blocks?

MayaGillilanContentful:
> > when the rich text field does not allow embedded entries in validation
> 
> @YvesRijckaert Do you mean neither inline nor blocks? Wouldn't it make more sense then to not add the mark in the first place on `/` ?

I think the listening logic would have to be a little spaghetti-ish if we were to prevent the mark in the first place, or else we'd have to run useCommands() twice which could have a performance impact (how much of an impact does `getCachedContentTypes()` have?)

PR #1193, opened 2022-07-20
feat: replace react-codemirror2 with @uiw/react-codemirror to allow codemirror 6

MayaGillilanContentful:
React-codemirror 2 has been deprecated for ~2 years and is no longer receiving updates, and seems to have been the cause of https://contentful.atlassian.net/browse/TOL-253. In addition because Codemirror 5 is not compatible with React versions >16, it was causing dependency compatibility issues with the default React settings with `create-contentful-app`, forcing users wishing to use the JSON Editor in their apps to either downgrade manually to an earlier version of React or to use the `--legacy-peer-deps` flag with NPM which is unreliable.

The props are obviously different for Codemirror 6, so I did my best to replicate the look and functionality of the old JSON editor in the new version.


bgutsol (approved):
Codewise looks good and works better than prev version 🏅
Only the tests should be fixed since the output layout has changed, there is no `textarea` anymore

z0al (approved):
Looks great. I wonder if this should cut a major release though? what do you think @contentful/team-tolkien ?

MohammedAl-Rowad:
> Looks great. I wonder if this should cut a major release though? what do you think contentful/team-tolkien ?

I think yes, to be on the safe side since we upgraded to codemirror `v6`  and now we are using `@uiw/react-codemirror`

but it depends on the change logs of codemirror, maybe we should check that?

MayaGillilanContentful:
> > Looks great. I wonder if this should cut a major release though? what do you think contentful/team-tolkien ?
> 
> I think yes, to be on the safe side since we upgraded to codemirror `v6` and now we are using `@uiw/react-codemirror`
> 
> but it depends on the change logs of codemirror, maybe we should check that?

The migration guide https://codemirror.net/docs/migration/ is quite.. extensive, and apparently the entire architecture is significantly different. I lean towards doing a major release personally

PR #1195, opened 2022-07-21
style: fix missing entry card styles [TOL-188]

Spring3:
Before:
Screenshot 2022-07-21 at 15 06 18

After:
Screenshot 2022-07-21 at 15 06 24

### DIFF
--- a/packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
+++ b/packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
@@ -17,7 +17,7 @@ export function MissingEntityCard(props: {
       <Flex alignItems="center" justifyContent="space-between">
         <div className={props.asSquare ? styles.squareCard : ''}>
           <SectionHeading marginBottom="none">
-            {`{props.entityType} is missing or inaccessible`}
+            {props.entityType} is missing or inaccessible
           </SectionHeading>
         </div>
         {!props.isDisabled && props.onRemove && (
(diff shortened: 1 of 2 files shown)


packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
```diff
@@ -13,23 +14,24 @@ export function MissingEntityCard(props: {
...
-            props.onRemove && props.onRemove();
-          }}
-        />
-      )}
+      <Flex alignItems="center" justifyContent="space-between">
+        <div className={props.asSquare ? styles.squareCard : ''}>
+          <SectionHeading marginBottom="none">
+            {props.entityType === 'Entry' && 'Entry is missing or inaccessible'}
+            {props.entityType === 'Asset' && 'Asset is missing or inaccessible'}
+          </SectionHeading>
```
luizfonseca, line 22: I wonder if it can be like

```suggestion
          <SectionHeading marginBottom="none">
            {`{props.entityType} is missing or inaccessible`}
          </SectionHeading>
```

packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
```diff
@@ -17,8 +17,7 @@ export function MissingEntityCard(props: {
       <Flex alignItems="center" justifyContent="space-between">
         <div className={props.asSquare ? styles.squareCard : ''}>
           <SectionHeading marginBottom="none">
-            {props.entityType === 'Entry' && 'Entry is missing or inaccessible'}
-            {props.entityType === 'Asset' && 'Asset is missing or inaccessible'}
+            {`{props.entityType} is missing or inaccessible`}
```
luizfonseca, line 20: ops, I think you accepted the suggestion before i added the correct string interpolation `$`, sorry :'(

Spring3: no worries, I also noticed. Will push a quick commit after I confirm that there is no issues

Spring3: now it should be fine

MohammedAl-Rowad (approved):

PR #1196, opened 2022-07-21
test: use cypress component tests

anho:
This is just about getting the ball rolling. There will be a few follow-up PRs to this one.

Worth mentioning:
* Switching to Cypress v10 is currently out of reach as `cypress-plugin-snapshots` is not fit for it (looks unmaintained as of now).
* I had to enhance the eslint config a bit as it was barely making use of the `@contentful/eslint-config-extension` packages (which is also deprecated). This also led to a few unrelated minor changes.
* I moved the `MultiRef*` integration tests as-is without any additional being put into making this a more scalable test setup.


cypress/plugins/index.js
```diff
@@ -1,39 +1,18 @@
 const webpack = require('@cypress/webpack-preprocessor');
 const { initPlugin: initSnapshotPlugin } = require('cypress-plugin-snapshots/plugin');
 
+const webpackFilename = __dirname + '/webpack.config.js';
```
z0al, line 4: It's not clear to me why this was needed in the first place but you probably want to use `path.join(__dirname, 'webpack.config.js')` instead to make Windows users happy.

z0al (approved):

PR #1201, opened 2022-07-27
fix: remove custom theme from jsonEditor

MayaGillilanContentful:


packages/json/src/JsonEditorField.tsx
```diff
@@ -2,7 +2,6 @@ import React from 'react';
 
 import { json } from '@codemirror/lang-json';
 import tokens from '@contentful/f36-tokens';
-import { createTheme } from '@uiw/codemirror-themes';
```
z0al, line 5: Let's remove the `@uiw/codemirror-themes` package as well

MayaGillilanContentful: Good catch, thanks

z0al (approved):

PR #1202, opened 2022-07-28
test: extend multipleEntryReference component tests

MayaGillilanContentful:
I also extended the fake API creators some so that we can test localization and invalid requests easier in the future

### DIFF
--- a/cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
+++ b/cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
@@ -153,6 +153,7 @@ describe('Multiple Reference Editor', () => {
     cy.findAllByTestId('cf-ui-entry-card').eq(2).findByText('published').should('be.visible');
   });
 
+  //Do we also want to test that the props are correct, or should that be done in a jest test?
   it(`provides the custom card render method with necessary properties`, () => {
     const sdk = createReferenceEditorTestSdk({
       initialValue: [asLink(fixtures.entry.published)],
@@ -160,9 +161,7 @@ describe('Multiple Reference Editor', () => {
     mount(
       <MultipleEntryReferenceEditor
         {...commonProps}
-        renderCustomCard={(props) => {
-          cy.wrap(props).as('customCardProps');
-
+        renderCustomCard={() => {
           return <Note testId="custom-card" title="Custom card" />;
         }}
         sdk={sdk}
@@ -170,10 +169,6 @@ describe('Multiple Reference Editor', () => {
     );
 
     cy.findByTestId('custom-card').should('be.visible');
-    cy.get('@customCardProps').should('deep.include', {
-      index: 0,
-      size: 'small',
-    });
   });
 
   it('shows localized displayField as title using the current locale', () => {
--- a/cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
+++ b/cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
@@ -348,13 +348,85 @@ describe('Multiple Reference Editor', () => {
     findDefaultCards().eq(0).findByTestId('cf-ui-drag-handle').should('not.exist');
   });
 
+  it('can hide edit action', () => {
+    const sdk = createReferenceEditorTestSdk({
+      initialValue: [asLink(fixtures.entry.published), asLink(fixtures.entry.changed)],
+    });
+
+    mount(
+      <MultipleEntryReferenceEditor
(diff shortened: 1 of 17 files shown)


cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
```diff
@@ -130,4 +136,292 @@ describe('Multiple Reference Editor', () => {
...
+  it('shows unlocalized displayField as title using the default locale', () => {
+    const sdk = createReferenceEditorTestSdk({
+      initialValue: [asLink(fixtures.entry.published)],
+    });
+    mount(<MultipleEntryReferenceEditor {...commonProps} sdk={sdk} />);
+
+    findDefaultCards().eq(0).findByTestId('title').should('have.text', 'The best article ever');
+  });
+
+  it('shows status of entries', () => {
```
aodhagan-cf, line 196: This looks like a duplicate of the first test

MayaGillilanContentful: It absolutely is, thanks

cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
```diff
@@ -130,4 +136,276 @@ describe('Multiple Reference Editor', () => {
...
+    });
+    mount(<MultipleEntryReferenceEditor {...commonProps} sdk={sdk} />);
+
+    cy.findAllByTestId('cf-ui-entry-card').should('have.length', 3);
+    cy.findAllByTestId('cf-ui-entry-card').eq(0).findByText('draft').should('be.visible');
+    cy.findAllByTestId('cf-ui-entry-card').eq(1).findByText('changed').should('be.visible');
+    cy.findAllByTestId('cf-ui-entry-card').eq(2).findByText('published').should('be.visible');
+  });
+
+  //Do we also want to test that the props are correct, or should that be done in a jest test?
```
z0al, line 156: Any reason not to do it here?

cypress/component/test-sdks.ts
```diff
@@ -1,7 +1,32 @@
...
+    invalid: invalidEntry,
+  },
+  asset: {
+    changed: changedAsset,
+    empty: emptyAsset,
+    published: publishedAsset,
+    invalid: invalidAsset,
+  },
+};
+export { fixtures };
```
z0al, line 27:
```suggestion
export const fixtures = {
  entry: {
    changed: changedEntry,
    empty: emptyEntry,
    published: publishedEntry,
    invalid: invalidEntry,
  },
  asset: {
    changed: changedAsset,
    empty: emptyAsset,
    published: publishedAsset,
    invalid: invalidAsset,
  },
};
```

packages/reference/src/__fixtures__/FakeSdk.ts
```diff
@@ -25,14 +25,16 @@ const newLink = (linkType: string, id: string): Link => ({
 
 // used for component testing
 export type ReferenceEditorSdkProps = {
+  initialValue?: any;
   validations?: any;
+  shouldDelay?: boolean;
```
z0al, line 30: `shouldDelay` isn't descriptive enough of precisely what will be delayed. e.g. why only delay asset & entry calls while we also have space, dialogs ..etc. It also doesn't allow modifying the delay value.

packages/reference/src/__fixtures__/FakeSdk.ts
```diff
@@ -72,6 +77,9 @@ export function newReferenceEditorFakeSdk(
       },
       Asset: {
         get: async ({ assetId }) => {
+          if (props?.shouldDelay) {
+            await new Promise((resolve) => setTimeout(resolve, 2000));
```
z0al, line 81: We can add a tiny `delay(ms)` to avoid duplication here

z0al (commented):
Looks good overall. Have some questions/comments

cypress/component/reference/MultipleEntryReferenceEditor.spec.tsx
MayaGillilanContentful: It doesn't seem possible to wrap the props inside the `mount()` function, since Cypress throws an error when run through the command line/CI. So we either need to come up with a kind of hacky solution here or do it in a jest test

z0al: Let's leave a todo. We can check again later

z0al (approved):

PR #1203, opened 2022-07-28
refactor: Number field-editor [BAU-722]

bgutsol:
- Use type="text" instead of type="number", [see why](https://github.com/contentful/field-editors/blob/ea23844c6d4ef2cbe0f9616a36b219a00b812e09/packages/number/src/NumberEditor.tsx#L120-L122)

- Number field editor does not allow entering nun-numeric input, like a native input type="number"
- Number field editor has step buttons, like a native input type="number"
- Arrow keys can be used to step as well, like a native input type="number"
- Don't use `inputmode=decimal/numeric` because some keyboards on mobile devices don't allow typing minus `-` or dot `.`. For example, Safari on IOS.

## Related PRs:
- https://github.com/contentful/field-editors/pull/1020
- https://github.com/contentful/field-editors/pull/1112


packages/number/package.json
```diff
@@ -24,9 +24,7 @@
     "@contentful/f36-components": "^4.0.27",
     "@contentful/f36-tokens": "^4.0.0",
     "@contentful/field-editor-shared": "^1.1.3",
-    "emotion": "^10.0.17",
-    "lodash": "^4.17.15",
-    "lodash-es": "^4.17.15"
```
massao, line 29: 🥇

massao (approved):
LGTM, should we consider adding this to forma??

bgutsol:
> LGTM, should we consider adding this to forma??

Yes 💯

mshaaban0 (approved):
LGTM. Not sure about the version release, is this a breaking change?

packages/number/src/parseNumber.spec.ts
```diff
@@ -1,16 +1,56 @@
...
+    expect(parseNumber('1', 'Decimal')).toBe(1);
+    expect(parseNumber('-99', 'Decimal')).toEqual(-99);
+    expect(parseNumber('2.89', 'Decimal')).toBe(2.89);
+    expect(parseNumber('-12.89', 'Decimal')).toEqual(-12.89);
+    expect(parseNumber('foo', 'Decimal')).toBeUndefined();
+  });
+});
+
+describe('isNumberInputValueValid', () => {
+  it('should correctrly validate input when type is Integer', () => {
```
giotiskl, line 21:
```suggestion
  it('should correctly validate input when type is Integer', () => {
```

packages/number/src/NumberEditor.tsx
```diff
@@ -20,73 +25,144 @@ export interface NumberEditorProps {
...
-    if (stringifiedSdkValue !== previousValue.current && stringifiedSdkValue !== inputValue) {
-      setInputValue(stringifiedSdkValue);
+  const updateExternalValue = (value: number | undefined) => {
+    if (sdkValue !== value) {
+      setValue(value);
+    }
+  };
+
+  const changeValueByStep = (type: 'increment' | 'decrement') => {
+    const currentValue = isNaN(+inputValue) ? 0 : +inputValue;
```
giotiskl, line 64:
```suggestion
    const currentValue = Number.isNaN(+inputValue) ? 0 : +inputValue;
```
we should use `Number.isNaN` over `isNaN`, explanation here: https://stackoverflow.com/questions/33164725/confusion-between-isnan-and-number-isnan-in-javascript

packages/number/src/NumberEditor.tsx
```diff
@@ -20,73 +25,144 @@ export interface NumberEditorProps {
...
+  const inputRef = React.useRef<HTMLInputElement>(null);
 
   React.useEffect(() => {
-    previousValue.current = valueToString(sdkValue);
+    const stringSdkValue = valueToString(sdkValue);
+    // Update the input value if the SDK value (numeric) changes
+    if (stringSdkValue !== inputValue) {
+      setInputValue(stringSdkValue);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- we want to trigger it only when sdkValue had been changed
```
giotiskl, line 54:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps -- we want to trigger it only when sdkValue has changed
```

packages/number/src/NumberEditor.tsx
```diff
@@ -20,73 +25,144 @@ export interface NumberEditorProps {
...
-    // Update the input value (string) if the SDK value (numeric) changes
-    if (stringifiedSdkValue !== previousValue.current && stringifiedSdkValue !== inputValue) {
-      setInputValue(stringifiedSdkValue);
+  const updateExternalValue = (value: number | undefined) => {
+    if (sdkValue !== value) {
+      setValue(value);
+    }
+  };
+
+  const changeValueByStep = (type: 'increment' | 'decrement') => {
```
giotiskl, line 63: instead of hardcoding strings in a union value I'd propose introducing an enum to avoid magic string usage:
```
enum StepChangeType {
  Increment = 'increment',
  Decrement = 'decrement',
}
```

```suggestion
  const changeValueByStep = (type: StepChangeType) => {
```

packages/number/src/NumberEditor.tsx
```diff
@@ -20,73 +25,144 @@ export interface NumberEditorProps {
...
+  const handleControlPointerDown: React.PointerEventHandler<HTMLButtonElement> = (event) => {
+    event.preventDefault();
+    inputRef.current?.focus();
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<any>) => {
+    const keyToFnMap: {
+      [key: string]: () => void;
+    } = {
+      ArrowUp: () => changeValueByStep('increment'),
```
giotiskl, line 85: this would then become:
```suggestion
      ArrowUp: () => changeValueByStep(StepChangeType.Increment),
```
if we introduce the enum, similarly for the item below

packages/number/src/parseNumber.ts
```diff
@@ -1,51 +1,16 @@
...
-    };
-  }
-
-  if (isNaN(floatVal)) {
-    return {
-      isValid: false,
-      value: undefined,
-    };
+export function parseNumber(value: string, type: string) {
+  if (!value || isNaN(+value)) {
```
giotiskl, line 2:
```suggestion
  if (!value || Number.isNaN(+value)) {
```

giotiskl (changes requested):
Looks good overall, my main suggestion would be to rely on `Number.isNaN` instead of `isNaN`, explanation in the comments.

bgutsol:
> Looks good overall, my main suggestion would be to rely on `Number.isNaN` instead of `isNaN`, explanation in the comments.

@giotiskl thanks for reviewing! I've addressed your comments, please take a look

giotiskl (approved):

packages/number/src/NumberEditor.tsx
```diff
@@ -20,73 +24,150 @@ export interface NumberEditorProps {
...
+        // See more details: https://github.com/facebook/react/issues/6556
+        type="text"
         testId="number-editor-input"
-        min={range.min !== undefined ? String(range.min) : ''}
-        max={range.max !== undefined ? String(range.max) : ''}
-        step={field.type === 'Integer' ? '1' : ''}
+        className={styles.input}
+        min={range.min}
+        max={range.max}
+        step={NUMBER_STEP}
```
denkristoffer, line 133: Does this work with text input?

denkristoffer (commented):
What does it look like?

bgutsol:
> What does it look like?

@denkristoffer you can take a look on preview https://deploy-preview-1203--contentful-field-editors.netlify.app/number

packages/number/src/NumberEditor.tsx
bgutsol: Yes, it's actually the same role that input type="number" has: https://www.digitala11y.com/spinbutton-role/

denkristoffer: I meant the `step` attribute. It doesn't seem to be valid: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step

denkristoffer:
Looks good, how about some hover and active styling on the buttons though?

bgutsol:
> Looks good, how about some hover and active styling on the buttons though?

Active styling is in place. With hover I agree, we can add some

packages/number/src/NumberEditor.tsx
bgutsol: step can be removed

denkristoffer:
> Active styling is in place. With hover I agree, we can add some

I mean the clicked state, I'm not seeing it:

https://user-images.githubusercontent.com/219017/185390585-683f38c2-d25d-479c-bb47-641fa9960917.mov

bgutsol:
> > Active styling is in place. With hover I agree, we can add some
> 
> I mean the clicked state, I'm not seeing it:
> 
>  Screen.Recording.2022-08-18.at.14.04.46.mov

The current active style doesn't work in firefox for some reason 🤔

bgutsol:
> > > Active styling is in place. With hover I agree, we can add some
> > 
> > 
> > I mean the clicked state, I'm not seeing it:
> > Screen.Recording.2022-08-18.at.14.04.46.mov
> 
> The current active style doesn't work in firefox for some reason 🤔

Unfortunately, there is no easy fix for it, it's just the way [Firefox implemented its core logic](https://bugzilla.mozilla.org/show_bug.cgi?id=771241), and it's different from the other browsers. 
So basically `event.preventDefault();`, that I call within `onPointerDown` callback, cancels the active state for our buttons (only in Firefox). But I can't remove the `event.preventDefault();` because I need it to switch the focus from the buttons back to the input for a11y purposes. So the only way to fix it: is by setting our custom `.active` class. But it doesn't worth the effort. What's your thoughts @denkristoffer?

PR #1204, opened 2022-07-29
chore: update `contentful-management` to the latest version

grassator:
Version is set to `^10.0.0` instead of a later one to provide maximum compatibility when `field-editors` is used with other libraries using `contentful-management`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9606,10 +9606,10 @@ contentful-import@^7.9.23:
     moment "^2.22.2"
     yargs "^16.0.3"
 
-contentful-management@^10.8.0:
-  version "10.8.0"
-  resolved "https://registry.yarnpkg.com/contentful-management/-/contentful-management-10.8.0.tgz#6f60264e180a1e97b1dd9dab849a553d069f731a"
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -9606,18 +9606,18 @@ contentful-import@^7.9.23:
...
+  integrity sha512-IwKwMhxvu3fO9Ei1FtaQtvMqjVZc/xb9gFl6Pi155rUQBGUWNn6HL93b2apKCxmO9q1neQ6qgZccFrMtviKnXA==
   dependencies:
     "@types/json-patch" "0.0.30"
     axios "^0.27.1"
     contentful-sdk-core "^7.0.1"
     fast-copy "^2.1.1"
     lodash.isplainobject "^4.0.6"
 
-contentful-management@^7.14.0, contentful-management@^7.22.0, contentful-management@^7.3.1, contentful-management@^7.42.1:
+contentful-management@^7.14.0, contentful-management@^7.3.1:
```
z0al, line 9620: Umm.. do we expect to have v7 to be still referenced in the lock file?

z0al (approved):

yarn.lock
grassator: Sadly yes, indirect dependencies still depend on older versions

PR #1207, opened 2022-08-08
fix: introduce callback to properly handle available content types [ZEND-2500]

FLoppix:
https://contentful.atlassian.net/browse/ZEND-2500

### DIFF
--- a/packages/reference/src/components/LinkActions/LinkEntityActions.tsx
+++ b/packages/reference/src/components/LinkActions/LinkEntityActions.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import { useMemo } from 'react';
+
+import { EditorPermissions } from '../../common/useEditorPermissions';
 import {
   Action,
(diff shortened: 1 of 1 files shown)


packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -90,32 +91,39 @@ export function useLinkActionsProps(props: LinkEntityActionsProps): LinkActionsP
     [sdk, entityType, onCreated]
   );
 
-  // Wrapping these two with useCallback caused a bug [ZEND-2154] where CTs were not propagated to the UI at all
```
ronaldronson, line 93: Doesn't the useCallback cause the issue any longer?

FLoppix: Nope I checked. It does not

mdmjg (approved):

PR #1214, opened 2022-08-11
feat: moving new entity store in [TOL-282]

aodhagan-cf:
Moving over the new and improve Entity store from the web app

- Fixing tsdx build issues with yarn resolutions
- Refactoring where we use the old entity store hooks in reference/rich-text packages with the new


package.json
```diff
@@ -121,7 +121,9 @@
     "@types/react": "^16.14.5",
     "babel-preset-react-app": "^9.1.2",
     "react": "16.14.0",
-    "react-dom": "16.14.0"
+    "react-docgen-typescript": "^2.2.2",
```
aodhagan-cf, line 124: 1. Importing the new EntityStore into the reference package caused a build issue with `tsdx` as it was using an old version of typescript
2. Set the typescript resolution to `"typescript": "^4.4.2"`
3. This broke the `docz` build because `docz` uses an old version of `react-docgen-typescript` which using an old version of typescript but now step 2 broke this.
4. Set the react-docgen-typescript resolution to `"react-docgen-typescript": "^2.2.2"`
5. It builds 🎉

packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
```diff
@@ -128,7 +113,7 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
       const builtinCardProps: Omit<WrappedAssetCardProps, 'isClickable'> = {
         ...commonProps,
         ...props,
-        getEntityScheduledActions: loadEntityScheduledActions,
+        getEntityScheduledActions: sdk.space.getEntityScheduledActions,
```
aodhagan-cf, line 116: Originally was using the new `useEntityLoader` hook here but it sends invalid requests, looking for environment id

Screenshot 2022-08-16 at 20 58 57

aodhagan-cf: Same in `FetchingWrappedEntryCard`

packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
```diff
@@ -28,25 +32,26 @@ interface FetchingWrappedInlineEntryCardProps {
...
-  const entry = React.useMemo(() => entries[props.entryId], [entries, props.entryId]);
+  const { data: entry, status: requestStatus } = useEntity<Entry>('Entry', props.entryId);
+  const { getEntityScheduledActions } = useEntityLoader();
+  const loadEntityScheduledActions = () => getEntityScheduledActions('Entry', props.entryId);
 
   const allContentTypes = props.sdk.space.getCachedContentTypes();
   const { onEntityFetchComplete } = props;
   const contentType = React.useMemo(() => {
-    if (!entry || entry === 'failed' || !allContentTypes) return undefined;
+    if (!entry || !allContentTypes) return undefined;
```
MayaGillilanContentful, line 42: Should we also add a condition for `requestStatus !== 'success'` here?

MayaGillilanContentful (commented):
Except for the `scheduled_actions` issue above, LGTM, but I'll let others who were more involved have the final say

packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
aodhagan-cf: Previously if the entry would equal the string `failed` if there was an error. Now if there is an error it returns undefined so I don't think that would be necessary

MayaGillilanContentful (approved):

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -1781,6 +1781,7 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
       // Necessary for reading the correct LocalStorage values as we do
       // the initial page load on the beforeEach hook
       cy.reload();
+      cy.wait(500);
```
z0al, line 1784: I think this can be avoided by making the assertion below retry correctly. similar to what we do in [richtext.expectValue](https://github.com/contentful/field-editors/blob/9d29a7aab8f073b98f29bde56c41f2615c2be51e/cypress/integration/rich-text/RichTextPage.ts#L104-L106). The difference being we will get a fresh value of `window.actions` every time cypress retry the `.should` command.

You can handle this in the follow up PR

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -57,21 +61,13 @@ async function openEntry(
...
 export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
-  const { getEntry, getAsset, loadEntityScheduledActions, entries } = useEntities();
-
-  React.useEffect(() => {
-    getEntry(props.entryId);
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: Evaluate the dependencies
-  }, [props.entryId]);
+  const { data: entry, status } = useEntity<Entry>('Entry', props.entryId);
+  const { getEntityScheduledActions } = useEntityLoader();
+  const loadEntityScheduledActions = () => getEntityScheduledActions('Entry', props.entryId);
```
z0al, line 66: I think these `loadEntityScheduledActions` need to be wrapped in a `useMemo` to avoid re-rendering in children?

z0al (approved):

cypress/integration/rich-text/RichTextEditor.spec.ts
aodhagan-cf: Will do

packages/reference/src/common/EntityStore.tsx
```diff
@@ -0,0 +1,445 @@
...
+}
+
+function EntityProvider({ children, ...props }: React.PropsWithChildren<EntityStoreProps>) {
+  const reactQueryClient = useMemo(() => {
+    const queryCache = new QueryCache();
+    const queryClient = new QueryClient({
+      queryCache,
+      defaultOptions: {
+        queries: {
+          useErrorBoundary: true,
```
anho, line 425: @z0al @aodhagan-cf can we change this to `false` please? It's on me to not pick the right default, but I'm certain we want the individual components to handle error cases and NOT escalate this to a potential non-existing error boundary.

aodhagan-cf: Yep I can do this in the follow up PR :)

PR #1216, opened 2022-08-17
feat: moving resource editors from web app [TOL-282]

aodhagan-cf:
- Moving in resource editors from the web app reference package
- Refactoring  `SingleResourceReferenceEditor` and `MultipleResourceReferenceEditor` to allow them to be consumed by the web app (and other apps)


MohammedAl-Rowad (approved):

packages/reference/src/resources/useResourceLinkActions.ts
```diff
@@ -0,0 +1,104 @@
+import { useCallback, useMemo } from 'react';
+
+import { FieldExtensionSDK } from '@contentful/app-sdk';
+import { EntryProps, KeyValueMap } from 'contentful-management';
+import noop from 'lodash/noop';
```
anho, line 5: Can we just do an empty function instead of importing lodash?

packages/reference/src/resources/SingleResourceReferenceEditor.spec.tsx
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
```
anho, line 1: This file should be a cypress component test instead.

packages/reference/src/resources/MultipleResourceReferenceEditor.spec.tsx
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
```
anho, line 1: This file should be a cypress component test instead.

anho (commented):
I'd really like to see the tests for the new editors as cypress component tests. Happy to pair on those next week if you need a hand.

packages/reference/src/resources/useResourceLinkActions.ts
aodhagan-cf: I've removed `noop` reference :)

aodhagan-cf: I've removed imports of lodash :)

aodhagan-cf:
> I'd really like to see the tests for the new editors as cypress component tests. Happy to pair on those next week if you need a hand.

I will address this in a follow up ticket

PR #1226, opened 2022-09-05
test: add cypress tests for xspace reference editors

MayaGillilanContentful:


packages/reference/src/__fixtures__/FakeSdk.ts
```diff
@@ -104,9 +108,28 @@ export function newReferenceEditorFakeSdk(
           if (assetId === changedAsset.sys.id) {
             return changedAsset;
           }
-          return Promise.reject();
```
aodhagan-cf, line 107: Do you know why this used to be a rejected promise?

MayaGillilanContentful: Looks like the normal reference editor tests need it to be a rejected Promise for some reason. I reverted this change to fix those tests.

aodhagan-cf (approved):

PR #1232, opened 2022-09-09
test(reference-editor): external changes should change cards

anho:
Cover previous fix https://github.com/contentful/field-editors/pull/1227 with a test.

The changes, to fit the need of these added tests, feel very much bolted on right now. This is mainly due to the fact that the mocks and fixtures come from the docs build with lots of hardcoded examples. Moving more jest and integration tests over to component tests can help refine those things.


packages/reference/src/__fixtures__/space/index.ts
```diff
@@ -0,0 +1,3 @@
+import indifferent from './indifferent_space.json';
+
+export { indifferent };
```
MayaGillilanContentful, line 3: Not an issue, but I wonder, why call it "indifferent"?

MayaGillilanContentful (approved):

packages/reference/src/__fixtures__/space/index.ts
anho: It's indifferent from the current space 🤷🏻‍♂️

PR #1234, opened 2022-09-12
feat: use urn from entries that are fetched from api

alicankargin:
## Purpose

Use urn from entries that are fetched from the api. EntityMetaSysProps type in [contentful-management.js](https://github.com/contentful/contentful-management.js) repository is missing urn. It's added in this [PR](https://github.com/contentful/contentful-management.js/pull/1479)

### DIFF
--- a/packages/reference/src/resources/useResourceLinkActions.ts
+++ b/packages/reference/src/resources/useResourceLinkActions.ts
@@ -1,11 +1,11 @@
 import { useCallback, useMemo } from 'react';
 
 import { FieldExtensionSDK } from '@contentful/app-sdk';
-import { EntryProps, KeyValueMap, WithResourceName } from 'contentful-management';
+import { EntryProps, WithResourceName } from 'contentful-management';
 
 import { LinkActionsProps } from '../components';
 
-const toLinkItem = (entry: WithResourceName<EntryProps<KeyValueMap>>) => ({
+const toLinkItem = (entry: WithResourceName<EntryProps>) => ({
   sys: {
     type: 'ResourceLink',
     linkType: 'Contentful:Entry',
@@ -19,10 +19,10 @@ export function useResourceLinkActions({
   onAfterLink,
 }: Pick<FieldExtensionSDK, 'field' | 'dialogs'> & {
   apiUrl: string;
-  onAfterLink?: (e: EntryProps<KeyValueMap>) => void;
+  onAfterLink?: (e: EntryProps) => void;
 }): LinkActionsProps {
   const handleAfterLink = useCallback(
-    (entries: EntryProps<KeyValueMap>[]) => {
+    (entries: EntryProps[]) => {
       if (!onAfterLink) {
         return;
       }
@@ -33,8 +33,8 @@ export function useResourceLinkActions({
   const multiple = field.type === 'Array';
 
   const onLinkedExisting = useMemo(() => {
-    return (entries: EntryProps<KeyValueMap>[]) => {
-      const entriesWithResourceName = entries as WithResourceName<EntryProps<KeyValueMap>>[];
+    return (entries: EntryProps[]) => {
+      const entriesWithResourceName = entries as WithResourceName<EntryProps>[];
       let updatedValue;
       if (multiple) {
         const linkItems = entriesWithResourceName.map(toLinkItem);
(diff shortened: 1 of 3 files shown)


alicankargin:
Waiting for [PR](https://github.com/contentful/experience-packages/pull/1730/files) to be merged before merging this one.

packages/reference/src/resources/useResourceLinkActions.ts
```diff
@@ -26,20 +25,17 @@ export function useResourceLinkActions({
   const multiple = field.type === 'Array';
 
   const toLinkItem = useMemo(() => {
-    function toUrn(entry: EntryProps<KeyValueMap>) {
-      return `crn:${apiUrl}:::content:spaces/${entry.sys.space.sys.id}/entries/${entry.sys.id}`;
-    }
-
     return (entry: EntryProps<KeyValueMap>) => {
```
anho, line 28: This can be moved outside the hook now. Let's also bump `contentful-management` and type the `entry` param properly.

anho (changes requested):

packages/reference/src/resources/useResourceLinkActions.ts
alicankargin: Alright I was waiting for the type to be ready. With this `This can be moved outside the hook now` do you mean that I should extract it from the hook and put it on top the file as a function?

packages/reference/src/resources/useResourceLinkActions.ts
```diff
@@ -1,12 +1,19 @@
 import { useCallback, useMemo } from 'react';
 
 import { FieldExtensionSDK } from '@contentful/app-sdk';
-import { EntryProps, KeyValueMap } from 'contentful-management';
+import { EntryProps, KeyValueMap, WithResourceName } from 'contentful-management';
 
 import { LinkActionsProps } from '../components';
 
+const toLinkItem = (entry: WithResourceName<EntryProps<KeyValueMap>>) => ({
```
anho, line 8:
```suggestion
const toLinkItem = (entry: WithResourceName<EntryProps>) => ({
```

yarn.lock
```diff
@@ -9691,6 +9691,17 @@ contentful-management@^10.0.0:
     fast-copy "^2.1.1"
     lodash.isplainobject "^4.0.6"
 
+contentful-management@^10.14.0:
```
anho, line 9694: Can you please run `npx yarn-deduplicate`?

packages/reference/src/resources/useResourceLinkActions.ts
```diff
@@ -25,39 +32,21 @@ export function useResourceLinkActions({
...
-        },
-      };
-    };
-  }, [apiUrl]);
-
   const onLinkedExisting = useMemo(() => {
-    if (multiple) {
-      return (entries: EntryProps<KeyValueMap>[]) => {
-        const linkItems = entries.map(toLinkItem);
+    return (entries: EntryProps<KeyValueMap>[]) => {
```
anho, line 36: I think you can kill all the `<KeyValueMap>` generics as this is the default anyway.

anho, line 36: Was there a specific itch you wanted to scratch by changing it to a single function? Doesn't feel more readable to me.

alicankargin: We did this with Luca while pair programming. I'm ok with reverting it. What do you think? @Bikappa

yarn.lock
anho: Then let's go for the CMA package only.

packages/reference/src/resources/useResourceLinkActions.ts
Bikappa: I avoids the duplication of the code that is now around the if/else block. Nothing massive oc, but it stops 2 chunks of code that do the same thing from growing

Bikappa: Perhaps we can bring back readability with (pseudocode):
```
...
const updatedValue = giveMeTheValue(multiple, prevValue, entriesWithResourceName)
...
```

anho: Don't want to bikeshed, but it's worth considering keeping 2 lines of duplicated code for the sake of readibility.

anho (approved):

packages/reference/src/resources/useResourceLinkActions.ts
alicankargin: Ok I have extracted it to a function. I think it's readable now and also there's no code duplication. We can all agree that extraction is a good practice as well :) I'll merge this tomorrow

PR #1237, opened 2022-09-14
fix: increase pqueue concurrency in entitystore

MayaGillilanContentful:
The current default concurrency limit is too low, causing increases in field editor loading times with large numbers of references.


Rowadz (approved):
:shipit:

z0al (approved):

PR #1239, opened 2022-09-15
fix: fix the way date is parsed in the DateEditor [BAU-943]

bgutsol:
The DatepickerInput should be time and timezone agnostic, thats why we use `moment().format()` instead of `moment().toDate()`. `moment().toDate()` takes into account time and timezone and converts it based on your system timezone which can result in the date change. e.g. for user who has a timezone +02:00, `moment('2022-09-16T00:00+4:00').toDate()` will return `2022-09-15`

### DIFF
--- a/packages/date/src/DatepickerInput.tsx
+++ b/packages/date/src/DatepickerInput.tsx
@@ -29,10 +29,17 @@ export const DatepickerInput = (props: DatePickerProps) => {
     return [fromDate, toDate];
   }, []);
 
+  // The DatepickerInput should be time and timezone agnostic,
+  // thats why we use moment().format() instead of moment().toDate().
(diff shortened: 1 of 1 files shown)


YvesRijckaert (approved):
👌 thank you for fixing this!

denkristoffer:
We shouldn't be saving the date as UTC then?

bgutsol:
> We shouldn't be saving the date as UTC then?

@denkristoffer you are right, we shouldn't, but I don't want to refactor the old field-date-editor-date logic just to fix this issue. It should be either completely refactored or it's better not to touch anything there.

denkristoffer:
I'm saying we should be, but this is not.

PR #1251, opened 2022-09-19
test: add tests for field-editor-date timezone logic

bgutsol:
Cover with tests following cases: 
- https://github.com/contentful/field-editors/pull/1239
- https://github.com/contentful/field-editors/pull/1255

### DIFF
--- a/packages/date/src/DatepickerInput.spec.tsx
+++ b/packages/date/src/DatepickerInput.spec.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable no-global-assign -- need for test purposes */
 import React from 'react';
 
-import { render, configure } from '@testing-library/react';
+import { render, configure, screen } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import timezonedDate from 'timezoned-date';
 
@@ -14,10 +14,10 @@ configure({
 
 let originalDate: typeof Date;
 beforeEach(() => {
-  originalDate = Date;
+  originalDate = global.Date;
 });
 afterEach(() => {
-  Date = originalDate;
+  global.Date = originalDate;
 });
 
 const renderDatepicker = (dateString: string) => {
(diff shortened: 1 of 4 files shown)


denkristoffer:
Does this also cover cases where the viewer's time zone is different from the one selected?

bgutsol:
> 

yes, that's why I added two tests with different time zones

denkristoffer:
@bgutsol Isn't it the field value being set?

bgutsol:
> @bgutsol Isn't it the field value being set?

what do you mean?

denkristoffer:
@bgutsol I mean you're setting different values for the field, but in both test cases the system time is the same. Looks to me like we should cover some cases where we use https://docs.cypress.io/api/commands/clock#Change-current-system-time

bgutsol:
> @bgutsol I mean you're setting different values for the field, but in both test cases the system time is the same. Looks to me like it would be beneficial for us to add some tests where we use https://docs.cypress.io/api/commands/clock#Change-current-system-time

@denkristoffer tbh it doesn't matter what system time is on the machine since it doesn't take system time into account when converting a date. The only system parameter that affects the conversion is the timezone. That's why I'm running two tests with different UTC offsets in the `initialValue` to cover most cases. The other way can be to read the current local system timezone and based on it prepare a UTC offset for the initial value, but it seems artificial

denkristoffer:
> @denkristoffer tbh it doesn't matter what system time is on the machine since it doesn't take system time into account when converting a date. The only system parameter that affects the conversion is the timezone. That's why I'm running two tests with different UTC offsets in the initialValue to cover most cases. The other way can be to read the current local system timezone and based on it prepare a UTC offset for the initial value, but it seems artificial to me. What are your thoughts?

@bgutsol But the problem we had before was exactly that, that the date displayed took local timezone into consideration, right? So I think it makes sense to cover that situation here with a test that makes sure that it **doesn't** change the displayed data no matter what time zone you're in.

Although I don't think we have to test it with an integration test.

bgutsol:
> > @denkristoffer tbh it doesn't matter what system time is on the machine since it doesn't take system time into account when converting a date. The only system parameter that affects the conversion is the timezone. That's why I'm running two tests with different UTC offsets in the initialValue to cover most cases. The other way can be to read the current local system timezone and based on it prepare a UTC offset for the initial value, but it seems artificial to me. What are your thoughts?
> 
> @bgutsol But the problem we had before was exactly that, that the date displayed took local timezone into consideration, right? So I think it makes sense to cover that situation here with a test that makes sure that it **doesn't** change the displayed data no matter what time zone you're in.
> 
> Although I don't think we have to test it with an integration test.

Added some unit tests, @denkristoffer pls take a look

packages/date/src/DatepickerInput.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  return render(<DatepickerInput value={date} onChange={jest.fn()} />);
+};
+
+describe('Date: DatepickerInput', function () {
+  describe('renders correct date, when the system UTC offset is +02:00 and the one in the props.value is', () => {
+    it('+04:00', async () => {
+      Date = timezonedDate.makeConstructor(120);
+      const { getByTestId } = renderDatepicker('2022-09-22T00:00+04:00');
```
denkristoffer, line 36: Just a suggestion, but it's worth reconsidering this pattern. It would fail this eslint rule: https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-screen-queries.md

packages/date/src/DatepickerInput.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+    uses12hClock: false,
+  });
+
+  return render(<DatepickerInput value={date} onChange={jest.fn()} />);
+};
+
+describe('Date: DatepickerInput', function () {
+  describe('renders correct date, when the system UTC offset is +02:00 and the one in the props.value is', () => {
+    it('+04:00', async () => {
+      Date = timezonedDate.makeConstructor(120);
```
denkristoffer, line 35: Does this work without setting `Date` in the global scope? The way I read it here we have to do that to hook into the date object properly? https://github.com/thorn0/timezoned-date

bgutsol: it works you can check it by logging the date. But I can change it to `global.Date` if it's gonna be more clear

mshaaban0 (approved):
💅🏼 💯

packages/date/src/DatepickerInput.spec.tsx
denkristoffer: Nah no worries, thanks for checking 💯

denkristoffer (approved):

PR #1256, opened 2022-09-26
chore: update f36-components from 4.0.33 to 4.20.6

MayaGillilanContentful:


YvesRijckaert (approved):
Does this fix the rich text being broken?

MayaGillilanContentful:
> Does this fix the rich text being broken?

Yes, there was some issue in the F36 tooltip that was fixed in a later version

PR #1268, opened 2022-10-31
feat: add new lifecycle hook to propagate the disabled state for a field [HOMER-1176]

FLoppix:
## Purpose 

Add a new lifecycle hook to propagate the disabled state for a field, when the underlying permission changes.

### DIFF
--- a/packages/_shared/src/FieldConnector.ts
+++ b/packages/_shared/src/FieldConnector.ts
@@ -1,7 +1,8 @@
 import React from 'react';
+
+import { FieldAPI, ValidationError } from '@contentful/app-sdk';
 import isEqual from 'lodash/isEqual';
 import throttle from 'lodash/throttle';
-import { FieldAPI, ValidationError } from '@contentful/app-sdk';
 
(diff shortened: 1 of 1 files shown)


packages/_shared/src/FieldConnector.ts
```diff
@@ -122,6 +123,20 @@ export class FieldConnector<ValueType> extends React.Component<
...
+    /**
+     * Only propagate the needed change when the value really changed
+     * so we don't end up in an infinite state loop
+     */
+    this.unsubscribeDisabled = field.onIsDisabledChanged((disabled: boolean) => {
+      if (this.state.disabled !== disabled) {
+        this.setState({ disabled });
+      }
+    });
+  }
```
Chaoste, line 138: Why not write it like this already in `componentDidMount` so we always check for `disabled` beforehand?

FLoppix: I am not sure what you exactly mean. `componentDidMount` is just called once in the lifecycle. So any following updates of the permissions are not propagated.

PR #1280, opened 2022-11-14
feat: superscipt and subscript support to rich text editor [TOL-30]

aodhagan-cf:
Adding subscript and superscript support to rich text editor

- [x] Add new RT editor support for sub/sup
- [x] Sub/sup marks can be mixed with existing marks e.g. bold, italic, code, underline. like this
- [x] However, sub/sup marks should cancel each other out. A text may only have a sub or sup mark at the same time.
- [x] No support for hotkeys (aka keyboard shortcuts)
- [x] Toolbar items for subscript/superscript should only be visible if the marks are enabled in enabledMarks
- [x] Text with sub/sup marks should be selectable, and copy/pasting it should retain the marks. Both inside the editor and from external sources.


packages/rich-text/src/RichTextEditor.mdx
```diff
@@ -61,7 +61,16 @@ import { RichTextEditor } from '@contentful/field-editor-rich-text';
     };
 
     const initialValue = JSON.parse(window.localStorage.getItem('initialValue')) || undefined
-    const [field, mitt] = useMemo(() => createFakeFieldAPI(mock => mock, initialValue), []);
+    const [fakeField, mitt] = useMemo(() => createFakeFieldAPI(mock => mock, initialValue), []);
+    // Overriding mark validation here to show subscript and superscript marks
```
aodhagan-cf, line 65: By default we don't show the superscript and subscript marks but showing in the preview so doing an override here

packages/rich-text/src/helpers/__tests__/validations.test.ts
```diff
@@ -0,0 +1,52 @@
+import { createFakeFieldAPI } from '@contentful/field-editor-test-utils';
```
aodhagan-cf, line 1: @MayaGillilanContentful testing the logic for validating marks in these unit tests

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -208,19 +208,45 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
   });
 
   describe('Marks', () => {
+    const findMarkViaToolvar = (mark: MARKS) => cy.findByTestId(`${mark}-toolbar-button`);
```
MayaGillilanContentful, line 211:
```suggestion
    const findMarkViaToolbar = (mark: MARKS) => cy.findByTestId(`${mark}-toolbar-button`);
```

MayaGillilanContentful: Should also be renamed where used

packages/rich-text/src/RichTextEditor.mdx
```diff
@@ -61,7 +61,16 @@ import { RichTextEditor } from '@contentful/field-editor-rich-text';
     };
 
     const initialValue = JSON.parse(window.localStorage.getItem('initialValue')) || undefined
-    const [field, mitt] = useMemo(() => createFakeFieldAPI(mock => mock, initialValue), []);
+    const fieldValidations = JSON.parse(window.localStorage.getItem('fieldValidations')) || undefined
```
MayaGillilanContentful, line 64: Should this default to undefined and not []?

MayaGillilanContentful (approved):

packages/rich-text/src/helpers/validations.ts
```diff
@@ -30,6 +30,10 @@ const isFormattingOptionEnabled = (field, validationType, nodeTypeOrMark) => {
   // TODO: In the future, validations will always be opt-in. In that case
   // we don't need this step.
   if (enabledFormattings === undefined) {
+    if ([MARKS.SUBSCRIPT, MARKS.SUPERSCRIPT].includes(nodeTypeOrMark)) {
+      return false;
+    }
```
z0al, line 35: What is this for?

packages/rich-text/src/plugins/Marks/Subscript.tsx
```diff
@@ -0,0 +1,40 @@
...
+    },
+    deserializeHtml: {
+      rules: [
+        {
+          validNodeName: ['SUB'],
+        },
+      ],
+      query: (el) => {
+        return !someHtmlElement(el, (node) => node.style.fontWeight === 'normal');
+      },
```
z0al, line 38: You don't need this AFAIK. It was only needed for `bold` because it could be inferred from the fontWeight.

packages/rich-text/src/plugins/Marks/Superscript.tsx
```diff
@@ -0,0 +1,40 @@
...
+    },
+    deserializeHtml: {
+      rules: [
+        {
+          validNodeName: ['SUP'],
+        },
+      ],
+      query: (el) => {
+        return !someHtmlElement(el, (node) => node.style.fontWeight === 'normal');
+      },
```
z0al, line 38: Ditto

packages/rich-text/src/plugins/Marks/helpers.ts
```diff
@@ -1,9 +1,22 @@
...
+export const toggleMarkAndDeactivateConflictingMarks = (
+  editor: PlateEditor<RichTextEditor>,
+  mark: MARKS
+) => {
+  toggleMark(editor, { key: mark });
+  if (mark === MARKS.SUBSCRIPT && isMarkActive(editor, MARKS.SUPERSCRIPT)) {
+    toggleMark(editor, { key: MARKS.SUPERSCRIPT });
+  } else if (mark === MARKS.SUPERSCRIPT && isMarkActive(editor, MARKS.SUBSCRIPT)) {
+    toggleMark(editor, { key: MARKS.SUBSCRIPT });
+  }
```
z0al, line 17: How about using the `clear` option of `toggleMark`?

z0al (commented):
Looks great. Some clarifying questions and suggestions

packages/rich-text/src/plugins/Marks/helpers.ts
MayaGillilanContentful: I'm not sure how the clear option works, but would that remove any other marks there (such as italic)?

packages/rich-text/src/RichTextEditor.mdx
aodhagan-cf: I'm falling back to undefined as I think local storage uses null by default. Then later on it will fallback to whatever value the mock gives us if it has one. I don't want to hard code this to []

packages/rich-text/src/helpers/validations.ts
aodhagan-cf: > sub/sup toolbar options are only shown when the enabledMarks validation explicitly allows them

Follow this AC

packages/rich-text/src/plugins/Marks/Subscript.tsx
aodhagan-cf: You are right this is not needed

packages/rich-text/src/plugins/Marks/helpers.ts
z0al: So we can have something like below:

```javascript
const subs = [MARKS.SUPERSCRIPT, MARKS.SUBSCRIPT]

const clear = subs.includes(mark) ? subs | []

toggleMark(editor, {key: mark, clear})
```

aodhagan-cf: The docs don't exactly make it clear how to use this it accepts a string or a list of strings. I've tried 
`toggleMark(editor, { key: MARKS.SUBSCRIPT, clear: MARKS.SUPERSCRIPT });` without any success.

Do you have any suggestions?

aodhagan-cf: Ah sorry the comment didn't update until after I wrote mine

packages/rich-text/src/helpers/validations.ts
z0al: But that's handled by [this code](https://github.com/contentful/field-editors/blob/5fee0717a38a1c83a518ddcf117624fe452f8bfa/packages/rich-text/src/Toolbar/index.tsx#L83). The logic for explicit marks vs implicit is handled in Content apps e.g. web app & studio. In other words, the rich-text field editor should always receive a list of explicit enabled marks/nodes in the `field.validation` attribute.

packages/rich-text/src/plugins/Marks/helpers.ts
aodhagan-cf: This indeed did work!

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
```diff
@@ -274,6 +274,22 @@ describe(
       });
     });
 
+    describe('Superscript and subscript marks', () => {
+      it('works when pasting subscript and superscript from another RT editor', () => {
+        // A simple "hello world" text with marks: superscript and subscript.
+        // Copied from the RT editor
```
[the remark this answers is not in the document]
aodhagan-cf: This is true! have updated the test name

packages/rich-text/src/helpers/validations.ts
aodhagan-cf: Okay cool, thank you for clarifying!

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
```diff
@@ -274,6 +274,22 @@ describe(
       });
     });
 
+    describe('Superscript and subscript marks', () => {
+      it('works when pasting subscript and superscript from a google doc', () => {
```
z0al, line 278: The test title was updated but the snapshot remained unchanged. I think you have to do that manually

z0al (approved):
LGTM. Don't forget to update the snapshot

cypress/integration/rich-text/RichTextEditor.Pasting.spec.ts
aodhagan-cf: Great catch! sorry about that

PR #1281, opened 2022-11-17
fix: explicitly setting sub super styles [TOL-30]

aodhagan-cf:
Explicitly setting superscript and subscript styles as running into issues with consumers normalising the styles

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2414,7 +2414,12 @@
   dependencies:
     "@contentful/rich-text-types" "^15.12.1"
 
-"@contentful/rich-text-types@15.14.0", "@contentful/rich-text-types@^15.12.1":
+"@contentful/rich-text-types@15.14.1":
+  version "15.14.1"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -2414,7 +2414,12 @@
   dependencies:
     "@contentful/rich-text-types" "^15.12.1"
 
-"@contentful/rich-text-types@15.14.0", "@contentful/rich-text-types@^15.12.1":
+"@contentful/rich-text-types@15.14.1":
+  version "15.14.1"
+  resolved "https://registry.yarnpkg.com/@contentful/rich-text-types/-/rich-text-types-15.14.1.tgz#292f4560d67165af0ced1546a22ae77df96325ff"
+  integrity sha512-1sPVDz7MBuCCMWGvfZXQ06SqBTtlLN5GCP5ddU4c5tfatxVXhv2h80/pDCeTZVV2oCMw02YuNmBgql60+0oL4Q==
+
+"@contentful/rich-text-types@^15.12.1":
```
DanweDE, line 2422: Looks like we still got to bump rt types version in the renderer so this can be deduped.

DanweDE (approved):

PR #1287, opened 2022-12-02
feat: dropdown toolbar rich text [TOL-764]

YvesRijckaert:
- [x] Add a dropdown menu to the toolbar after underline
- [x] It triggers a dropdown with hidden options: Superscript, Subscript, Code
- [x] We only show the dropdown menu when: Whenever at least bold , italic and underline are available
- [x] If nothing that would go inside ... is available, we hide it
- [x] If bold, italic and underline aren’t available, we show the options from ... in the toolbar.
- [x] To indicate an item inside the ... menu is toggled, use the same behaviour as for the other marks.
- [x] The menu item should be marked as active.

### DIFF
--- a/packages/rich-text/src/Toolbar/_tests_/toolbar.test.tsx
+++ b/packages/rich-text/src/Toolbar/_tests_/toolbar.test.tsx
@@ -26,14 +26,7 @@ const mockSdk = (marks?: MARKS[]): any => {
       locale: 'en-US',
       validations: [
         {
-          enabledMarks: marks || [
-            'bold',
-            'italic',
-            'underline',
-            'code',
-            'superscript',
-            'subscript',
-          ],
+          enabledMarks: marks || Object.values(MARKS),
         },
       ],
     },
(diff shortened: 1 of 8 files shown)


packages/rich-text/src/Toolbar/_tests_/toolbar.test.tsx
```diff
@@ -25,39 +24,86 @@ beforeEach(() => {
...
       locale: 'en-US',
+      validations: [
+        {
+          enabledMarks: marks || [
+            'bold',
+            'italic',
+            'underline',
+            'code',
+            'superscript',
+            'subscript',
```
MayaGillilanContentful, line 35: We can use the MARKS enum here

packages/rich-text/src/Toolbar/_tests_/toolbar.test.tsx
```diff
@@ -25,39 +24,86 @@ beforeEach(() => {
...
+        'bold',
+        'italic',
+        'underline',
+        'dropdown',
+        'hyperlink',
+        'quote',
+        'ul',
+        'ol',
+        'hr',
+        'table',
```
MayaGillilanContentful, line 72: also here

MayaGillilanContentful (approved):

PR #1288, opened 2022-12-02
fix: handle unavailable marks

MayaGillilan:
Open question: What do you think about the name `unavailableMarks`? Another suggestion was `restrictedMarks`

### DIFF
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -28,14 +28,14 @@ type ConnectedProps = {
   onChange?: (doc: Contentful.Document) => unknown;
   isToolbarHidden?: boolean;
   actionsDisabled?: boolean;
-  unavailableMarks?: string[];
+  restrictedMarks?: string[];
 };
 
 export const ConnectedRichTextEditor = (props: ConnectedProps) => {
   const id = getContentfulEditorId(props.sdk);
   const plugins = React.useMemo(
-    () => getPlugins(props.sdk, props.onAction ?? noop, props.unavailableMarks),
-    [props.sdk, props.onAction, props.unavailableMarks]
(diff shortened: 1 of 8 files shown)


DanweDE (approved):
I think the approach makes sense if this will effectively remove any sub/sup we get in through copy/paste if the new prop is set accordingly. Would be great to get a test to see how it'll take effect when you pass sub/sup.

packages/rich-text/src/plugins/Text/createTextPlugin.ts
```diff
@@ -65,6 +65,19 @@ export function createTextPlugin(): RichTextPlugin {
 
       return editor;
     },
+    normalizer: [
+      {
+        match: () => Text.isText,
```
z0al, line 70: Shouldn't this be `match: Text.isText` ?

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -140,6 +144,7 @@ const RichTextEditor = (props: Props) => {
             onAction={onAction}
             isDisabled={disabled}
             onChange={setValue}
+            unavailableMarks={unavailableMarks}
```
z0al, line 147: I think `restrictedMarks` may be more descriptive but have no strong opinion. This is a public prop so let's spend a few minutes thinking about a good name

z0al (commented):
Looks good overall

packages/rich-text/src/RichTextEditor.tsx
giotiskl: 👍🏻 to what Ahmed said, I'd suggest `restrictedMarks` or perhaps `prohibitedMarks`

DanweDE: `forbiddenMarks`, `marksYouShallNotPass` :mage_man: :laughing:

PR #1307, opened 2022-12-12
feat: upgrade to plate v16 [TOL-812]

aodhagan-cf:


packages/rich-text/src/plugins/List/transforms/toggleList.ts
```diff
@@ -73,30 +76,26 @@ export const toggleList = (editor: PlateEditor, { type }: { type: string }) =>
...
 
       if (
-        listTypes.includes((commonEntry[0] as TElement).type) ||
+        listTypes.includes((commonEntry[0] as Element).type) ||
         commonEntry[0].type === BLOCKS.LIST_ITEM
       ) {
         let listType = commonEntry[0].type;
         if (commonEntry[0].type === BLOCKS.LIST_ITEM) {
-          listType = (Editor.parent(editor, commonEntry[1])[0] as TElement).type;
+          listType = (getParentNode(editor, commonEntry[1] as Location)?.[0] as Element).type;
```
chrishelgert, line 88:
```suggestion
          listType = (getParentNode(editor, commonEntry[1] as Location)?.[0] as Element | undefined)?.type;
```

Due the optional chaining the accessed element should be `Element | undefined`, for safety we should also use optional chaining to access the type

chrishelgert (commented):
Great work, I like the Idea of the wrapper.
I stumbled about some small things, not sure if they are really needed to do or not

aodhagan-cf:
> Great work, I like the Idea of the wrapper. I stumbled about some small things, not sure if they are really needed to do or not

@chrishelgert Thanks so much for your suggested changes super helpful 😄

chrishelgert (approved):
LGTM 🥇

packages/rich-text/src/internal/types/editor.ts
```diff
@@ -0,0 +1,83 @@
...
+  children: TextOrCustomElement[];
+  data: T;
+  isVoid?: boolean;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- TODO: explain this disable
+export type CustomRenderElementProps<T = any, O = any> = Omit<RenderElementProps, 'element'> & {
+  element: CustomElement<T>;
+} & O;
+
```
z0al, line 61: ❗  These are legacy types that should have been removed by the removal of `src/types.ts` file. `CustomText` is now `Text`, `CustomElement` is `Element` , `TextOrCustomELement` is simply the `Node` I think. And we should be using the `RenderElementsProps` type defined below.

packages/rich-text/src/plugins/List/transforms/moveListItems.ts
```diff
@@ -2,22 +2,29 @@
...
+  getParentPath,
+  isAncestorPath,
+} from '../../../internal/queries';
+import { Span, Location, PlateEditor, Path, PathRef } from '../../../internal/types';
 import { moveListItemDown } from './moveListItemDown';
 
 export type MoveListItemsOptions = {
   increase?: boolean;
-  at?: EditorNodesOptions['at'];
+  at?: Location | Span | undefined;
```
z0al, line 20: `EditorNodesOptions` is already exported from `internal/types/editor.ts`

packages/rich-text/src/plugins/List/utils.ts
```diff
@@ -1,77 +1,84 @@
...
-
-import { CustomElement, RichTextEditor } from '../../types';
-
-const isList = (node: CustomElement) =>
-  [BLOCKS.OL_LIST, BLOCKS.UL_LIST].includes(node.type as BLOCKS);
-
-export const hasListAsDirectParent = (editor: RichTextEditor, [, path]: NodeEntry) => {
-  const [parentNode] = (getParent(editor, path) || []) as NodeEntry;
-  return isList(parentNode as CustomElement);
+import { getAboveNode, getBlockAbove, getParentNode } from '@udecode/plate-core';
```
z0al, line 2: Should be imported from `internal`

packages/rich-text/src/plugins/Normalizer/withNormalizer.ts
```diff
@@ -1,23 +1,19 @@
-import { match, getPluginType } from '@udecode/plate-core';
+import { getPluginType } from '@udecode/plate-core';
```
z0al, line 1: Already exported from `internal`

packages/rich-text/src/plugins/Table/createTablePlugin.ts
```diff
@@ -1,18 +1,25 @@
 import { BLOCKS, CONTAINERS } from '@contentful/rich-text-types';
-import { getBlockAbove, getParent, getLastChildPath, WithPlatePlugin } from '@udecode/plate-core';
+import { Value } from '@udecode/plate-core';
```
z0al, line 2: ❗ should use the `Value` from `internal`

packages/rich-text/tsconfig.json
```diff
@@ -4,10 +4,9 @@
   "compilerOptions": {
     "module": "esnext",
     "lib": ["dom", "esnext"],
-    "rootDir": "./src",
-    "baseUrl": "./",
+    "baseUrl": "./src",
     "paths": {
-      "*": ["src/*", "node_modules/*"]
+      "internal/*": ["internal/*"]
```
z0al, line 9: Revert this as well since it's not working?

packages/rich-text/src/plugins/Hyperlink/createHyperlinkPlugin.ts
```diff
@@ -2,17 +2,13 @@ import * as React from 'react';
...
 import { isLinkActive, unwrapLink } from '../../helpers/editor';
 import { transformRemove } from '../../helpers/transformers';
-import {
-  RichTextPlugin,
-  CustomRenderElementProps,
-  CustomElement,
-  RichTextEditor,
-} from '../../types';
+import { PlatePlugin, KeyboardHandler } from '../../internal/types';
+import { CustomRenderElementProps, CustomElement } from '../../internal/types';
```
z0al, line 11: We should not have `CustomRenderElementProps` or `CustomElement` as explained above.

z0al (approved):
Looks good overall. Thank you for the huge effort 👏🏻 

Some comments need to be addressed, but I'm okay with merging this first and creating a follow-up PR to avoid leaving it hanging and dealing with merge conflicts.

aodhagan-cf:
> Looks good overall. Thank you for the huge effort 👏🏻
> 
> Some comments need to be addressed, but I'm okay with merging this first and creating a follow-up PR to avoid leaving it hanging and dealing with merge conflicts.

Thanks so much for all your help along the way, I learned loads ❤️

PR #1309, opened 2022-12-23
chore: adding fallback when urn not in resource link [ZEND-3052]

aodhagan-cf:
Falling back to generating urn when one does not exist on the entity, this was causing the resource editor to not display the xspace references.

### DIFF
--- a/packages/reference/src/resources/useResourceLinkActions.ts
+++ b/packages/reference/src/resources/useResourceLinkActions.ts
@@ -5,22 +5,28 @@ import { EntryProps, WithResourceName } from 'contentful-management';
 
 import { LinkActionsProps } from '../components';
 
-const toLinkItem = (entry: WithResourceName<EntryProps>) => ({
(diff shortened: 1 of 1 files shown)


packages/reference/src/resources/useResourceLinkActions.ts
```diff
@@ -5,29 +5,36 @@ import { EntryProps, WithResourceName } from 'contentful-management';
 
 import { LinkActionsProps } from '../components';
 
-const toLinkItem = (entry: WithResourceName<EntryProps>) => ({
+const toLinkItem = (entry: WithResourceName<EntryProps>, apiUrl: string) => ({
   sys: {
     type: 'ResourceLink',
     linkType: 'Contentful:Entry',
-    urn: entry.sys.urn,
```
aodhagan-cf, line 12: Sometimes this comes in as undefined so falling back to generating

2wce (approved):
LGTM

PR #1313, opened 2023-01-02
feat: lower search and update width entry ref [TOL-247]

YvesRijckaert:
- Display the search when there are 5+ options in the add content dropdown
- Make the dropdown have a reasonable `max-width`: from 240px to 300px


2wce (approved):
LGTM

PR #1314, opened 2023-01-02
fix: solve multi ref drag issue [TOL-823]

2wce:
This fixes the issue with https://contentful.atlassian.net/browse/TOL-823. When reordering multi referenced items the item being dragged would disappear. Giving it a higher zIndex fixes the issue. I chose one of the constants from the f36-tokens package

### DIFF
--- a/packages/reference/src/common/SortableLinkList.tsx
+++ b/packages/reference/src/common/SortableLinkList.tsx
@@ -17,7 +17,7 @@ const styles = {
   }),
   item: css({
     marginBottom: tokens.spacingM,
-    zIndex: tokens.zIndexModalContent, // setting this to an index above 99 fixes [TOL-823]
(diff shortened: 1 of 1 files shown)


packages/reference/src/common/SortableLinkList.tsx
```diff
@@ -17,6 +17,7 @@ const styles = {
   }),
   item: css({
     marginBottom: tokens.spacingM,
+    zIndex: tokens.zIndexModalContent, // setting this to an index above 99 fixes [TOL-823]
```
MayaGillilanContentful, line 20: I think instead of giving the ticket number here it would be better to describe the issue briefly (e.g. `fixes dragged item disappearing`). People will see the ticket number in the git blame

MayaGillilanContentful (approved):

PR #1315, opened 2023-01-02
fix: a11y issues focus rich text

YvesRijckaert:
Fixes some A11Y issues with tabbing in rich text.
Rich Text Tables plugin introduced a bug where you could not tab _out_ of the rich text field anymore. This PR should fix that.
Also when you are in a table and you want to exit the rich text editor with a keyboard, you can now use the ESC key. Previously this was impossible with a keyboard.


giotiskl (approved):

z0al:
Could you add a Cypress test for the tabbing issue?

PR #1316, opened 2023-01-03
test: rich text keyboard navigation [TOL-141]

YvesRijckaert:
Add a test for keyboard navigation for rich text editor

Follow-up for https://github.com/contentful/field-editors/pull/1315


2wce (approved):
LGTM

PR #1317, opened 2023-01-05
feat: custom actions in resource reference editors [DANTE-793]

veu:
Allow custom actions to be passed to resource reference editors via the existing `renderCustomActions` parameter.

### DIFF
--- a/packages/reference/src/components/LinkActions/LinkEntityActions.tsx
+++ b/packages/reference/src/components/LinkActions/LinkEntityActions.tsx
@@ -11,6 +11,7 @@ import {
   Asset,
   NavigatorSlideInfo,
 } from '../../types';
+import { CombinedLinkActions } from './CombinedLinkActions';
 import { createEntity, selectMultipleEntities, selectSingleEntity } from './helpers';
 import { LinkActions, LinkActionsProps } from './LinkActions';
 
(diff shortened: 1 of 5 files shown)


packages/reference/src/resources/SingleResourceReferenceEditor.tsx
```diff
@@ -41,7 +42,10 @@ export function SingleResourceReferenceEditor(
             />
           ) : (
             // TODO: support custom actions once publicly available
```
Silhoue, line 44: Should we delete the comment now?

packages/reference/src/components/LinkActions/LinkEntityActions.tsx
```diff
@@ -177,3 +178,16 @@ export function LinkEntityActions({
...
+  renderCustomActions,
+  ...props
+}: LinkActionsProps & {
+  renderCustomActions?: (props: LinkActionsProps) => React.ReactElement;
+}) {
+  const renderLinkActions = renderCustomActions
+    ? renderCustomActions
+    : (props: LinkActionsProps) => <CombinedLinkActions {...props} />;
+
+  return renderLinkActions(props);
```
Silhoue, line 192: Would this not work? I find it a bit more readable
```suggestion
  return renderCustomActions ? renderCustomActions(props) : <CombinedLinkActions {...props} />;
```

Silhoue (approved):

packages/reference/src/components/LinkActions/LinkEntityActions.tsx
veu: Sure

packages/reference/src/resources/SingleResourceReferenceEditor.tsx
veu: Indeed. I’ve removed it.

PR #1320, opened 2023-01-10
fix: enter key open asset modal [TOL-889]

YvesRijckaert:
Pressing enter key should fire the onEdit event for asset field


giotiskl (approved):

packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx
```diff
@@ -138,6 +139,18 @@ export const WrappedAssetCard = (props: WrappedAssetCardProps) => {
             }
           : undefined
       }
+      /* todo - remove this when onKeyDown is allowed as a prop for BaseCard in forma 36
+      // @ts-expect-error */
+      onKeyDown={
```
chrishelgert, line 144: ❓ Don't know how it is handled, but when testing i found out that the WrappedAssetLink would also miss the onKeyDown handler, i would assume that it is out of this ticket?

PR #1327, opened 2023-01-17
fix(rte): [TOL-989] command prompt not always visible and working

chrishelgert:
- missing z-index on the poper (not visible in the web app and also some visual issues in compose)
- added support for command prompt in other keyboards (QWERTZ, ...)

### DIFF
--- a/packages/rich-text/src/plugins/CommandPalette/onKeyDown.spec.ts
+++ b/packages/rich-text/src/plugins/CommandPalette/onKeyDown.spec.ts
@@ -1,0 +1,32 @@
+import * as internal from '../../internal/transforms';
+import { createTestEditor } from '../../test-utils';
+import { COMMAND_PROMPT } from './constants';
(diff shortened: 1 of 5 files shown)


packages/rich-text/src/plugins/CommandPalette/onKeyDown.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { editor } = createTestEditor({});
+
+  const addMark = jest.spyOn(internal, 'addMark');
+  const onCommandPaletteAction = jest.spyOn(editor.tracking, 'onCommandPaletteAction');
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  test.each`
```
YvesRijckaert, line 16: Nice!

YvesRijckaert (approved):
LGTM

PR #1331, opened 2023-01-20
fix: mail to links replace ampersand [TOL-923]

YvesRijckaert:
We use `dompurify` to sanitize the markdown preview output. Dompurify uses innerHTML at some point, which transforms `&` to`&amp` inside of href links. This causes an issue with mailto links not working as expected. Everything after `&amp` gets ignored.

For example: `<a href="mailto:test@test.com?subject=test&body=test">Test</a>`
turns into:
`<a href="mailto:test@test.com?subject=test&amp;body=test">Test</a>`, which makes the `mailto` link ignore the body parameter.

I added a regex that transforms the `&amp` to `&` again inside mail to links. Unfortunately I was not able to add any exclusion inside the dompurify package (see this issue in the package: https://github.com/cure53/DOMPurify/issues/379#issuecomment-571997013)

### DIFF
--- a/packages/markdown/src/components/MarkdownPreview.tsx
+++ b/packages/markdown/src/components/MarkdownPreview.tsx
@@ -1,9 +1,12 @@
 import React from 'react';
-import Markdown from 'markdown-to-jsx';
+
 import tokens from '@contentful/f36-tokens';
 import DOMPurify from 'dompurify';
 import { css, cx } from 'emotion';
+import Markdown from 'markdown-to-jsx';
+
 import { EditorDirection, PreviewComponents } from '../types';
+import { replaceMailtoAmp } from '../utils/replaceMailtoAmp';
 
 const styles = {
   root: css`
@@ -199,7 +202,7 @@ export const MarkdownPreview = React.memo((props: MarkdownPreviewProps) => {
   // See the list of allowed Tags here:
   // https://github.com/cure53/DOMPurify/blob/main/src/tags.js#L3-L121
   const cleanHTML = React.useMemo(() => {
-    return DOMPurify.sanitize(props.value);
+    return replaceMailtoAmp(DOMPurify.sanitize(props.value));
   }, [props.value]);
 
   return (
--- a/packages/markdown/src/utils/replaceMailtoAmp.spec.ts
+++ b/packages/markdown/src/utils/replaceMailtoAmp.spec.ts
@@ -1,0 +1,27 @@
+import { replaceMailtoAmp } from './replaceMailtoAmp';
+
+describe('replace inside mailto', () => {
+  let str =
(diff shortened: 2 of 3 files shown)


aodhagan-cf (approved):
Looks good! 👍

z0al (changes requested):
Sorry for blocking but this gets modified by `DOMPurify` to avoid security issues. I don't recommend altering the behavior this way. 

Either we continue to sanitize the attribute and communicate that in our docs or we add an exclusion rule for the `mailto` attribute (see [here](https://github.com/cure53/DOMPurify)). Please consult our security team

YvesRijckaert:
> Sorry for blocking but this gets modified by `DOMPurify` to avoid security issues. I don't recommend altering the behavior this way.
> 
> Either we continue to sanitize the attribute and communicate that in our docs or we add an exclusion rule for the `mailto` attribute (see [here](https://github.com/cure53/DOMPurify)). Please consult our security team

I will contact the security team for a review. The exclusion does not work since dompurify uses innerHtml under the hood, which transforms the `&` (see the linked issue)
Let's see what the security team says, if we cannot ship this change, let's go for option 1.

roryscarson:
@z0al Can you explain the concern to me? DOMPurify is still being used on the initial input and the new replace function only accepts the input string, not a value to replace, so it can't be as easily manipulated. This seems fine to me but I could be missing something.

z0al:
> @z0al Can you explain the concern to me? DOMPurify is still being used on the initial input and the new replace function only accepts the input string, not a value to replace, so it can't be as easily manipulated. This seems fine to me but I could be missing something.

The idea behind DOMPurify is to sanitize the markdown output we are previewing within Contentful web app to avoid XSS. With this change, we are taking the sanitized output and changing it again undoing some of the sanitization done by the library. 

I'm not a subject matter expert, and not sure if this is will be enough to open a security threat but it raised a flag for me. I'm fine with going with the security team's recommendation.

roryscarson (approved):

packages/markdown/src/components/MarkdownPreview.tsx
```diff
@@ -199,7 +202,7 @@ export const MarkdownPreview = React.memo((props: MarkdownPreviewProps) => {
   // See the list of allowed Tags here:
   // https://github.com/cure53/DOMPurify/blob/main/src/tags.js#L3-L121
   const cleanHTML = React.useMemo(() => {
-    return DOMPurify.sanitize(props.value);
+    return replaceMailtoAmp(DOMPurify.sanitize(props.value));
```
karankohli-cf, line 205: might want to look at 
https://github.com/cure53/DOMPurify/blob/f5c3e0b3f8c17b4d3fa0a827038f0a91e2e244ad/README.md?plain=1#L280

YvesRijckaert: Unfortunately this did not fix the issue

PR #1333, opened 2023-01-24
refactor: fixing types and imports [TOL-917]

aodhagan-cf:
Feedback from: https://github.com/contentful/field-editors/pull/1307

### DIFF
--- a/packages/.eslintrc.js
+++ b/packages/.eslintrc.js
@@ -35,7 +35,7 @@ module.exports = {
       {
         'ts-expect-error': false,
         'ts-ignore': true,
-        'ts-nocheck': false,
+        'ts-nocheck': true,
         'ts-check': true,
       },
     ],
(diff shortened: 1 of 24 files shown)


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -4,14 +4,14 @@ import { FieldExtensionSDK } from '@contentful/app-sdk';
 import { EntityProvider } from '@contentful/field-editor-reference';
 import { FieldConnector } from '@contentful/field-editor-shared';
 import * as Contentful from '@contentful/rich-text-types';
-// FIXME: extract this out to internal folder?
 import { Plate, getPlateActions } from '@udecode/plate-core';
```
z0al, line 7: Intentionally didn't export this from `internal`?

z0al (approved):
Looks good. Thank you for taking care of that

packages/rich-text/src/RichTextEditor.tsx
aodhagan-cf: Yeah, decided that this makes sense as one place to actually use Plate as is. It felt fine to use here as it's the root of the project, what do you think?

aodhagan-cf: I don't think the Plate component will be used more than once

PR #1335, opened 2023-01-31
fix: remove command mark after space [TOL-945]

YvesRijckaert:
A customer reached out that they are having troubles with the rich text commands plugin.
They have content that contains a slash /. Entering the slash forces the RTE to start a new span (because of the rich text commands plugin) and the slash is not connected to previous content anymore.
They have a workaround for this, to press ESC; But this also blurs the editor and they have to click on the RTE again.

This PR removes the span (rich text commands mark) when the user types a space after typing a slash /.

The only downside to this approach is that the search functionality will not work anymore when you type a space (since the rich text command gets removed).

### DIFF
--- a/packages/rich-text/src/plugins/CommandPalette/onKeyDown.ts
+++ b/packages/rich-text/src/plugins/CommandPalette/onKeyDown.ts
@@ -36,6 +36,10 @@ export const createOnKeyDown = (): KeyboardHandler => {
         }
       }
 
+      if (isHotkey('space', event)) {
+        removeMark(editor, COMMAND_PROMPT, range);
+      }
+
       if (isHotkey('escape', event)) {
         removeMark(editor, COMMAND_PROMPT, range);
         editor.tracking.onCommandPaletteAction('cancelRichTextCommandPalette');
--- a/packages/rich-text/src/plugins/CommandPalette/onKeyDown.ts
+++ b/packages/rich-text/src/plugins/CommandPalette/onKeyDown.ts
@@ -1,6 +1,7 @@
 import isHotkey from 'is-hotkey';
 
 import { getRange, getAboveNode, isMarkActive, addMark, removeMark } from '../../internal';
+import { focusEditor } from '../../internal/misc';
 import { KeyboardHandler, NodeEntry } from '../../internal/types';
 import { COMMAND_PROMPT } from './constants';
 
@@ -36,13 +37,11 @@ export const createOnKeyDown = (): KeyboardHandler => {
         }
       }
 
-      if (isHotkey('space', event)) {
-        removeMark(editor, COMMAND_PROMPT, range);
-      }
-
       if (isHotkey('escape', event)) {
+        event.stopPropagation();
         removeMark(editor, COMMAND_PROMPT, range);
         editor.tracking.onCommandPaletteAction('cancelRichTextCommandPalette');
+        focusEditor(editor);
       }
     };
   };


packages/rich-text/src/plugins/CommandPalette/onKeyDown.ts
```diff
@@ -36,6 +36,10 @@ export const createOnKeyDown = (): KeyboardHandler => {
         }
       }
 
+      if (isHotkey('space', event)) {
```
chrishelgert, line 39: couldn't we close/remove it on every key except arrow keys and enter?

for example : `contentful.com/developers` would also have this issue

YvesRijckaert: Well we have a search enabled for rich text commands. Typing something after the slash will sort the entries based on what you typed

YvesRijckaert: So this would remove the search completely. Now it is only removed when you type a space

aodhagan-cf: How about escape only closes the command palette and keeps focus?

aodhagan-cf: Then the second time you press escape it closes the command palette?

YvesRijckaert: Not sure if I understood correctly, but this is exactly what they complained about, they don't want to use the escape button since it blurs the editor and they'd have to focus it again

aodhagan-cf: I mean change the behaviour so that we don't lose focus the first time escape is clicked, it might involved focusing again in code to the point they are currently at

aodhagan-cf: - /
- "some text"
- [ESCAPE] _closes command palette but keeps focus in editor_
- [ESCAPE] _loses focus_

YvesRijckaert: We would still have the issue where the user types content with a slash, does not want anything from the rich text commands palette, and keeps typing... the content is still split up:

Screenshot 2023-01-31 at 16 04 50

aodhagan-cf: https://github.com/contentful/field-editors/blob/7fbebc371800f12a12adbd7278c22449c6ff85f1/packages/rich-text/src/internal/misc.ts#L23 we have this function

YvesRijckaert: If we go that route I think we should have a clear UI that shows that the rich text command span is currently applied, and that they need to press ESC to remove it

chrishelgert: Maybe this works for using space inside a search:
```
if (isHotkey(['backspace', 'space'], event)) {
  // if it is the last character in the command string
  if (range.focus.offset - range.anchor.offset === 1) {
    removeMark(editor, COMMAND_PROMPT, range);
  }
}
 ```
 
-  `/` + `space` -> remove command prompt
- `/` + `entry editor` -> still open and can search

YvesRijckaert: This would only remove the mark when you type a space _right after_ the slash right?
So what if you type:

> You can use /slash to open the command palette.

This would still apply the two spans :/

chrishelgert: ah thats sad, it should probably also remove the mark if it is closed without applying content right?

z0al, line 39: I'm not a fan of this

> The only downside to this approach is that the search functionality will not work anymore when you type a space (since the rich text command gets removed).

I think that's a huge downside. Breaking the search functionality may lead to more customer issues. Let's re-think this holistically and find a proper fix. We can start by looking at other similar RT implementations and how they handle that and what are our options.

> to press ESC; But this also blurs the editor and they have to click on the RTE again.

Have you tried re-focusing in the editor in the escape block below? We probably want it to only re-focus when there is a `COMMAND_PROMPT` mark (you can use `isMarkActive` for that) otherwise it may cause accessibility issues.

z0al: > We would still have the issue where the user types content with a slash, does not want anything from the rich text commands palette, and keeps typing... the content is still split up:

What's the issue with the split up? If it's bad we can normalize that in the editor or in the preprocessing stage before calling `onChange`. Also, Slate should automatically normalize that when the two text nodes share the same marks. See rule 2 [here](https://docs.slatejs.org/concepts/11-normalizing)

YvesRijckaert: I agree we need to fix this holistically. Ideally we would want a UI for when the command palette mark is active, because now if you search for something and there are no results, everything you type will still be inside the command palette mark.

I will create a ticket for a better fix in the backlog.

For now I implemented the ESC solution. After typing ESC once, it will remove the span, and ESC again will blur the editor. I think we could ship that to unblock the customer

aodhagan-cf (approved):

PR #1338, opened 2023-02-06
fix: improve types

jfctfl:
`getAssetTitle` incorrectly used type Entry as a type. Asset is not compatible with the function signature of `getFieldValue` because of the special type signature of Assets.fields.file => to fix this we now pass the title field only to `getFieldValue`. 

`useEntityInfo` now requires the target to be a `Link<ContentEntityType>` instead of a generic string.

### DIFF
--- a/packages/_shared/src/utils/entityHelpers.ts
+++ b/packages/_shared/src/utils/entityHelpers.ts
@@ -1,7 +1,8 @@
 import get from 'lodash/get';
 import isObject from 'lodash/isObject';
 import isString from 'lodash/isString';
-import { File, ContentType, Entry, ContentTypeField } from '../typesEntity';
+
+import { Asset, ContentType, ContentTypeField, Entry, File } from '../typesEntity';
 
 function titleOrDefault(title: string | undefined, defaultTitle: string): string {
   if (!isString(title)) {
@@ -49,13 +50,13 @@ export function getAssetTitle({
   defaultLocaleCode,
   defaultTitle,
 }: {
-  asset: Entry;
+  asset: Asset;
   localeCode: string;
   defaultLocaleCode: string;
   defaultTitle: string;
 }) {
   const title = getFieldValue({
-    entity: asset,
+    entity: { fields: { title: asset.fields?.title } },
     fieldId: 'title',
     localeCode,
     defaultLocaleCode,
@@ -71,6 +72,7 @@ export function getAssetTitle({
  */
 export const isAssetField = (field: ContentTypeField): boolean =>
   field.type === 'Link' && field.linkType === 'Asset';
+
 /**
  * Returns true if field is a Title
  */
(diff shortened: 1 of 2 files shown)


packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -27,7 +28,7 @@ export function getFieldValue({
   defaultLocaleCode,
 }: {
   entity: {
-    fields: { [key: string]: { [valueKey: string]: string | undefined } };
+    fields: Record<string, Record<string, string | undefined>>;
```
andipaetzold, line 31: What's the benefit of this change?

yarn.lock
```diff
@@ -1708,6 +1708,11 @@
     open "^8.0.5"
     ora "^5.4.0"
 
+"@contentful/app-sdk@^4.15.0":
+  version "4.15.0"
+  resolved "https://registry.yarnpkg.com/@contentful/app-sdk/-/app-sdk-4.15.0.tgz#ce35116168b59f3c1e04ac4f869c18a4bc874f3e"
+  integrity sha512-dcawQN7fAOVmKsrAykpJB8O6hkBH2b9jWAn3/Ts/h16xUKs5R7409HPt3akB1pBKiifJYRMuTZAYC6xkj17+nw==
+
 "@contentful/app-sdk@^4.2.0", "@contentful/app-sdk@^4.3.0", "@contentful/app-sdk@^4.6.0":
```
andipaetzold, line 1716: Should we try to get everything to the same version and dedupe App SDK?

jfctfl: Good idea, done!

jfctfl: > * Does this change the interface for the users?
`useEntityInfo` has now more strict types, but it's not like we supported anything for other string values than `Entry` and `Asset`. 
> * Do we have to bump the peer dependencies? This normally requires a breaking change as we drop support for old versions
I updated the app-sdk devDependencies for all relevant packages and thought it would be odd to have a lower peerDependency requirement. But technically you are right, it is not required. Would you prefer to revert that?
> * Do we want to release a new version?
Yes, I want to release a version which is then used in experience packages to align the versions of app-sdk used in this repo with the one in experience packages.

ronaldronson (approved):

packages/rich-text/src/plugins/Hyperlink/useEntityInfo.ts
```diff
@@ -1,6 +1,7 @@
-import { useState, useEffect } from 'react';
+import { useEffect, useState } from 'react';
 
-import { FieldExtensionSDK, Link, Entry, Asset, ScheduledAction } from '@contentful/app-sdk';
+import { Asset, Entry, FieldExtensionSDK, Link, ScheduledAction } from '@contentful/app-sdk';
+import { ContentEntityType } from '@contentful/app-sdk/dist/types/utils';
```
andipaetzold, line 4: Only import from `@contentful/app-sdk` directly

andipaetzold (changes requested):

andipaetzold (approved):

PR #1340, opened 2023-02-08
chore: updating plate version to 20 [TOL-943]

aodhagan-cf:
Updating plate to latest version.

Still WIP


packages/rich-text/src/plugins/Text/__tests__/createTextPlugin.test.tsx
```diff
@@ -28,7 +28,7 @@ describe('delete backward', () => {
     },
     {
       title:
-        'deletes the empty paragraph at the beginning of the RTE followed by another paragraph',
+        'does not delete the empty paragraph at the beginning of the RTE followed by another paragraph',
```
aodhagan-cf, line 31: No other Rich Text I tested against has this behaviour, I argue that we should do the opposite and not remove the empty paragraph at the beginning

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -31,23 +31,42 @@ type ConnectedProps = {
...
 
-  const onValueChanged = useOnValueChanged({
-    editorId: id,
-    handler: props.onChange,
-    skip: pendingExternalUpdate || isFirstRender,
-    onSkip: () => setPendingExternalUpdate(false),
-  });
+  const editor = usePlateSelectors(id).editor();
+
+  // TODO: double check this
```
aodhagan-cf, line 63: Need to verify this isn't breaking stuff

cypress/integration/rich-text/__snapshots__/RichTextLists.spec.ts.snap
aodhagan-cf, line 1: Double check this

cypress/integration/rich-text/RichTextEditor.spec.ts
```diff
@@ -1654,11 +1648,13 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
 
           // Inserts paragraph before embed because it's in the first line.
           richText.editor.get('[data-entity-id="example-entity-id"]').first().click();
-          pressEnter();
+          // TODO add key config object back
+          pressEnter({ force: true });
 
-          // inserts paragraph in-between embeds.
+          // // inserts paragraph in-between embeds.
```
aodhagan-cf, line 1654: remove extract comment //

PR #1342, opened 2023-02-10
fix(rte): [TOL-960] conditional command palette plugin enablement

chrishelgert:
Only enable the command palette plugin if at least one action type is allowed

https://contentful.atlassian.net/browse/TOL-960

### DIFF
--- a/packages/rich-text/src/plugins/CommandPalette/useCommands.ts
+++ b/packages/rich-text/src/plugins/CommandPalette/useCommands.ts
@@ -47,14 +47,32 @@ const removeQuery = (editor: PlateEditor) => {
   }
 };
 
+export function isCommandPromptPluginEnabled(sdk: FieldExtensionSDK) {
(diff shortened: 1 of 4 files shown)


packages/rich-text/src/plugins/CommandPalette/useCommands.ts
```diff
@@ -47,14 +47,32 @@ const removeQuery = (editor: PlateEditor) => {
...
+  const assetsAllowed = isNodeTypeEnabled(sdk.field, BLOCKS.EMBEDDED_ASSET);
+
+  return {
+    inlineAllowed,
+    entriesAllowed,
+    assetsAllowed,
+  };
+}
 
+function getCommandPermissons(sdk: FieldExtensionSDK, editor: PlateEditor) {
```
z0al, line 62: A small typo
```suggestion
function getCommandPermissions(sdk: FieldExtensionSDK, editor: PlateEditor) {
```

z0al (approved):
Thank you for taking care of this. Care about adding a cypress test? could be in a follow up

aodhagan-cf (approved):

PR #1350, opened 2023-02-24
fix(markdown-editor): textlink sometimes not centered in help text []

chrishelgert:
Fixes the issue that the TextLink inside the help text is not always centered.
Using flex would need more adjustments as it also affects the whitespaces.

# Current
Screenshot 2023-02-24 at 12 01 49

# Flex
Screenshot 2023-02-24 at 12 00 23

# Line-Height
Screenshot 2023-02-24 at 12 00 57


MayaGillilanContentful (approved):

denkristoffer (approved):
🙇

PR #1353, opened 2023-02-27
fix: do not focus RTE when another slide is open [TOL-989]

DanweDE:
In the rich text editor, when embedding an entry or asset block, ensure that we take into account the entity selector option of `withCreate: true` which can cause the slide-in editor to open (web app).

When the slide-in editor opens, then we only want to re-focus the rich text editor after the slide-in editor (and any other slides opened on top of that) closes.

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
+++ b/packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
@@ -2,8 +2,9 @@ import { BLOCKS, TEXT_CONTAINERS } from '@contentful/rich-text-types';
 
 import { focus, insertEmptyParagraph, moveToTheNextChar } from '../../helpers/editor';
 import newEntitySelectorConfigFromRichTextField from '../../helpers/newEntitySelectorConfigFromRichTextField';
+import { watchCurrentSlide } from '../../helpers/sdkNavigatorSlideIn';
 import { getText, getAboveNode, getLastNodeByLevel } from '../../internal/queries';
-import { select, insertNodes, setNodes } from '../../internal/transforms';
+import { insertNodes, setNodes } from '../../internal/transforms';
 import { PlateEditor } from '../../internal/types';
 import { TrackingPluginActions } from '../../plugins/Tracking';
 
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
```diff
@@ -21,19 +22,22 @@ export async function selectEntityAndInsert(
...
   if (!entity) {
     logAction('cancelCreateEmbedDialog', { nodeType });
-    return;
+  } else {
+    insertBlock(editor, nodeType, entity);
+    ensureFollowingParagraph(editor);
+    logAction('insert', { nodeType });
   }
-
-  select(editor, selection);
```
DanweDE, line 31: @aodhagan-cf @MayaGillilanContentful @YvesRijckaert any thoughts on removing this? I did all kind of tests, but couldn't see any negative impact. Same goes for Cypress tests which are still running.

This line seems odd to me, maybe it did serve a purpose in a past Slate/Plate version, but right now I don't see how the selection should be gone and why we have to reset it, it seems like ne only thing that's necessary is to re-focus the editor and the cursor position is in the right place.

aodhagan-cf: I would say if it passes unit tests and you can't see anything glaringly obvious on problems it causes happy for you to remove it. If it is a problem in future, we write a test to capture that

packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
```diff
@@ -21,19 +22,22 @@ export async function selectEntityAndInsert(
     baseConfig.entityType === 'Asset' ? dialogs.selectSingleAsset : dialogs.selectSingleEntry;
   const config = { ...baseConfig, withCreate: true };
 
-  const { selection } = editor;
+  const rteSlide = watchCurrentSlide(sdk.navigator);
```
aodhagan-cf, line 25: What happens in Compose?

packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
DanweDE: Yes, if nothing pops up, we can probably remove it in a whole lot of places. But waiting for @z0al perhaps he can shed some light.

packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
```diff
@@ -21,19 +22,22 @@ export async function selectEntityAndInsert(
...
   const entity = await selectEntity(config);
+
   if (!entity) {
     logAction('cancelCreateEmbedDialog', { nodeType });
-    return;
+  } else {
+    insertBlock(editor, nodeType, entity);
```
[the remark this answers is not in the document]
aodhagan-cf: https://github.com/contentful/field-editors/blob/7dbb4eceff641fa96a1a3114bb86df4871a0c9f9/packages/rich-text/src/plugins/CommandPalette/utils/insertBlock.ts#L33

Here is the focus that is causing the problem

packages/rich-text/src/plugins/EmbeddedEntityBlock/Util.ts
DanweDE: In this case, yes. There are no slide events.

In some cases, e.g. calling `sdk.navigator.openEntry(entryId, {slideIn: true})` we open  the web app instead, `{slideIn: true}` is simply ignored and also doesn't trigger any slide-in events.

YvesRijckaert (approved):

aodhagan-cf (approved):

PR #1354, opened 2023-03-03
fix: 🐛 do not focus RTE on inline embed + create entry

DanweDE:
Same as #1353 but for inline embeds. When we create a new entry for an inline embed in the web app, the slide-in editor opens. In this case we  want to focus the slided in entry editor's first field, not the rich text editor in the background tab. Only after the slide closes again, we will focus the rich text editor again.

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
@@ -15,8 +15,9 @@ import { useContentfulEditor } from '../../ContentfulEditorProvider';
 import { focus, moveToTheNextChar } from '../../helpers/editor';
 import { IS_CHROME } from '../../helpers/environment';
 import newEntitySelectorConfigFromRichTextField from '../../helpers/newEntitySelectorConfigFromRichTextField';
+import { watchCurrentSlide } from '../../helpers/sdkNavigatorSlideIn';
 import { findNodePath } from '../../internal/queries';
-import { setSelection, insertNodes, removeNodes } from '../../internal/transforms';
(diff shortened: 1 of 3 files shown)


YvesRijckaert (approved):

packages/rich-text/src/plugins/EmbeddedEntityInline/index.tsx
```diff
@@ -120,26 +123,18 @@ async function selectEntityAndInsert(
...
-
+  const rteSlide = watchCurrentSlide(sdk.navigator);
   const entry = await sdk.dialogs.selectSingleEntry<Entry>(config);
-  focus(editor); // Dialog steals focus from editor, return it.
 
   if (!entry) {
     logAction('cancelCreateEmbedDialog', { nodeType: INLINES.EMBEDDED_ENTRY });
-    return;
+  } else {
+    insertNodes(editor, createInlineEntryNode(entry.sys.id));
```
chrishelgert, line 132: ❓ Not sure, but don't we need the setSelection to replace the content with the new inserted Node?

chrishelgert: ok nvm still works

DanweDE: Yes, I also thought about it and tested that and other cases, but I couldn't find anything broken. E.g.

https://user-images.githubusercontent.com/101926/222710048-8ed8a4d9-84ce-4821-b243-c601be589911.mp4

<br/>
Also, see the [discussion here](https://github.com/contentful/field-editors/pull/1353#discussion_r1119819940) where I ask about potential issues.

PR #1362, opened 2023-03-07
fix(RTE): remove whitespaces only in tables and lists [TOL-104]

chrishelgert:
Currently when pasting in a text with links inside, the whitespace before the links are removed.
After this changes it will now only remove whitespaces that can lead to misleading behaviour (inside of tables and lists)


cypress/integration/rich-text/__snapshots__/RichTextEditor.Pasting.spec.ts.snap
```diff
@@ -2124,7 +2124,7 @@ exports[`Rich Text Editor > Microsoft Word (.docx) deserialization > paragraphs,
           "data": {},
           "marks": [],
           "nodeType": "text",
-          "value": "marks and"
+          "value": " marks and "
```
YvesRijckaert, line 2127: is there any concern with the extra spaces here?
Maybe @z0al could have a look at this PR before merging?

z0al:
@chrishelgert and maybe @contentful/team-tolkien  I was playing around with the [latest Table plugin](https://codesandbox.io/s/github/udecode/plate-playground) from Plate and it doesn't seem to suffer from the issue covered in the tests. The editor also doesn't have the bug with hyperlink spacing. It seems like a few plugins have changed a lot since we integrated tables. I wonder if it is worth experimenting a bit by updating our relevant code.

Keep in mind that this likely won't come out of the box when upgrading Plate (at least some features won't) because we have to opt-in manually in our plugin implementations (this affects both tables and lists). That's because we are [overriding what gets registered on plugin initialization](https://github.com/contentful/field-editors/blob/096b686fa006245f0481bdd9dc557009266951fe/packages/rich-text/src/plugins/Table/createTablePlugin.ts#L38-L52)

Since we have to upgrade Plate anyway soon it may be worth the effort in my opinion. What do you think?

chrishelgert:
> @chrishelgert and maybe @contentful/team-tolkien I was playing around with the [latest Table plugin](https://codesandbox.io/s/github/udecode/plate-playground) from Plate and it doesn't seem to suffer from the issue covered in the tests. The editor also doesn't have the bug with hyperlink spacing. It seems like a few plugins have changed a lot since we integrated tables. I wonder if it is worth experimenting a bit by updating our relevant code.
> 
> Keep in mind that this likely won't come out of the box when upgrading Plate (at least some features won't) because we have to opt-in manually in our plugin implementations (this affects both tables and lists). That's because we are [overriding what gets registered on plugin initialization](https://github.com/contentful/field-editors/blob/096b686fa006245f0481bdd9dc557009266951fe/packages/rich-text/src/plugins/Table/createTablePlugin.ts#L38-L52)
> 
> Since we have to upgrade Plate anyway soon it may be worth the effort in my opinion. What do you think?

We've already a [ticket](https://contentful.atlassian.net/browse/TOL-943) and a [draft pr](https://github.com/contentful/field-editors/pull/1340/files) for the update, we should adjust the AC, then estimate it and then have a look how we can bring it into the roadmap 👍 
I think we already thought about having a look again at the implementation and our custom code around the RTE regarding some safari issues.

As it might take some time to get the update shipped, I would suggest that we ship the current workarounds and check with/after the update if we can get rid of them again, what do you think?

cypress/integration/rich-text/__snapshots__/RichTextEditor.Pasting.spec.ts.snap
z0al: @chrishelgert could you explain why we needed to update the snapshot in this case?

chrishelgert: The whitespaces were removed from the previous replace regex, with the adjustment and only cleaning up tables this snapshot changed.
In this case it removed the space between <bold>, <a>, .. and regular text, which we wanted to fix

z0al (approved):

PR #1364, opened 2023-03-07
fix(rich-text): pasting hyperlinks with blocks

z0al:
Slate removes blocks (e.g. `h1`) inside inline elements (`<a>`) resulting in a loss of content. This PR unwraps all unsupported elements to preserve the inner content.


chrishelgert (approved):
LGTM

PR #1367, opened 2023-03-07
fix: support disabling single line editor

martin3walker:


packages/single-line/src/SingleLineEditor.tsx
```diff
@@ -56,7 +61,7 @@ export function SingleLineEditor(props: SingleLineEditorProps) {
               className={direction === 'rtl' ? styles.rightToLeft : ''}
               isRequired={field.required}
               isInvalid={errors.length > 0}
-              isDisabled={disabled}
+              isDisabled={props.isDisabled ?? disabled}
```
z0al, line 64: How about we implement this in the [FieldConnector](https://github.com/contentful/field-editors/blob/master/packages/_shared/src/FieldConnector.ts) so we have a single place that computes whether a field is disabled or not? Currently, the FieldConnector already handles `isInitiallyDisabled` and manages the `disabled` attribute we get in the callback.

martin3walker: The `isIntiallyDisabled` prop doesn't seem to be working properly. The editor is immediately reset to enabled from the context of app even if `isInitiallyDisabled` is set to `true`. Is this something Tolkien is aware of ?

packages/_shared/src/FieldConnector.ts
```diff
@@ -61,7 +62,7 @@ export class FieldConnector<ValueType> extends React.Component<
       externalReset: 0,
       value: initialValue,
       lastRemoteValue: initialValue,
-      disabled: props.isInitiallyDisabled,
+      disabled: props.isDisabled ?? props.isInitiallyDisabled,
```
martin3walker, line 65: @z0al How does this look?

z0al: This will have the same effect as `isInitiallyDisabled`. If you want to keep the field disabled no matter what if `isDisabled` is passed then you need to conditionally [subscribe to `isDisabledChanged` updates](https://github.com/contentful/field-editors/blob/32bfe1dfaf6d976c69690a99b69b5beff570eba6/packages/_shared/src/FieldConnector.ts#L106-L110) here.

```javascript
if(isDisabled) { 
    this.unsubscribeDisabled = field.onIsDisabledChanged((disabled: boolean) => {
      this.setState({
        disabled,
      });
    });
}
```

However, I'd encourage you to look into why `isInitiallyDisabled` didn't work for your use case first

martin3walker: Hey @z0al we took a slightly different approach, but I think would work a bit better because it would still allow for the listener to be subscribed while also allowing for a manual override based on the `isDisabled` prop.

The problem is that the `field.onIsDisabledChanged` method is being provided by the `app sdk`, which takes the state of of the field from the web app. The web app considers the field to be enabled, because it just sees it as rendering an app. It doesn't understand what is happening inside the app. So when the `field.onIsDisabledChanged` gets called on the initial render it immediately switches the field editor within the app to enabled.

z0al (approved):

PR #1374, opened 2023-03-14
fix: not removing whitespaces around inline elements [TOL-1041]

aodhagan-cf:
We have an issue where when you paste from Microsoft word it will remove the spaces around elements with a background color.

- This PR adjusts the regex to first trim the start and end of a block element
- Then removes empty elements from the end of a block

### DIFF
--- a/packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
+++ b/packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
@@ -56,8 +56,6 @@ export const sanitizeHTML = (html: string): string => {
         /(?:<[^>^/]*>)?\s*(?:<\/[^>]*>)?<\/(div|p|table|thead|tbody|tr|td|th|caption|col|colgroup|ol|ul|li)/g,
         '</$1'
       );
-
-    console.log(doc.body.innerHTML);
   } while (doc.body.innerHTML !== previous);
 
   return doc.body.innerHTML;


packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
```diff
@@ -47,9 +47,17 @@ export const sanitizeHTML = (html: string): string => {
...
+        '<$1$2$3><$4$5'
+      )
+      // Removes empty elements before the closing tag of the capture group
+      // <p><span> </span></p> -> <p></p>
+      .replace(
+        /(?:<[^>^/]*>)?\s*(?:<\/[^>]*>)?<\/(div|p|table|thead|tbody|tr|td|th|caption|col|colgroup|ol|ul|li)/g,
+        '</$1'
       );
+
+    console.log(doc.body.innerHTML);
```
chrishelgert, line 60: Could you pelase remove the `console.log`? :)

chrishelgert (commented):
I guess you need to change your commit message to be `fix` otherwise it will not trigger a new version through semantic release ;)

packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
aodhagan-cf: Have removed

z0al (approved):
I'm gonna pretend that I understand how that regex works. :shipit:

chrishelgert (approved):

PR #1377, opened 2023-03-20
fix(RTE): split up sanitize regex to prevent invalid markup

chrishelgert:
The previously introduced regex for cleaning up unwanted linebreaks and whitespaces had an issue with some inline tags and produced therefore invalid markup.
We now split up the regex into two replacements, this will fix the behaviour and the RTE no longer crashes the page.

### DIFF
--- a/packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
+++ b/packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
@@ -21,6 +21,8 @@ const stripMetaTags = (doc: Document): Document => {
   return doc;
 };
 
+type Replacer = (innerHTML: string) => string;
+
 // Attention: Order is important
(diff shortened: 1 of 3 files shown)


aodhagan-cf (approved):

packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
```diff
@@ -31,31 +33,50 @@ export const sanitizeHTML = (html: string): string => {
...
+    // remove whitespaces before the ending block element tag
+    (innerHTML) =>
+      innerHTML.replace(
+        /\s*<\/(div|p|table|thead|tbody|tr|td|th|caption|col|colgroup|ol|ul|li)/g,
+        '</$1'
+      ),
+  ];
+
+  // Call this replacer only if there are tables inside, as it can have a larger performance impact
+  if (doc.querySelector('table')) {
```
chrishelgert, line 60: Had an example where it took ~4.5s to run this (and it didn't change anything)

PR #1384, opened 2023-03-30
fix(rte): pasting tables from ms word or google docs freezes the page

chrishelgert:
Fixes the bug that pasting tables from ms word or google docs are freezing the page or crashing the browser.
Previously we had this regex `/(.*\s)?<div.*>\s*(<table(?:.|\s)*<\/table>)\s*<\/div>(.*\s)?/g`.

What causes the issue was this one `<div.*>` as it also allows to include a nested div like
```
<div className="a">
  <div className="b"></div>
  <table>...</table>
</div>
```
to be parsed to this
```
<div className="a">
<table>...</table>
```
which will be invalid markup.

So the adjusted regex `/<div[^>]*>\s*(<table(?:.|\s)*<\/table>)\s*<\/div>/g` will not allow nested tags and is therefore safer to use. Also the `(.*\s)?` capture groups are not necessary and the removal will speed things up.


2wce (approved):

packages/rich-text/src/plugins/PasteHTML/utils/sanitizeHTML.ts
```diff
@@ -59,13 +59,11 @@ export const sanitizeHTML = (html: string): string => {
...
     // remove div container from tables
-    // capture groups explained:
-    // 1. and 3. every content/linebreaks before and after the div container
-    // 2. the table inside the container, including content and linebreaks
     // The div container including attributes and possible linebreaks inside wil be removed
-    replacers.unshift((innerHtml) =>
-      innerHtml.replace(/(.*\s)?<div.*>\s*(<table(?:.|\s)*<\/table>)\s*<\/div>(.*\s)?/g, '$1$2$3')
-    );
+    replacers.unshift((innerHtml) => {
+      const result = innerHtml.replace(/<div[^>]*>\s*(<table(?:.|\s)*<\/table>)\s*<\/div>/g, '$1');
```
DanweDE, line 64: @chrishelgert :see_no_evil: `(.|\s)*` can lead to [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html) prefer `[\s\S]*`

Also, wonder if that one's supposed to be greedy (matters when there are multiple tables)

PR #1385, opened 2023-04-03
fix: correct misleading action origin on paste event - rich text

MayaGillilanContentful:
Since we can't (without a major reworking) differentiate between shortcut and viewport action origins in this case, this adds a basically "could be either" origin. Even though it doesn't add any information atm, it does clear things up on the tracking side, and if we ever for some reason add pasting to the toolbar, it will be differentiated from there.


chrishelgert (approved):
Do we need to adapt a segment schema before we ship it?

MayaGillilanContentful:
> Do we need to adapt a segment schema before we ship it?

Nope! It's only changing an action name which is not limited to a certain set of strings in schema.

PR #1391, opened 2023-04-05
feat: upgrade cypress [TOL-1036]

2wce:
This upgrades cypress to the latest version & fixes the issues that arise from it including upgrading webpack


cypress.config.ts
```diff
@@ -0,0 +1,50 @@
+import { defineConfig } from 'cypress';
+
+// @ts-expect-error -- no types
+import webpackConfig from './webpack.config.js';
```
chrishelgert, line 4: we could also define it as `.ts` file 🤔 
What do you think about naming it, so we know that it is for cypress? or inline it inside the cypress.config.ts

cypress/component/mount.tsx
```diff
@@ -2,15 +2,18 @@ import React from 'react';
 
 import { GlobalStyles } from '@contentful/f36-core';
 import tokens from '@contentful/f36-tokens';
-import { mount as cyMount } from '@cypress/react';
-import { MountOptions } from '@cypress/react/dist/mount';
+import { mount as cyMount, MountReturn } from '@cypress/react';
+import { MountOptions } from '@cypress/react';
```
chrishelgert, line 6: NIT: we could combine it with the previous `import` statement ;)

cypress/tsconfig.json
```diff
@@ -5,14 +5,14 @@
     "baseUrl": "../node_modules",
     "lib": ["es5", "dom"],
     "target": "ES6",
-    "types": ["cypress", "@types/testing-library__cypress"],
-    "resolveJsonModule": true,
+    "types": ["cypress", "@types/testing-library__cypress", "jest"],
```
chrishelgert, line 8: Do we still need the `jest` after we added the webpack.config?

chrishelgert (approved):
Some small comments, but looks great otherwise 🥇 
The tests are running a lot faster and more stable 🎉

PR #1400, opened 2023-04-19
fix: [RTE] embed entries in correct place when using hotkeys

MayaGillilanContentful:
Removing selection in https://github.com/contentful/field-editors/pull/1354 and https://github.com/contentful/field-editors/pull/1353 introduced a regression specifically for users of hotkey insertion. Focusing improvements in those PRs are retained.


YvesRijckaert:
@MayaGillilanContentful Can we add a test that covers this please? So we don't remove it again in the future!

MayaGillilanContentful:
> @MayaGillilanContentful Can we add a test that covers this please? So we don't remove it again in the future!

I thought about this, not sure if we can actually since we only have a mock slide-in on the field-editors testing which doesn't actually "exist" or trigger any of this `sdk.navigator` watching logic that Daniel implemented. Or do you have another idea on how we could test it?

PR #1412, opened 2023-05-09
chore(rich-text-app): update dependencies to be startable again []

chrishelgert:
Currently the rich-text-app cannot be started, update dependencies to get it working again on the short term.
For the long term it needs some more magic to be working well again. (Follow up ticket)

### DIFF
--- a/apps/rich-text-app/package.json
+++ b/apps/rich-text-app/package.json
@@ -3,11 +3,11 @@
   "version": "2.0.0",
   "private": true,
   "dependencies": {
-    "@contentful/app-sdk": "^4.3.0",
+    "@contentful/app-sdk": "^4.6.0",
     "@contentful/f36-components": "^4.0.27",
     "@contentful/f36-tokens": "^4.0.0",
-    "@contentful/field-editor-rich-text": "^2.0.0-next.45",
-    "@contentful/field-editor-single-line": "^0.15.6",
+    "@contentful/field-editor-rich-text": "^3.4.21",
+    "@contentful/field-editor-single-line": "^1.2.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   },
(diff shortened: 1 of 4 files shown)


aodhagan-cf (approved):

apps/rich-text-app/package.json
```diff
@@ -20,7 +20,7 @@
     "upload-ci": "contentful-app-scripts upload --ci --bundle-dir ./build  --organization-id $CONTENTFUL_ORG_ID --definition-id $CONTENTFUL_APP_DEF_ID --token $CONTENTFUL_ACCESS_TOKEN"
   },
   "eslintConfig": {
-    "extends": "react-app"
```
andipaetzold, line 23: We shouldn't use relative paths here. Otherwise, people cannot copy that folder without changing the package.json. The example should not depend on anything outside of its folder.

apps/rich-text-app/src/index.tsx
```diff
@@ -1,24 +1,24 @@
 import * as React from 'react';
 import { render } from 'react-dom';
-import { init, FieldExtensionSDK, locations } from '@contentful/app-sdk';
+import { FieldExtensionSDK, DialogExtensionSDK, init, locations } from '@contentful/app-sdk';
```
andipaetzold, line 3: As we are already changing things here: `FieldAppSDK` and `DialogAppSDK` are the new names for these types. The old types will be removed in App SDK v5

andipaetzold (changes requested):

apps/rich-text-app/src/index.tsx
chrishelgert: We don't have this types here yet ("@contentful/app-sdk": "^4.6.0") and I assume we have to do this upgrade then also for all the Editors?
So would vote to do this in a separate ticket

andipaetzold: True, I assumed you upgraded to latest

-> another PR

package.json
andipaetzold, line 1: As we don't have a dependency to the root anymore, we can also remove this, right?

chrishelgert: Sadly not, if I remove the changes here, eslint complains about a plugin conflict for `react` :(

andipaetzold: That feels like something is not configured correctly, then :/ 
As written before: the example should be able to live alone

PR #1413, opened 2023-05-11
feat: setup swc compilation [TOL-1070]

2wce:
Migrating from TSDX to SWC transpilation only 💪
Migrating from Docz to Storybook 📕
Supporting ESM only packages 🤖


cypress/e2e/DateEditor.spec.ts
```diff
@@ -79,15 +118,18 @@ describe('Date Editor', () => {
...
-
       selectors.getDateInput().should('be.visible').should('have.value', '');
       selectors
         .getTimeInput()
         .should('be.visible')
         .should('have.attr', 'date-time-type', '24')
         .should('have.attr', 'placeholder', '00:00')
         .should('have.value', '00:00');
-      selectors.getTimezoneInput().should('be.visible').should('have.value', '+00:00');
+      // TODO find a better way
```
chrishelgert, line 128: out of curiosity: why did we have to change the whole date behavior in this file?

cypress/e2e/MarkdownOrganizeLinks.spec.ts
```diff
@@ -1,16 +1,18 @@
+import { getIframe } from '../fixtures/utils';
+
 describe('Markdown Editor / Organize Links', () => {
   const selectors = {
     getInput: () => {
-      return cy.get('[data-test-id="markdown-textarea"] [contenteditable]');
+      return getIframe().findByTestId('markdown-textarea').find('.CodeMirror-code');
```
chrishelgert, line 6: on the other ones we still use contenteditable, whats the difference here for .CodeMirror-code?

cypress/support/commands.js
```diff
@@ -8,34 +8,42 @@
...
-import '@testing-library/cypress/add-commands';
-import { configure } from '@testing-library/cypress';
 import { register } from '@cypress/snapshot';
+import { configure } from '@testing-library/cypress';
+import '@testing-library/cypress/add-commands';
 
 register();
 configure({ testIdAttribute: 'data-test-id' });
 
+const getIframeWindow = () => {
```
chrishelgert, line 18: that's the same as in cypress/fixtures/utils, can't we reuse it?

packages/date/.swcrc
```diff
@@ -0,0 +1,18 @@
+{
```
chrishelgert, line 1: ❓ could the .swcrc config also be shared?
 having one base config and then import/export it here?

packages/json/tsconfig.json
```diff
@@ -3,6 +3,8 @@
   "include": ["src", "types"],
   "compilerOptions": {
     "module": "esnext",
+    "moduleResolution": "bundler",
```
chrishelgert, line 6: maybe we can also move this into the base tsconfig?

packages/url/package.json
```diff
@@ -14,10 +22,16 @@
...
-    "test:ci": "tsdx test --env=jsdom --ci",
+    "watch": "yarn concurrently \"yarn:watch:*\"",
+    "watch:cjs": "yarn build:cjs -w",
+    "watch:esm": "yarn build:esm -w",
+    "watch:types": "yarn build:types --watch",
+    "build": "yarn build:types && yarn build:cjs && yarn build:esm",
+    "build:types": "tsc --outDir dist/types --emitDeclarationOnly",
+    "build:cjs": "swc src -d dist/cjs -C module.type=commonjs",
+    "build:esm": "swc src -d dist/esm",
+    "test": "jest --env=jsdom --watch",
```
chrishelgert, line 33: as we set the env in test and test:ci and in every project, speaks something against moving this into the jest.config (or base jest config)

cypress/e2e/rich-text/RichTextEditor.Commands.spec.ts
```diff
@@ -0,0 +1,198 @@
+/* eslint-disable mocha/no-setup-in-describe */
+
+import { BLOCKS } from '@contentful/rich-text-types';
```
chrishelgert, line 3: Great think splitting the rich text test up 👍

cypress/e2e/DateEditor.spec.ts
aodhagan-cf: cy.clock doesn't work in iframe :(

aodhagan-cf: as a work around I adding some functions to handle the time stuff, not ideal but the best option I could come up with. Happy to hear suggestions is you have some

packages/date/.swcrc
aodhagan-cf: Same thing came up in @andipaetzold 's PR there isn't really a nice way today to share configs in SWC so we are leaving them up to the packages themselves

cypress/e2e/MarkdownOrganizeLinks.spec.ts
aodhagan-cf: Honestly not sure at this point, it was a long time since this change. I think they both should work. I vaguely remember having some problems with HTML attributes

package.json
```diff
@@ -158,7 +177,7 @@
     ]
   },
   "engines": {
-    "node": ">=14.17.0",
+    "node": ">=16",
```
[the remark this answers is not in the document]
aodhagan-cf: hmmmmmmmmmmm not sure, possibly not for the consumer of the packages but for those who want to extend field editors then yes I guess

packages/date/.swcrc
2wce: i've done the same now

cypress/support/commands.js
aodhagan-cf: Done

cypress/e2e/MarkdownOrganizeLinks.spec.ts
2wce: the test fails when we use contenteditable, the switch to code mirror code was the fix

packages/checkbox/jest.config.js
```diff
@@ -7,4 +7,5 @@ const packageName = package.name.split('@contentful/')[1];
 
 module.exports = {
   ...baseConfig(packageName),
+  transformIgnorePatterns: [`/node_modules/(?!nanoid)`],
```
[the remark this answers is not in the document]
2wce: moved to base config 👍🏼

packages/rich-text/stories/RichTextEditor.stories.tsx
```diff
@@ -0,0 +1,205 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+import * as React from 'react';
```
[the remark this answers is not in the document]
aodhagan-cf: can we please sort this out as a follow up? this PR is getting out of hand

packages/reference/src/assets/WrappedAssetCard/WrappedAssetLink.tsx
```diff
@@ -32,6 +32,7 @@ export interface WrappedAssetLinkProps {
 export const WrappedAssetLink = (props: WrappedAssetLinkProps) => {
   const { className, href, onEdit, onRemove, isDisabled } = props;
 
+  // @ts-expect-error
```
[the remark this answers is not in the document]
aodhagan-cf: Cool!

package.json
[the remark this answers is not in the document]
aodhagan-cf: Think about this more I don't think this is a breaking change, the packages themselves still work the same it just the way we build them. No props / interface changes needed to make this work on the client side, so not a breaking change IMO

PR #1414, opened 2023-05-11
refactor(rich-text): bump plate to v18

z0al:
A split from #1340 to focus only on v18 as v19+ requires an additional slate bump which I'd rather not do right away.

The significant change in v17 is the introduction of `PlateProvider`. Plate store is now accessible only below `PlateProvider` which challenged the current code structure but also gave us a chance to clean up some code.

Happy to hear any suggestions regarding the approach taken in this PR.


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -77,40 +52,32 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
...
-
   return (
     <SdkProvider sdk={props.sdk}>
       <ContentfulEditorIdProvider value={id}>
         <div className={styles.root} data-test-id="rich-text-editor">
-          <Plate
+          <PlateProvider
             id={id}
+            initialValue={value}
+            normalizeInitialValue={true}
```
aodhagan-cf, line 62: Okay so this replaces the useEffect that we used to have right?

packages/rich-text/src/helpers/callbacks.ts
```diff
@@ -0,0 +1,32 @@
...
+  // value.
+  let cache: unknown = null;
+
+  return debounce((document: unknown) => {
+    if (equal(document, cache)) {
+      return;
+    }
+
+    cache = document;
+    const doc = removeInternalMarks(
```
aodhagan-cf, line 23: Where did this logic live before?

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -39,36 +38,12 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
...
 
-  const onValueChanged = useOnValueChanged({
-    editorId: id,
-    handler: props.onChange,
-    skip: pendingExternalUpdate || isFirstRender,
-    onSkip: () => setPendingExternalUpdate(false),
-  });
+  const value = toSlateValue(props.value);
 
-  useEffect(() => {
```
aodhagan-cf, line 52: We now handle this useEffect in the `SyncEditorValue` function?

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -77,40 +52,32 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
     props.isToolbarHidden && styles.hiddenToolbar
   );
 
-  useEffect(() => {
```
aodhagan-cf, line 80: This is now done in the `toSlaveValue` hook?

packages/rich-text/src/SyncEditorValue.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * A void component that's responsible for keeping the editor
+ * state in sync with incoming changes (aka. external updates)
+ *
+ * This component is a hack to workaround the limitation of Plate v17+
+ * where we can no longer access the editor instance outside the Plate
+ * provider.
+ */
+export const SyncEditorValue = ({ incomingValue }: SyncEditorStateProps) => {
```
aodhagan-cf, line 50: `IncomingValueEditorSyncer` don't know if this sounds any better just throwing it out there

aodhagan-cf:
In my PR I also removed the piece of logic that deletes the first line. If you compare the behaviour to most other rich text editors they do not delete the line

packages/rich-text/src/RichTextEditor.tsx
aodhagan-cf: if so does this happen now more than just on first render?

packages/rich-text/src/helpers/callbacks.ts
z0al: It wasn't there. Looking at the existing implementation we used to have [useOnValueChanged](https://github.com/contentful/field-editors/blob/master/packages/rich-text/src/useOnValueChanged.ts) that did two things:
1. Skip selection changes. We don't want to trigger updates for those.
2. Avoid triggering `onChange` from incoming/external updates (via `onSkip` & `skip` which were passed in `RichTextEditor.tsx`)

To me, that's just a complicated way of saying: Don't trigger `onChange` if the actual JSON didn't change. Hence the change in this PR. If all things work out correctly then it should clean up the several `useState` and `useEffects` we had.

packages/rich-text/src/SyncEditorValue.tsx
z0al: Open to suggestions. I will rethink this when I get the PR green 😅

packages/rich-text/src/RichTextEditor.tsx
z0al: The two existing useEffects complements each other AFAIK so both are replaced by `SyncEditorValue`. The normalization bit is replaced by the `normalizeInitialValue` flag + the automatic normalization that happens when using `setEditorContent` function.

`normalizeInitialValue` didn't work for us in the past but I'm giving it a second shot. If it doesn't work then I'd repurpose the `SyncEditorState` to set the initial value as well. There is a trade-off for each solution.

packages/rich-text/src/RichTextEditor.tsx
z0al: Possibly, however, I'm not sure if the old logic _really_ triggered only on the first render. One of these two useEffects must have triggered to propagate `props.value` changes.

chrishelgert (approved):
LGTM 🏅

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -1,31 +1,31 @@
...
+import { SyncEditorValue } from './SyncEditorValue';
 import Toolbar from './Toolbar';
 import StickyToolbarWrapper from './Toolbar/components/StickyToolbarWrapper';
-import { useOnValueChanged } from './useOnValueChanged';
 
 type ConnectedProps = {
   sdk: FieldExtensionSDK;
   onAction?: RichTextTrackingActionHandler;
   minHeight?: string | number;
-  value?: object;
```
aodhagan-cf, line 28: @chrishelgert

PR #1422, opened 2023-05-31
feat: add resource block

veu:
Introduce new components to create and render new `embedded-resource-block` node.

Details:
* All code is copied from the entry block and adapted to the new block.
* Updating contentful-slatejs-adapter introduced some new challenges. The old version didn’t properly expose the types but the new version does and causes errors. My solution is to just typecast them as the types I’d need to use are either not exposed or not even defined (e.g. the `SlateNode` with additional data that is somehow represented as `Record<string, any>`).


packages/rich-text/src/plugins/links-tracking.tsx
```diff
@@ -1,17 +1,14 @@
-import React, { ComponentProps } from 'react';
+import { useCallback } from 'react';
 
 import { useContentfulEditorRef } from '../ContentfulEditorProvider';
-import { HyperlinkElementProps } from './Hyperlink/components/EntityHyperlink';
 
-export function withLinkTracking(Component: React.ComponentType<HyperlinkElementProps>) {
```
veu, line 6: HOC didn’t play well with TypeScript so the HOC became a hook.

msieroslawska (approved):
LGTM besides the component used in the plugin creation.

cypress/e2e/rich-text/RichTextEditor.spec.ts
```diff
@@ -1832,6 +1842,134 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
     }
   });
 
+  describe('Embedded Resource Blocks', () => {
```
aodhagan-cf, line 1845: I'm wondering could you please move these tests to a separate file? Something like `RichTextEditor.EmbeddedResourceBlocks.spect.ts`. Cypress can only split tests by files and not individual describes to we are currently extracting out some of these describes to their own files

packages/rich-text/src/useOnValueChanged.ts
```diff
@@ -31,8 +31,13 @@ export const useOnValueChanged = ({ editorId, handler, skip, onSkip }: OnValueCh
   const onChange = useMemo(
     () =>
       debounce((document: unknown) => {
-        const contentfulDocument = toContentfulDocument({ document, schema });
-        const cleanedDocument = removeInternalMarks(contentfulDocument);
+        const contentfulDocument = toContentfulDocument({
+          // eslint-disable-next-line -- parameter type is not exported @typescript-eslint/no-explicit-any
+          document: document as any,
```
aodhagan-cf, line 36: why does this need to be any now?

packages/rich-text/src/plugins/Hyperlink/components/EntityHyperlink.tsx
```diff
@@ -33,7 +34,7 @@ export function EntityHyperlink(props: HyperlinkElementProps) {
   const editor = useContentfulEditor();
   const sdk: FieldExtensionSDK = useSdkContext();
   const { target } = props.element.data;
-  const { onEntityFetchComplete } = props;
+  const { onEntityFetchComplete } = useLinkTracking();
```
aodhagan-cf, line 37: Nice

packages/rich-text/src/helpers/__tests__/removeInternalMarks.test.tsx
```diff
@@ -11,58 +11,58 @@ describe('internal mark', () => {
     const data = [
       {
         title: 'Paragraph mark is removed',
-        input: toContentfulDocument(
-          (
+        input: toContentfulDocument({
+          document: (
```
aodhagan-cf, line 15: Could you please explain why you now need to nest everything in a document?

packages/rich-text/src/helpers/getAllowedResourcesForNodeType.ts
```diff
@@ -0,0 +1,35 @@
...
+ * 3) the `validations` array may not have a `nodes` validation, and
+ * 4) the `nodes` validation may not validate the `nodeType`.
+ *
+ * @param {object} field
+ * @param {string} nodeType
+ * @returns {AllowedResource[]}
+ */
+export default function getAllowedResourcesForNodeType(field, nodeType): AllowedResource[] {
+  return flow(
+    (v) => find(v, 'nodes'),
```
aodhagan-cf, line 32: I know this is probably inspired by old code, but could you please rename `v` to something more meaningful? I think it's validations maybe?

packages/rich-text/src/plugins/EmbeddedResourceBlock/LinkedResourceBlock.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
```
aodhagan-cf, line 1: A lot of the code here is copied directly from the LinkedEntityBlock. Considering they are so similar could you please look into extracting out the shared logic please?

packages/rich-text/src/plugins/EmbeddedResourceBlock/Util.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+  children: [{ text: '' }],
+  isVoid: true,
+});
+
+// TODO: DRY up copied code from HR
+export function insertBlock(editor: PlateEditor, entity) {
+  if (!editor?.selection) return;
+
+  const linkedResourceBlock = createNode(entity);
```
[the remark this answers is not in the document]
aodhagan-cf: Would it be possible to create a shared Util.ts, which you can specify the differences? They is quite a bit of copied code between the 2

packages/rich-text/src/useOnValueChanged.ts
veu: I wrote about this in the description. The old build of `contentful-slatejs-adapter` didn’t properly expose its types so nobody ever checked that the types make sense. Updating to the new version of `contentful-slatejs-adapter` fixed the types but now all the errors pop up here. Fixing the types is out of scope of this change.
E.g. this code calls two functions from `contentful-slatejs-adapter` and passes the return value of the first to the second and somehow the types are incompatible.

packages/rich-text/src/plugins/Tracking/utils.ts
```diff
@@ -6,7 +6,8 @@ import { PlateEditor } from '../../internal/types';
 
 export function getCharacterCount(editor: PlateEditor) {
   const document = contentfulSlateJSAdapter.toContentfulDocument({
-    document: editor.children,
+    // eslint-disable-next-line -- parameter type is not exported @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
veu: `toContentfulDocument` expects a list of `SlateNode` but we pass a list of `Element`. The two are not compatible. It didn’t surface before because the types in `contentful-slatejs-adapter` weren’t working.

cypress/e2e/rich-text/RichTextEditor.spec.ts
veu: Done

packages/rich-text/src/plugins/EmbeddedResourceBlock/Util.ts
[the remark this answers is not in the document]
veu: Done

packages/rich-text/src/helpers/getAllowedResourcesForNodeType.ts
veu: Done

packages/rich-text/src/plugins/EmbeddedResourceBlock/LinkedResourceBlock.tsx
veu: Done

packages/rich-text/src/helpers/__tests__/removeInternalMarks.test.tsx
veu: It’s the format that `toContentfulDocument` expects according to its types which were unavailable in the old `contentful-slatejs-adapter` version.

veu: Here’s the error i get when using the old format with the new `contentful-slatejs-adapter` version.
image

aodhagan-cf: Okay cool thanks for letting me know

aodhagan-cf (approved):
Awesome thanks for making those adjustments, much appreciated 🥇

PR #1436, opened 2023-06-27
fix(rich-text): onChange callback is not triggered [TOL-1284]

z0al:
In some instances, the `onChange` callback for RT does not get triggered. One of which is updating the URL value of a hyperlink after a fresh load of a page (i.e., initial render).

There is one test I skipped that I will fix in a follow-up PR


aodhagan-cf (approved):
Awesome

PR #1438, opened 2023-06-28
feat: remove json extensions [TOL-1286]

2wce:
this temporarily removes JSON extensions to fix the bug caused when importing extensions from codemirror

### DIFF
--- a/packages/json/src/JsonEditorField.tsx
+++ b/packages/json/src/JsonEditorField.tsx
@@ -1,7 +1,5 @@
 import * as React from 'react';
 
-import { json } from '@codemirror/lang-json';
-import { EditorView } from '@codemirror/view';
 import tokens from '@contentful/f36-tokens';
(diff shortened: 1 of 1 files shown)


packages/json/src/JsonEditorField.tsx
```diff
@@ -61,7 +59,6 @@ export function JsonEditorField(props: JsonEditorFieldProps) {
         value={props.value}
         onChange={props.onChange}
         theme="light"
-        extensions={[json(), EditorView.lineWrapping]}
```
z0al, line 64: What are the consequences of this? Is it fair to expect support tickets regarding it?

2wce: we are going to test the difference this makes before merging, it looks more like a styling thing at first glance but once we confirm we'll know the potential impact

PR #1443, opened 2023-07-10
fix: skip set_selection onChange events

z0al:
Re-implemented the way we handle subscribing to `onChange` events. Previously, we used to skip `set_selection` events, but I mistakenly removed it [during the recent upgrade](https://github.com/contentful/field-editors/pull/1414).

This PR brings back the logic.


2wce (approved):
LGTM

chrishelgert (approved):
LGTM 👍

packages/rich-text/src/SyncEditorChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+      return;
+    }
+
+    lastIncomingValue.current = incomingValue;
+    setEditorValue(editor, incomingValue);
+  }, [editor, incomingValue]);
+};
+
+/**
+ * Attaches a custom onChange callback that
```
YvesRijckaert, line 33: This comment might not be complete? 😄

z0al: 🤦🏻‍♂️ I just noticed that hahah

packages/rich-text/src/SyncEditorChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+        });
+
+        if (operations.length === 0) {
+          return;
+        }
+
+        cb(document);
+      },
+    });
+  }, [editor, onChange, setEditorOnChange]);
```
YvesRijckaert, line 56: Does createOnChangeCallback need to be part of the dependencies here?

packages/rich-text/src/SyncEditorChanges.tsx
z0al: chatGPT failed me :) . damn it

packages/rich-text/src/SyncEditorChanges.tsx
z0al: No, this is an external dependency that doesn't change.

PR #1447, opened 2023-07-11
feat: add max height property to rich text editor [SPA-1200]

ChidinmaOrajiaku:
https://github.com/contentful/field-editors/assets/30434146/dff0d32c-94c7-46c8-877e-25f086eab97f

### DIFF
--- a/packages/rich-text/src/RichTextEditor.styles.ts
+++ b/packages/rich-text/src/RichTextEditor.styles.ts
@@ -15,7 +15,7 @@ export const styles = {
     fontSize: tokens.spacingM,
     fontFamily: tokens.fontStackPrimary,
     minHeight: '400px',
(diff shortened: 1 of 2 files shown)


packages/rich-text/src/RichTextEditor.styles.ts
```diff
@@ -15,6 +15,7 @@ export const styles = {
     fontSize: tokens.spacingM,
     fontFamily: tokens.fontStackPrimary,
     minHeight: '400px',
+    overflow: 'scroll',
```
z0al, line 18: - I think this should be `overflow-y` as we don't want horizontal scrolling. 
- Any reason for `scroll` instead of `auto`?

z0al (approved):
LGTM

PR #1452, opened 2023-07-24
fix: remove lodash-es to fix cjs issues [TOL-1346]

2wce:
@contentful/field-editor-shared is importing lodash-es which seems to cause some issues: https://discord.com/channels/1030176999471321129/1111350294434025613/1132070425254445186

In General:
ESM can import CJS
CJS can't import ESM

@contentful/field-editor-shared is published as CJS and "ESM". Therefore it should only import CJS dependencies. FieldConnector.ts is importing from lodash-es. Also, the package depends on lodash-es. Removing the dependency and avoid importing from lodash-es to have a proper CJS package

### DIFF
--- a/packages/single-line/src/SingleLineEditor.test.tsx
+++ b/packages/single-line/src/SingleLineEditor.test.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 
 import { createFakeFieldAPI, createFakeLocalesAPI } from '@contentful/field-editor-test-utils';
 import { cleanup, configure, fireEvent, render, waitFor } from '@testing-library/react';
-import identity from 'lodash/identity';
 
 import '@testing-library/jest-dom/extend-expect';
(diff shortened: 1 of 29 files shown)


packages/single-line/src/SingleLineEditor.test.tsx
```diff
@@ -11,13 +10,13 @@ configure({
...
 });
 
-jest.mock(
-  'lodash/throttle',
-  () => ({
-    default: identity,
-  }),
-  { virtual: true }
-);
+// jest.mock(
```
andipaetzold, line 13: Why not completely remove this code?

andipaetzold (commented):
Did a quick search and found some remaining usaged of `lodash-es`. Should we remove those as well to prevent future devs from using lodash-es?

image

Rest looks good

Should we merge this as a bugfix release rather than a feature release?

andipaetzold (approved):

PR #1463, opened 2023-08-10
feat: adding environment support in the WrappedEntryCard

Cyberxon:
in this PR we want to be able to correctly fetch an entry from a crn that contains environment information whether it is an implicit master (crn contains`/environments/master`) or any other non master environment (crn contains `/environments/staging` for example). We also would. like to show the environment information in the tooltip of the wrapped entry card as shown in the screenshot below: 

Screenshot 2023-08-10 at 10 54 38

### DIFF
--- a/packages/reference/src/resources/Cards/ResourceCard.spec.tsx
+++ b/packages/reference/src/resources/Cards/ResourceCard.spec.tsx
@@ -22,7 +22,7 @@ jest.mock('react-intersection-observer', () => ({
 
 // explicit master
 const resolvableEntryUrn = 'crn:contentful:::content:spaces/space-id/entries/linked-entry-urn';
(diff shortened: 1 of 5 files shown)


packages/reference/src/resources/Cards/ResourceCard.spec.tsx
```diff
@@ -19,7 +20,12 @@ jest.mock('react-intersection-observer', () => ({
   useInView: jest.fn().mockReturnValue({}),
 }));
 
+// explicit master
 const resolvableEntryUrn = 'crn:contentful:::content:spaces/space-id/entries/linked-entry-urn';
+const resolvableEntryUrnWithImplicitMaster =
```
veu, line 25: This should be called "explicit"

veu (approved):

PR #1468, opened 2023-08-21
feat(reference): enable sharing query cache between editors

z0al:
exports `EntityCacheProvider` (for the lack of a better name) that can used in a host application to wrap multiple reference editors resulting query client/cache sharing.

### DIFF
--- a/packages/reference/src/common/queryClient.tsx
+++ b/packages/reference/src/common/queryClient.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+
+import { QueryClient, useQuery as useRQ } from '@tanstack/react-query';
+
+/**
+ * A custom client context ensures zero conflict with host apps also using
+ * React Query.
+ */
+const clientContext = React.createContext<QueryClient | undefined>(undefined);
+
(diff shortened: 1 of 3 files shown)


packages/reference/src/common/queryClient.tsx
```diff
@@ -0,0 +1,47 @@
...
+          retry: false,
+        },
+      },
+    });
+  }, [client]);
+}
+
+// @ts-expect-error
+export const useQuery: typeof useRQ = (key, fn, opt) => {
+  return useRQ(key, fn, { ...opt, context: clientContext });
```
aodhagan-cf, line 36: I don't fully understand why you need to override the `useQuery` wondering could you please explain that to me?

aodhagan-cf (approved):
Looks great mate 🎉

packages/reference/src/common/queryClient.tsx
z0al: It's only to pass the custom `context` without it it will crash

chrishelgert (approved):
🥇

PR #1473, opened 2023-08-25
feat: switch to react-markdown, lazy loading [TOL-501]

aodhagan-cf:
- Replace markdown-to-jsx with react-markdown 
- Lazily load the package when needed (lazy loading the preview tab)

### DIFF
--- a/packages/markdown/src/index.tsx
+++ b/packages/markdown/src/index.tsx
@@ -1,6 +1,5 @@
 import './codemirrorImports';
 
 export { MarkdownEditorConnected as MarkdownEditor } from './MarkdownEditor';
-export { MarkdownPreview } from './components/MarkdownPreview';
 export { openMarkdownDialog } from './dialogs/openMarkdownDialog';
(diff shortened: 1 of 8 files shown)


z0al (approved):
Looks great 🙇🏻‍♂️

packages/markdown/src/index.tsx
```diff
@@ -1,6 +1,5 @@
 import './codemirrorImports';
 
 export { MarkdownEditorConnected as MarkdownEditor } from './MarkdownEditor';
-export { MarkdownPreview } from './components/MarkdownPreview';
```
z0al, line 4: Why?

aodhagan-cf: This was an attempt to remove it from the main bundle when lazy loading however it didn't work, want me to revert?

z0al: It's a breaking change technically so let's revert it and revisit in the follow up

aodhagan-cf: done!

PR #1482, opened 2023-09-05
feat: [TOL-1405] Replace throttling with debounce in field editors

colomolo:
Replacing throttling logic with debounce to prevent laggy updates of state inside of input fields.


aodhagan-cf (approved):
LGTM 🎉

One question does this introduce any strange behaviour for dropdown for example? Basically editors which do not have very fast changes anyways

colomolo:
> LGTM 🎉
> 
> One question does this introduce any strange behaviour for dropdown for example? Basically editors which do not have very fast changes anyways

It shouldn't as currently throttling is turned off for dropdowns, lists, etc. We do the same with debouncing.

PR #1485, opened 2023-09-07
feat: updating plate to 23, fixing breaking imports []

aodhagan-cf:
Todo
- [x] Fix e2e tests
- [x] Fix types


packages/rich-text/src/internal/types/editor.ts
```diff
@@ -2,14 +2,15 @@
...
-import * as p from '@udecode/plate-core';
+import * as p from '@udecode/plate-common';
 import * as s from 'slate';
 import * as sr from 'slate-react';
-import {
-  SelectionMoveOptions as SlateSelectionMoveOptions,
-  SelectionCollapseOptions as SlateSelectionCollapseOptions,
-} from 'slate/dist/transforms/selection';
-import { TextInsertTextOptions as SlateTextInsertTextOptions } from 'slate/dist/transforms/text';
+// TODO fix me
```
chrishelgert, line 8: What's with this one?

aodhagan-cf: The type seems to be no longer exported in the later version, still not sure how to handle this

packages/rich-text/package.json
```diff
@@ -49,25 +49,25 @@
...
-    "@udecode/plate-paragraph": "18.15.0",
-    "@udecode/plate-reset-node": "18.15.0",
-    "@udecode/plate-select": "18.15.0",
-    "@udecode/plate-serializer-docx": "18.15.0",
-    "@udecode/plate-serializer-html": "18.15.0",
-    "@udecode/plate-table": "18.15.0",
-    "@udecode/plate-trailing-block": "18.15.0",
+    "@udecode/plate-basic-marks": "23.7.0",
+    "@udecode/plate-break": "23.7.0",
+    "@udecode/plate-core": "23.6.0",
```
cdun, line 54: Is there a reason that this is a minor version below it's siblings (e.g. 23.6 vs 23.7)?

packages/rich-text/package.json
```diff
@@ -49,25 +49,25 @@
...
     "is-hotkey": "^0.2.0",
     "is-plain-obj": "^3.0.0",
     "react-popper": "^2.3.0",
-    "slate": "0.80.0",
+    "slate": "0.94.1",
     "slate-history": "0.66.0",
-    "slate-hyperscript": "0.67.0",
-    "slate-react": "0.80.0"
+    "slate-hyperscript": "0.77.0",
+    "slate-react": "0.98.3"
```
cdun, line 70: This will result in some issues - `@udecode/plate-core` depends on `slate >=0.94.0`. The conflict in slate dependencies between the field-editors packages and experience-components currently adds nearly 700kb to the web app's bundle size.
We should either try and align the versions between the two, or adopt proper semver specifiers if Slate's release process supports it.

packages/rich-text/src/internal/types/editor.ts
chrishelgert: import type { SelectionMoveOptions as SlateSelectionMoveOptions } from 'slate/dist/interfaces/transforms/selection'

packages/rich-text/package.json
aodhagan-cf: "23.7.0" doesn't exit for the core package for some reason unknown to me. I bumped to everything to the latest there was (at the time)

packages/rich-text/package.json
aodhagan-cf: Great callout! we will address this in a follow up

chrishelgert (approved):

PR #1497, opened 2023-09-20
refactor: [] Replace `react-sortable-hoc` with `dnd-kit` (TagEditor)

colomolo:
Replacing `react-sortable-hoc` in favor of `dnd-kit` for better compatibility with React in TagEditor component.

screencast 2023-09-26 16-03-13

### DIFF
--- a/packages/tags/src/TagsEditor.tsx
+++ b/packages/tags/src/TagsEditor.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useState } from 'react';
 
-import { Pill, TextInput } from '@contentful/f36-components';
-import { DragIcon } from '@contentful/f36-icons';
+import { DragHandle, Pill, TextInput } from '@contentful/f36-components';
 import tokens from '@contentful/f36-tokens';
 import { DndContext } from '@dnd-kit/core';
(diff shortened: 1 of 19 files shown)


packages/tags/src/TagsEditor.tsx
```diff
@@ -55,61 +57,83 @@ const styles = {
...
-    onClose={() => {
-      if (!props.isSortablePillDisabled) {
-        props.onRemove(props.index);
+const SortablePill = ({ id, label, index, disabled, onRemove }: SortablePillProps) => {
+  const { listeners, setNodeRef, setActivatorNodeRef, transform, transition } = useSortable({ id });
+  const style = {
+    transform: transform ? CSS.Transform.toString({ ...transform, scaleX: 1, scaleY: 1 }) : '',
+    transition,
+  };
+  return (
```
chrishelgert, line 90: Wondering if we could use https://f36.contentful.com/components/drag-handle?

colomolo: Good catch! I will try it.

aodhagan-cf (approved):
LGTM

chrishelgert (approved):

PR #1499, opened 2023-09-22
fix: [] Replace deprecated `*ExtensionSDK` types with `*AppSDK` types

colomolo:


ghepting (approved):

ghepting:
Nice! Do we need any additional/exhaustive integration tests around these changes?

colomolo:
> Nice! Do we need any additional/exhaustive integration tests around these changes?

It's just typings, should be fine.

andipaetzold:
`entry-app-collapsible` still uses `*ExtensionSDK`. Should that also be replaced?

andipaetzold (changes requested):
The new types were introduced in App SDK [4.17.1](https://github.com/contentful/ui-extensions-sdk/releases/tag/v4.17.1). We should bump the app-sdk version to that version

andipaetzold (approved):
Should we bump the version for this as we change the dependencies?

colomolo:
> Should we bump the version for this as we change the dependencies?

Yeah, makes sense.

Issue #1346, opened 2023-02-15
Rich Text Commands ignores linkContentType restriction


dwin0:
**Environment**: Contentful Live Site
**Component**: Rich Text Editor

The [Rich Text Commands](https://www.contentful.com/faq/rich-text/#what-are-rich-text-commands) suggests embeds of content types, which are restricted via RTE validation rules.

Here's my config for the Rich Text Editor:

```js
validations: [
      {
        enabledMarks: ['bold', 'italic'],
      },
      {
        enabledNodeTypes: [
          'heading-2',
          'heading-3',
          'heading-4',
          'heading-5',
          'embedded-entry-block',
          'hyperlink',
          'ordered-list',
          'unordered-list',
        ],
      },
      {
        nodes: {
          'embedded-entry-block': [
            {
              linkContentType: [ContentTypesEnum.customList],
              message: 'Only custom lists are allowed to be embedded.',
            },
            {
              ...minMaxRestriction(0, 5),
              message: '0 to 5 list are allowed',
            },
          ],
        },
      },
    ]
```

<img width="425" alt="Screenshot 2023-02-15 at 17 11 11" src="https://user-images.githubusercontent.com/11162196/219085229-65018f4f-4e5f-4e19-b786-ff4fcd3d3253.png">

PR #1503, opened 2023-09-26
fix: rich text commands listens to validations [TOL-1473]

YvesRijckaert:
- Fixes: https://github.com/contentful/field-editors/issues/1346
- Inline embedded entries are working again with rich text commands
- Add missing key props to ignore warnings
- Set colour when rich text command is active
- Add asset icon when no thumbnail available for assets

Todo
- [ ] Design review

### DIFF
--- a/packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
+++ b/packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
@@ -92,13 +92,13 @@ const CommandListItems = ({
 }) => {
   return (
     <>
-      {commandItems.map((command, i) => {
+      {commandItems.map((command) => {
         return 'group' in command ? (
-          <Group key={i} commandGroup={command} selectedItem={selectedItem} />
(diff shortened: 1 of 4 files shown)


packages/rich-text/src/plugins/CommandPalette/components/CommandList.tsx
```diff
@@ -83,13 +92,13 @@ const CommandListItems = ({
 }) => {
   return (
     <>
-      {commandItems.map((command) => {
+      {commandItems.map((command, i) => {
         return 'group' in command ? (
-          <Group commandGroup={command} selectedItem={selectedItem} />
+          <Group key={i} commandGroup={command} selectedItem={selectedItem} />
```
chrishelgert, line 97: as every command should be unique, can't we use the command name instead of the index as key?

YvesRijckaert: Yes, you're right! Fixed

chrishelgert (approved):
LGTM

PR #1504, opened 2023-09-26
chore: upgrade to node LTS [EXT-4805]

t-col:


package.json
```diff
@@ -48,7 +48,6 @@
     "@babel/preset-react": "7.13.13",
     "@commitlint/cli": "^12.0.1",
     "@commitlint/config-conventional": "^12.0.1",
-    "@contentful/contentful-extension-scripts": "^0.20.4",
```
t-col, line 51: unused and was causing installation issues due to incompatible peer dependencies

PR #1508, opened 2023-10-01
feat: [] Migrate reference editor to dnd-kit

colomolo:
For reference editor we have to create an ids array to help `dnd-kit` sort items. Still not happy with overall solution. Feels like we need more deep/centralized refactoring with singleton, possibly reusing it in tags editor as well.


packages/reference/src/common/MultipleReferenceEditor.tsx
```diff
@@ -61,30 +62,32 @@ function Editor(props: EditorProps) {
...
 
-  const onSortStart: SortStartHandler = useCallback((_, event) => {
-    if (event instanceof MouseEvent) {
-      document.body.classList.add('grabbing');
-    }
-    event.preventDefault();
+  const { rearrangeSortIDs } = useSortIDs(items);
+
+  const onSortStart = useCallback(() => {
+    document.body.classList.add('grabbing');
```
YvesRijckaert, line 68: Why did we remove `if (event instanceof MouseEvent) {`
I guess sorting can also be done with the keyboard, maybe then we don't want to have the `cursor: grabbing` ?

packages/reference/package.json
```diff
@@ -51,7 +50,9 @@
     "moment": "^2.20.0",
     "p-queue": "^4.0.0",
     "react-intersection-observer": "9.4.0",
-    "react-sortable-hoc": "^2.0.0",
+    "@dnd-kit/core": "^6.0.8",
```
chrishelgert, line 53: NIT: package.json is sorted alphabetical normally ;)

packages/reference/src/common/SortableLinkList.tsx
```diff
@@ -1,23 +1,27 @@
...
 import { ReferenceEditorProps } from './ReferenceEditor';
 
 const styles = {
   container: css({
     position: 'relative',
   }),
   item: css({
+    position: 'relative',
     marginBottom: tokens.spacingM,
-    zIndex: tokens.zIndexModal, // setting this to an index above 99 fixes dragged item disappearing issue. Should not be higher than 100 so it does not overlap the asset modal.
```
chrishelgert, line 20: should we keep this comment? can maybe prevent bugs in the future :)

colomolo: Not sure how it was working previously or maybe it was, but with react-sortable-hoc. It hasn't had `position: relative;`. Currently I'm setting it only on drag.

packages/reference/src/utils/useSortIds.ts
```diff
@@ -0,0 +1,26 @@
...
+  const sortIDs = React.useRef<{ id: string }[]>([]);
+
+  React.useEffect(() => {
+    if (items.length !== sortIDs.current.length) {
+      sortIDs.current = items.map((item, index) => ({
+        id:
+          item.sys.type === 'ResourceLink' ? `${item.sys.urn}-${index}` : `${item.sys.id}-${index}`,
+      }));
+    }
+  }, [items]);
```
YvesRijckaert, line 19:
```suggestion
  React.useEffect(() => {
    if (items.length !== sortIDs.current.length) {
      sortIDs.current = items.map((item, index) => {
        const { type, urn, id } = item.sys;
        return {
          id: type === 'ResourceLink' ? `${urn}-${index}` : `${id}-${index}`,
        };
      });
    }
  }, [items]);
```

YvesRijckaert (approved):
LGTM 👍

packages/reference/src/common/SortableLinkList.tsx
```diff
@@ -26,55 +30,118 @@ type SortableContainerChildProps<IType> = Pick<
...
+  children,
+}: SortableLinkProps<T>) => {
+  const { listeners, setNodeRef, setActivatorNodeRef, transform, transition, isDragging } =
+    useSortable({ id });
+  const style = {
+    transform: transform ? CSS.Transform.toString({ ...transform, scaleX: 1, scaleY: 1 }) : '',
+    transition,
+  };
+
+  const DragHandle = (props: { drag: React.ReactElement }) => {
```
chrishelgert, line 70: Can't we use `https://docs.dndkit.com/api-documentation/draggable` and extract it into it's own component?
Currently it will be recreated through each re-rendering

chrishelgert: Also on the useSortable documentation:
```
function SortableItem(props) {
  const {arguments, listeners, setNodeRef} = useDraggable({
    id: props.id,
  });
  
  return (
    <li ref={setNodeRef}>
      {/* ... */}
      <button {...listeners} {...arguments}>Drag handle</button>
    </li>
  );
}
```

packages/reference/src/common/MultipleReferenceEditor.tsx
colomolo: `DragStartEvent` do not extend MouseEvent

packages/reference/src/common/SortableLinkList.tsx
colomolo: No, because we use `useSortable` both in handle (see [`setActivatorNodeRef`](https://docs.dndkit.com/presets/sortable/usesortable#activator)) and in card itself.

Or maybe I'm missing something here?

colomolo: Yeah, I get it, but it will be tricky to pass props there as DragHandle (that comes from EntryCard component) accepts only function with props: { drag }. How about just using `useCallback`?

chrishelgert: Yeah or `useMemo`, also fine if one of this works then 👍

aodhagan-cf:
Not sure if this is a problem but I tested out your branch in storybook locally and it behaves differently then prods storybook

I guess all of these entries have the same id so it doesn't do the move, just something consider I guess

https://github.com/contentful/field-editors/assets/109096340/19374d1d-c37b-4cb0-87a2-671635c90092

packages/reference/src/resources/MultipleResourceReferenceEditor.tsx
```diff
@@ -34,8 +35,8 @@ type EditorProps = ReferenceEditorProps &
 function ResourceEditor(props: EditorProps) {
   const { setValue, items, apiUrl } = props;
 
-  const onSortStart: SortStartHandler = useCallback((_, event) => event.preventDefault(), []);
-  const onSortEnd: SortEndHandler = useCallback(
+  const onSortStart = useCallback(() => noop(), []);
```
aodhagan-cf, line 38: Seems a bit of a waste having an empty function in a useCallback, maybe just `noop` is fine

packages/reference/src/common/SortableLinkList.tsx
```diff
@@ -26,55 +30,128 @@ type SortableContainerChildProps<IType> = Pick<
...
+  const { listeners, setNodeRef, setActivatorNodeRef, transform, transition, isDragging } =
+    useSortable({ id });
+  const style = {
+    transform: transform ? CSS.Transform.toString({ ...transform, scaleX: 1, scaleY: 1 }) : '',
+    transition,
+  };
+
+  const DragHandle = React.useCallback(
+    (props: { drag: React.ReactElement }) => {
+      const SortableDragHandle = () => props.drag;
```
aodhagan-cf, line 72: Could this simply be `children`?

aodhagan-cf (approved):
Looks great 🎉 just double check the video I attached and see if that's a problem

PR #1511, opened 2023-10-04
test: migrate date editor to component tests

MayaGillilanContentful:


cypress/component/date/DateEditor.spec.tsx
```diff
@@ -112,16 +88,18 @@ describe('Date Editor', () => {
...
+      mount(<DateEditor field={fieldSdk} isInitiallyDisabled={false} />);
 
       selectors.getDateInput().should('have.value', '03 Jan 2018');
       selectors.getTimeInput().should('have.value', '05:53');
       selectors.getTimezoneInput().should('have.value', '+03:00');
     });
 
     it('should render date, time (24 format) and timezone inputs by default', () => {
-      openPage();
+      const [fieldSdk] = createFakeFieldAPI();
```
YvesRijckaert, line 100: To avoid repetition, maybe we can extract this into a function?

```suggestion
      function setupDateEditor() {
          const [fieldSdk] = createFakeFieldAPI();
          mount(<DateEditor field={fieldSdk} isInitiallyDisabled={false} />);
      }
```

MayaGillilanContentful: Sure, could do that with some args.

cypress/component/date/DatepickerInput.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    mount(<DatepickerInput value={date} onChange={cy.stub()} />);
+  };
+
+  describe('renders correct date, when the system UTC offset is +02:00 and the one in the props.value is', () => {
+    it('+04:00', () => {
+      global.Date = timezonedDate.makeConstructor(120);
+      renderDatepicker('2022-09-22T00:00+04:00');
+      cy.findByTestId('cf-ui-datepicker-input').should('have.value', '22 Sep 2022');
+      // expect(screen.getByTestId('cf-ui-datepicker-input')).toHaveValue('22 Sep 2022');
```
YvesRijckaert, line 32: should this be uncommented?

cypress/component/date/DateEditor.spec.tsx
MayaGillilanContentful: Although we do need to use the `fieldSdk` directly in some tests. Do you see any readability issues with returning it from the setup func?

cypress/component/date/DatepickerInput.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    it('+10:00', () => {
+      global.Date = timezonedDate.makeConstructor(120);
+      renderDatepicker('2022-09-22T00:00+10:00');
+      cy.findByTestId('cf-ui-datepicker-input').should('have.value', '22 Sep 2022');
+      // expect(screen.getByTestId('cf-ui-datepicker-input')).toHaveValue('22 Sep 2022');
```
YvesRijckaert, line 39: same here, uncommented I guess?

cypress/component/date/DatepickerInput.spec.tsx
MayaGillilanContentful: Woops, no, those are jest remnants I forgot to remove

cypress/component/date/DateEditor.spec.tsx
```diff
@@ -15,93 +21,63 @@ const MONTHS = {
...
+const now = new Date();
+
+const selectors = {
+  getDateInput: () => cy.findByTestId('cf-ui-datepicker-input'),
+  getCalendarTrigger: () => cy.findByTestId('cf-ui-datepicker-button'),
+  getTimeInput: () => cy.findByTestId('time-input'),
+  getTimezoneInput: () => cy.findByTestId('timezone-input'),
+  getClearBtn: () => cy.findByTestId('date-clear'),
+  getCalendar: () => cy.get('.rdp'),
+  getCalendarMonth: () => cy.get('.rdp select[aria-label="Month: "]'),
```
chrishelgert, line 33: Wondering if we could use `cy.findByLabel('Month: ')` instead 🤔 
Same for year
Then we would also test the a11y :)

cypress/component/date/DateEditor.spec.tsx
MayaGillilanContentful: So it would look like this

```const setupDateEditor = (
  initialValue?: string,
  initiallyDisabled = false,
  parameters?: Parameters
) => {
  const [fieldSdk] = createFakeFieldAPI(undefined, initialValue);
  mount(
    <DateEditor field={fieldSdk} isInitiallyDisabled={initiallyDisabled} parameters={parameters} />
  );
  return fieldSdk;
};```

YvesRijckaert: I think that looks good 👍

YvesRijckaert (approved):
Thank you! looks great :)

chrishelgert (approved):
LGTM

PR #1512, opened 2023-10-05
fix: fixing broken hyperlink display []

aodhagan-cf:
Fixing broken hyperlink display due to https://github.com/contentful/forma-36/commit/a331cc1aff7ee44b54f40d849cc320f61031432e


YvesRijckaert (approved):
Thank you :)

PR #1516, opened 2023-10-10
refactor: Sync inline nodes code with the structure of block nodes [DANTE-1167]

msieroslawska:
This refactoring PR is the **first part** of introducing embedded inline cross-space links and is supposed to make the introduction easier. Once we have a similar code structure for block and inline nodes, the effort of adding inline resource links should be lower.

There are no functionality changes in the scope of this PR, only renaming components and moving the code around.


packages/rich-text/src/plugins/shared/LinkedInlineWrapper.tsx
```diff
@@ -0,0 +1,66 @@
...
+  };
+}>;
+
+export function LinkedInlineWrapper({
+  attributes,
+  card,
+  children,
+  element,
+}: LinkedInlineWrapperProps) {
+  const link = element.data.target;
```
Silhoue, line 44: Could we only pass the `target` down if that's all we need? Or will we need more later on?

Silhoue (approved):
Great that you're aiming to split it into smaller PRs 👍

packages/rich-text/src/plugins/shared/LinkedInlineWrapper.tsx
```diff
@@ -0,0 +1,66 @@
...
+    display: 'inline-block',
+    margin: `0 ${tokens.spacing2Xs}`,
+    fontSize: 'inherit',
+    span: {
+      userSelect: 'none',
+    },
+  }),
+};
+
+const isResourceLink = (link: EntryLink | ResourceLink): link is ResourceLink =>
```
chrishelgert, line 25: I think we have this one also in the LinkedBlockWrapper, can we maybe extract it to an util?
Or maybe even this snipped `isResourceLink(link) ? link.sys.urn : link.sys.id` as a function

chrishelgert (approved):
Nice cleanup 👍

PR #1517, opened 2023-10-11
fix(markdown): improve handling for external updates []

chrishelgert:
Prevents the loss of focus during typing

Previously the whole MarkdownEditor was completely re-created due to the key `externalReset`.
Due to that, the editor loses the focus and also the scroll position.

### DIFF
--- a/packages/markdown/src/MarkdownEditor.tsx
+++ b/packages/markdown/src/MarkdownEditor.tsx
@@ -47,11 +47,12 @@ export interface MarkdownEditorProps {
 export function MarkdownEditor(
   props: MarkdownEditorProps & {
     disabled: boolean;
-    initialValue: string | null | undefined;
+    value: string | null | undefined;
     saveValueToSDK: Function;
   }
 ) {
-  const [currentValue, setCurrentValue] = React.useState<string>(props.initialValue ?? '');
+  const initialExternalValue = React.useRef(props.value);
+  const [currentValue, setCurrentValue] = React.useState<string>(props.value ?? '');
   const [selectedTab, setSelectedTab] = React.useState<MarkdownTab>('editor');
   const [editor, setEditor] = React.useState<InitializedEditorType | null>(null);
   const [canUploadAssets, setCanUploadAssets] = React.useState<boolean>(false);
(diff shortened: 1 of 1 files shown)


packages/markdown/src/MarkdownEditor.tsx
```diff
@@ -152,18 +161,9 @@ export function MarkdownEditorConnected(props: MarkdownEditorProps) {
       field={props.sdk.field}
       isInitiallyDisabled={props.isInitiallyDisabled}
     >
-      {({ value, disabled, setValue, externalReset }) => {
```
DanweDE, line 155: Regarding [`externalReset logic](https://github.com/contentful/field-editors/blob/3ae9c46f02565c17a9323b1d319950bbb5adc900/packages/_shared/src/FieldConnector.ts#L116): This might be important to reset the editor whenever a remote change (other user, CMA change) happens.

If you remove it, the editor might not properly pick up on remote changes, or there could be certain glitches, leading to the remote change being overwritten again. Not quite sure.

The question is though: If the user experiences a focus loss issue, but there's no other user editing remotely, why is the external reset triggered?

chrishelgert: Through the auto-save of the editor if you type during it it comes to this glitch

With these changes it works and I could also tried the update magic with two different tabs and it looks also fine

chrishelgert: On other Editors (RichText, SingleLine) we also don't have this `externalReset`
RichTextEditor uses basically the same approach as this one

YvesRijckaert: @chrishelgert let me know if you want to test two different users making edits in the same markdown field

chrishelgert: @DanweDE Did some small adjustments and tested it together with @YvesRijckaert 
It works if we both work on the same editor correctly

packages/markdown/src/MarkdownEditor.tsx
```diff
@@ -47,11 +47,13 @@ export interface MarkdownEditorProps {
...
   props: MarkdownEditorProps & {
     disabled: boolean;
-    initialValue: string | null | undefined;
+    value: string | null | undefined;
     saveValueToSDK: Function;
+    externalReset?: number;
   }
 ) {
-  const [currentValue, setCurrentValue] = React.useState<string>(props.initialValue ?? '');
+  const initialExternalReset = React.useRef(props.externalReset);
```
DanweDE, line 55: Nit: as you keep this up-to-date,
```suggestion
  const prevExternalReset = React.useRef(props.externalReset);
```

DanweDE (approved):
LGTM :+1:

PR #1519, opened 2023-10-18
feat: add resource hyperlink node type

veu:
Add new `resource-hyperlink` node type rendering and extend the hyperlink modal with the new option.

### Unrelated fixes
* Fix error when entity selection modal was closed without selecting an entity.
* Prevent hyperlink modal being submitted with wrong entity.
* Align _Remove selection_ link with _Link target_ label


packages/rich-text/src/plugins/Hyperlink/useResourceEntityInfo.ts
msieroslawska, line 1: Good idea with this hook 👍

msieroslawska (approved):
Works like a charm! 🚀

chrishelgert (approved):
LGTM 👍

PR #1523, opened 2023-10-24
feat: increase space for space name

veu:
Double the space for the space name in resource link cards.

Before
image

After
image


luizfonseca (approved):
100% improvement

PR #1526, opened 2023-10-27
feat: moving the json editor to esm only []

aodhagan-cf:
Migrating the JSON editor to ESM-only


packages/json/jest.config.js
```diff
@@ -1,10 +1,9 @@
 /* eslint-disable */
+const baseConfig = require('../../baseESMJestConfig.js');
 
-const baseConfig = require('../../baseJestConfig');
+const packageJSON = require('./package.json');
+const packageName = packageJSON.name.split('@contentful/')[1];
```
chrishelgert, line 5:
```suggestion
const packageName = packageJSON.name.replace('@contentful/', '');
```

packages/json/package.json
```diff
@@ -1,17 +1,9 @@
 {
   "name": "@contentful/field-editor-json",
   "version": "3.3.6",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
+  "type": "module",
+  "main": "dist/esm/index.js",
   "types": "dist/types/index.d.ts",
-  "exports": {
```
chrishelgert, line 7: Do we need them or not?
I saw some post where it was mentioned that it was required for pure esm
other post are not mentioning them

packages/json/src/JsonEditorField.tsx
```diff
@@ -4,10 +4,10 @@ import { json } from '@codemirror/lang-json';
 import { indentUnit } from '@codemirror/language';
 import { EditorView } from '@codemirror/view';
 import tokens from '@contentful/f36-tokens';
-import CodeMirror from '@uiw/react-codemirror';
+import CodeMirror from '@uiw/react-codemirror/esm/index.js';
```
chrishelgert, line 7: ❓ why do we need to import from `esm`
Shouldn't be automatically detected?

packages/json/package.json
aodhagan-cf: @andipaetzold going to tag you on this comment if that's okay? I don't think we need it, I've tested this in the web app and it works, but have also seen some articles recommending to have it in

andipaetzold: No, we don't need it

PR #1537, opened 2023-11-07
fix: location null lat/lon issue []

aodhagan-cf:
Before (when lat and lon are both null):
Screenshot 2023-11-07 at 12 15 55

After (when lat and lon are both null):
Screenshot 2023-11-07 at 12 15 26

- If either lat or lon is set but not the other than setting the other to 0
- If both lat or lon are not set then setting localValue to undefined to remove disabled state
- If both lat and lon are set then should show both

### DIFF
--- a/packages/location/src/LocationEditor.tsx
+++ b/packages/location/src/LocationEditor.tsx
@@ -66,12 +66,13 @@ export class LocationEditor extends React.Component<
     super(props);
 
     this.state = {
-      localValue: props.value
-        ? {
-            lng: props.value.lon,
-            lat: props.value.lat,
-          }
-        : undefined,
+      localValue:
+        props?.value?.lon && props?.value.lat
+          ? {
+              lng: props.value.lon,
+              lat: props.value.lat,
+            }
+          : undefined,
       mapsObject: null,
     };
   }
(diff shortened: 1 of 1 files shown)


packages/location/src/LocationEditor.tsx
```diff
@@ -168,16 +171,18 @@ export function LocationEditorConnected(props: LocationEditorConnectedProps) {
...
-            // on external change reset component completely and init with initial value again
-            key={`location-editor-${externalReset}`}
-            value={value}
-            disabled={disabled}
-            setValue={setValue}
-            googleMapsKey={googleMapsKey}
-            selectedView={selectedView}
-            setSelectedView={setSelectedView}
-          />
+          <>
```
2wce, line 174: any reason why we are now returning a fragment instead of just the location editor?

2wce (approved):
LGTM

packages/location/src/LocationEditor.tsx
chrishelgert, line 174: why do we need a fragment here? :o

aodhagan-cf: hahaha damn, was testing stuff and it was left over. I've removed thanks for spotting!

aodhagan-cf: Have removed, sorry it was left over from testing stuff

PR #1542, opened 2023-11-13
feat: hyperlink edit popover modal

YvesRijckaert:
## Description
It is currently not possible to edit the text after you added a hyperlink. You can however use the arrow keys on your keyboard to edit the link. This is causing some bad UX. This PR will improve that by introducing a link popover element.

Screenshot 2023-11-14 at 16 05 14

## Acceptance criteria:
A mouse click brings the caret to the right position so the user can change the link text, but it also opens the popup menu underneath. This popup menu has a button to remove the link, copy the link (when they are not entity links), open the link a new tab and a button to open the edit hyperlink modal. In this modal the user can edit the hyperlink href.

### DIFF
--- a/cypress/e2e/rich-text/RichTextEditor.Tracking.spec.ts
+++ b/cypress/e2e/rich-text/RichTextEditor.Tracking.spec.ts
@@ -2,7 +2,7 @@
 
 import { BLOCKS, INLINES, MARKS } from '@contentful/rich-text-types';
 
-import { getIframe } from '../../fixtures/utils';
+import { getIframe, openEditLink } from '../../fixtures/utils';
 import { RichTextPage } from './RichTextPage';
 
 // the sticky toolbar gets in the way of some of the tests, therefore
@@ -592,9 +592,7 @@ describe('Rich Text Editor - Tracking', { viewportHeight: 2000 }, () => {
 
           // Part 2:
           // Update hyperlink to entry link
-
-          richText.editor.findByTestId('cf-ui-text-link').click({ force: true });
-
+          openEditLink();
           form.linkType.select('entry-hyperlink');
           form.linkEntityTarget.click();
           form.submit.click();
@@ -609,9 +607,7 @@ describe('Rich Text Editor - Tracking', { viewportHeight: 2000 }, () => {
 
           // Part 3:
           // Update entry link to asset link
-
-          richText.editor.findByTestId('cf-ui-text-link').click({ force: true });
-
+          openEditLink();
           form.linkType.select('asset-hyperlink');
           form.linkEntityTarget.click();
           form.submit.click();
@@ -629,9 +625,7 @@ describe('Rich Text Editor - Tracking', { viewportHeight: 2000 }, () => {
 
           // Part 4:
           // Update asset link to hyperlink
-
-          richText.editor.findByTestId('cf-ui-text-link').click({ force: true });
-
+          openEditLink();
           form.linkType.select('hyperlink');
           form.linkTarget.type('https://zombo.com');
           form.submit.click();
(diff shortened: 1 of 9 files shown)


packages/rich-text/src/plugins/Hyperlink/components/EntityHyperlink.tsx
```diff
@@ -1,15 +1,17 @@
 import * as React from 'react';
 
 import { FieldAppSDK, Link } from '@contentful/app-sdk';
-import { Tooltip, TextLink } from '@contentful/f36-components';
+import { Text } from '@contentful/f36-components';
+import { Path } from 'slate';
```
chrishelgert, line 5: should we import this from src/internal/queries?
because we wrap there `plate` and `slate`

chrishelgert: (`isChildPath`)

packages/rich-text/src/plugins/Hyperlink/components/EntityHyperlink.tsx
```diff
@@ -33,45 +35,45 @@ export type HyperlinkElementProps = {
...
+    <LinkPopover
+      isLinkFocused={isLinkFocused}
+      handleEditLink={() => handleEditLink(editor, sdk, pathToElement)}
+      handleRemoveLink={() => handleRemoveLink(editor)}
+      popoverText={popoverText}
     >
-      <TextLink
-        as="a"
-        onClick={handleClick}
+      <Text
```
chrishelgert, line 67: Can't we use `TextLink` with `as="span"`? Then we would not need to re-implement the styles

YvesRijckaert: No you can only set it as `span` or `button`

packages/rich-text/src/plugins/Hyperlink/components/LinkPopover.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </Tooltip>
+        <Tooltip placement="bottom" content="Remove link" usePortal>
+          <IconButton
+            onClick={handleRemoveLink}
+            className={styles.iconButton}
+            size="small"
+            variant="transparent"
+            aria-label="Remove link"
+            icon={
+              <svg
```
chrishelgert, line 67: Is there no forma icon we can use?

packages/rich-text/src/plugins/Hyperlink/components/UrlHyperlink.tsx
```diff
@@ -31,36 +33,33 @@ type HyperlinkElementProps = {
 export function UrlHyperlink(props: HyperlinkElementProps) {
   const editor = useContentfulEditor();
   const sdk: FieldAppSDK = useSdkContext();
-  const { uri } = props.element.data;
+  const focus = editor.selection?.focus;
```
chrishelgert, line 36: Maybe not for this MR, but we have some duplication in the Hyperlink components, maybe we can have a base component/hook to share the logic and only provide the specifics to them

YvesRijckaert: I'll create a ticket so we can work on that (maybe during cooldown)

cypress/e2e/rich-text/RichTextEditor.Tracking.spec.ts
```diff
@@ -610,8 +608,9 @@ describe('Rich Text Editor - Tracking', { viewportHeight: 2000 }, () => {
           // Part 3:
           // Update entry link to asset link
 
-          richText.editor.findByTestId('cf-ui-text-link').click({ force: true });
+          richText.editor.findByTestId('cf-ui-text-link');
```
chrishelgert, line 611: any expections/assertions here? same for Line 632

YvesRijckaert: ahh missed those, we can remove them 👍

chrishelgert (approved):

PR #1544, opened 2023-11-17
fix: rich text link copy to google docs [TOL-1647]

YvesRijckaert:
## Description
After merging https://github.com/contentful/field-editors/pull/1542 , copy pasting rich text content with links behaved differently. Before, a link would get copied over successfully. Afterwards, a link would not get copied anymore and it would include the link separately. This PR fixes that.

Screenshot 2023-11-17 at 12 39 03

Ticket: https://contentful.atlassian.net/browse/TOL-1647

## How?
I render the edit link popover component with a portal. This introduces some issues with positioning. First of all, we have to now check when the popover component should be displayed. It will only display when the editor is focussed and the caret is inside a link OR when the mouse interacts with the popover.
I've also changed the Text to TextLink in the rich text editor, which will make sure that when an editor copies over the content, it will be copied over as a link.

### DIFF
--- a/packages/rich-text/src/plugins/Hyperlink/components/LinkPopover.tsx
+++ b/packages/rich-text/src/plugins/Hyperlink/components/LinkPopover.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { Popover, IconButton, Tooltip, Flex } from '@contentful/f36-components';
 import { EditIcon, CopyIcon } from '@contentful/f36-icons';
 
+import { useFocused } from '../../../internal/hooks';
 import { styles } from './styles';
 
 type LinkPopoverProps = {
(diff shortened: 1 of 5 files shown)


packages/rich-text/src/plugins/Hyperlink/components/LinkPopover.tsx
```diff
@@ -21,67 +22,94 @@ export const LinkPopover = ({
...
-      >
-        {popoverText}
-        {handleCopyLink && (
-          <Tooltip placement="bottom" content="Copy link" usePortal>
+}: LinkPopoverProps) => {
+  const isEditorFocused = Slate.useFocused();
+  const popoverContent = React.useRef<HTMLDivElement | null>(null);
+  const [isPopoverContentClicked, setIsPopoverContentClicked] = React.useState(false);
+
+  const handleMouseDown = (event) => {
```
chrishelgert, line 30: NIT: move it into the useEffect so it will be created only once ;)

packages/rich-text/src/plugins/Hyperlink/components/LinkPopover.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 
 import { Popover, IconButton, Tooltip, Flex } from '@contentful/f36-components';
 import { EditIcon, CopyIcon } from '@contentful/f36-icons';
+import * as Slate from 'slate-react';
```
chrishelgert, line 5: see import from previously PR
I think we want to import slate and plate always from internal to make updates easier

chrishelgert (approved):
LGTM

PR #1547, opened 2023-11-20
chore: fix dependabot automerge

MayaGillilan:


.github/workflows/auto-merge.yml
```diff
@@ -14,13 +16,8 @@ jobs:
...
-  worker:
-    permissions:
-      contents: write
-      id-token: write
-    runs-on: ubuntu-latest
-    if: ${{github.actor == 'dependabot[bot]' && steps.dependabot-metadata.outputs.update-type != 'version-update:semver-major' && !contains(steps.dependabot-metadata.outputs.dependency-names, 'slate') && !contains(steps.dependabot-metadata.outputs.dependency-names, 'plate')}}
-    steps:
-      - uses: contentful/github-auto-merge@v1
+      - name: Auto-merge
+        if: ${{steps.dependabot-metadata.outputs.update-type != 'version-update:semver-major' && !contains(steps.dependabot-metadata.outputs.dependency-names, 'slate') && !contains(steps.dependabot-metadata.outputs.dependency-names, 'plate')}}
```
z0al, line 20: [dependabot.yml](https://github.com/contentful/field-editors/blob/master/.github/dependabot.yml) has native support for excluding dependencies. Why not use that to avoid having PRs in the first place?

MayaGillilan: We discussed this some before and the reason was so that we would still have a signal that a new plate/slate version is available, but it wasn't a strong preference

z0al (approved):

PR #1562, opened 2023-12-15
fix: use available CTs instead of only readable

SofiaMargariti:
The way reference field works, it will take the available CTs for the field - via validations - filter out the ones that user has no permissions to read and pass the remaining list to the Entity Search modal. If an empty array of CTs is provided Entity Search will render all CTs of the environment.

When a user has partial access to content types e.g. via tags these content types will not appear in the list of readable CTs and will result in an empty array to be passed to Entity Search which will then result in all available CTs to be listed overriding the field validations.

If instead we pass the available CTs then Entity Search will still respect the validations and is smart enough to filter out individual entities the user might not have permissions to read.

See the following example:
* fields allows only `Component:Text` entries to be linked via validations
* the user has access to some entries of the `Component:Text` via tags but cannot read any entry of that CT

Before all CTs would be listed:
Screenshot 2023-12-15 at 16 38 15
After, only the `Component:Text` is listed and only the entries that user can read:
Screenshot 2023-12-15 at 16 46 40

Related tickets:
* https://contentful.atlassian.net/browse/ZEND-4455
* https://contentful.atlassian.net/browse/ZEND-4432

### DIFF
--- a/packages/reference/src/common/useContentTypePermissions.ts
+++ b/packages/reference/src/common/useContentTypePermissions.ts
@@ -15,7 +15,6 @@ type ContentTypePermissionsProps = {
 
 type ContentTypePermissions = {
   creatableContentTypes: ContentType[];
-  readableContentTypes: ContentType[];
   availableContentTypes: ContentType[];
 };
 
@@ -46,7 +45,6 @@ export function useContentTypePermissions(
     return props.allContentTypes;
   }, [props.allContentTypes, props.validations.contentTypes, props.entityType]);
   const [creatableContentTypes, setCreatableContentTypes] = useState(availableContentTypes);
-  const [readableContentTypes, setReadableContentTypes] = useState(availableContentTypes);
   const { canPerformActionOnEntryOfType } = useAccessApi(props.sdk.access);
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


Chaoste (approved):
Nice fix. If you have time for this, I would appreciate a small comment next to the code, so we don't break this later again

z0al (changes requested):
Sorry to block but just want to make sure I understand this. it seems `readableContentTypes` are only ever used in the places you changed in this PR, so does that mean we should remove them from the reference editor props?

Also, any reason why not to fix this on the app level?

SofiaMargariti:
> Sorry to block but just want to make sure I understand this. it seems `readableContentTypes` are only ever used in the places you changed in this PR, so does that mean we should remove them from the reference editor props?
>

Good point, I didn't notice they are not used anymore, we could remove them in this case.

> Also, any reason why not to fix this on the app level?
Do you mean the web app here? This logic is all encapsulated inside the field-editor itself, it essentially determines the CTs to show based on field validations and permissions, they cannot be overridden from the outside atm and introducing new parameters seemed that it would complicate the logic even more. Do you have another suggestion?

z0al:
> Do you mean the web app here? This logic is all encapsulated inside the field-editor itself

You're right @SofiaMargariti , I was under the impression that the list is passed down from the host app but I was wrong. In that case, let's remove please the `readableContentTypes` related logic and 🚢

z0al (approved):

z0al (approved):

z0al (approved):

PR #1564, opened 2023-12-19
test: pass google maps key to LocationEditor component tests

SofiaMargariti:
LocationEditor component tests [keep failing](https://app.circleci.com/pipelines/github/contentful/field-editors/6441/workflows/e47211c0-e270-498e-910e-d93fdc0305ec/jobs/23234) due to googleMapsKey not being set properly. These tests were recently migrated to component tests. Before they were ran on top of storybook stories which were passing the key [correctly](https://github.com/contentful/field-editors/blob/ce6b53a046d6e79718c8e196556563d81f44df84/packages/location/stories/LocationEditor.stories.tsx#L36). This migration is most likely related to these failures, but I am still not sure why they passed [here](https://github.com/contentful/field-editors/pull/1550)

### DIFF
--- a/cypress/component/location/LocationEditor.spec.tsx
+++ b/cypress/component/location/LocationEditor.spec.tsx
@@ -7,9 +7,22 @@ import { mount } from '../mount';
 
 const renderLocationEditor = (isInitiallyDisabled = false) => {
   const [fieldSdk] = createFakeFieldAPI();
-  mount(<LocationEditor field={fieldSdk} isInitiallyDisabled={isInitiallyDisabled} />);
+  mount(
+    <LocationEditor
+      field={fieldSdk}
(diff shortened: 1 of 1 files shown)


Chaoste (approved):

z0al:
Thanks for taking a look @SofiaMargariti , but we shouldn't really need a key here, accessing GMaps directly would cause flakyness again which is why it was [mocked](https://github.com/contentful/field-editors/blob/ce6b53a046d6e79718c8e196556563d81f44df84/cypress/support/commands.js#L160-L166). However, it seems after [this PR](https://github.com/contentful/field-editors/pull/1550) the mocks are being applied a bit too late when you compare the e2e version vs component.

z0al (approved):
Thanks :bow:

PR #1568, opened 2024-01-04
refactor: hyperlink  extract common logic [TOL-1637]

YvesRijckaert:


chrishelgert (commented):
Wondering if we can also have a base `component`?
So we only need to provide the handleEditLink, handleRemoveLink and the toooltipContent and the children to it? 🤔

YvesRijckaert:
> Wondering if we can also have a base `component`? So we only need to provide the handleEditLink, handleRemoveLink and the toooltipContent and the children to it? 🤔

I tried something like that, but we are calling different hooks in all those components, the types are also different and there are other quirks (different data attributes, different forma  components (link vs text), etc) It wasn't very readable in the end so I left it the way it is now

packages/rich-text/src/plugins/Hyperlink/components/useHyperlinkCommon.tsx
```diff
@@ -0,0 +1,15 @@
+import { FieldAppSDK } from '@contentful/app-sdk';
+
+import { useContentfulEditor } from '../../../ContentfulEditorProvider';
+import { findNodePath, isChildPath } from '../../../internal/queries';
+import { useSdkContext } from '../../../SdkProvider';
+
+export function useHyperlinkCommon(element) {
+  const editor = useContentfulEditor();
```
aodhagan-cf, line 8: nit: I'm not sure I see the benefits for `editor` and `sdk` feels like adding a layer of complexity rather than removing. `isLinkFocused` and `pathToElement` make sense

chrishelgert (approved):

packages/rich-text/src/plugins/Hyperlink/components/useHyperlinkCommon.tsx
aodhagan-cf: Please feel free to ignore though just my opinion

YvesRijckaert: Hmm yes not sure, I think we need `editor` to calculate `pathToElement`

PR #1582, opened 2024-01-22
fix(rich-text): show the selected headline inside of lists

chrishelgert:
We always picked the first item to check if it's a headline.
Inside lists it would be deeper nested so we need a further iteration.


aodhagan-cf (approved):
LGTM one other test you could consider would be when you press enter, should it keep the type?

If you have a heading 1 in a list and you make a new list item, should it also be a heading 1?

chrishelgert:
> LGTM one other test you could consider would be when you press enter, should it keep the type?
> 
> If you have a heading 1 in a list and you make a new list item, should it also be a heading 1?

That should be still the same behaviour as before - it goes back to paragraph.

PR #1598, opened 2024-01-30
feat: add rich text undo redo btns [TOL-1038]

YvesRijckaert:
## Description
Adding undo and redo buttons to rich text editors
https://contentful.atlassian.net/browse/TOL-1038

Screenshot 2024-01-30 at 09 43 03

### Todo
- [ ] Add tests

### DIFF
--- a/cypress/component/rich-text/RichTextEditor.Tables.spec.ts
+++ b/cypress/component/rich-text/RichTextEditor.Tables.spec.ts
@@ -168,7 +168,7 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
         BLOCKS.HEADING_4,
         BLOCKS.HEADING_5,
         BLOCKS.HEADING_6,
-      ].forEach((type) => getDropdownItem(type).get('button').should('not.be.disabled'));
+      ].forEach((type) => getDropdownItem(type).should('not.be.disabled'));
     });
 
     describe('Inserting Tables', () => {
--- a/packages/rich-text/src/Toolbar/components/ButtonRedo.tsx
+++ b/packages/rich-text/src/Toolbar/components/ButtonRedo.tsx
@@ -1,0 +1,24 @@
+import React from 'react';
+
+import { Icon } from '@contentful/f36-components';
+
+import { useContentfulEditor } from '../../ContentfulEditorProvider';
+import { ToolbarButton } from '../../plugins/shared/ToolbarButton';
+
+export const ButtonRedo = () => {
+  const editor = useContentfulEditor();
+
+  return (
+    <ToolbarButton
+      title="Redo"
+      testId="redo-toolbar-button"
+      onClick={editor.redo}
+      isActive={false}
+      isDisabled={editor.history.redos.length === 0}
+    >
+      <Icon variant="secondary">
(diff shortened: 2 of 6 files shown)


chrishelgert (approved):
Nice

aodhagan-cf:
Is there any case where we could now have too many toolbar buttons? Also would be nice to have some component tests for this when the component test branch is merged (as a follow up)

aodhagan-cf (approved):

YvesRijckaert:
> Is there any case where we could now have too many toolbar buttons? Also would be nice to have some component tests for this when the component test branch is merged (as a follow up)

We already do (especially in Live Preview) 😞
I will follow up before merging this with Fidan and Fabian, but we have a ticket to make the toolbar responsive: https://contentful.atlassian.net/browse/TOL-1720

packages/rich-text/src/Toolbar/components/ButtonRedo.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <ToolbarButton
+      title="Redo"
+      testId="redo-toolbar-button"
+      onClick={editor.redo}
+      isActive={false}
+      isDisabled={editor.history.redos.length === 0}
+    >
+      <Icon variant="secondary">
+        <path d="M18.4,10.6C16.55,9 14.15,8 11.5,8C6.85,8 2.92,11.03 1.54,15.22L3.9,16C4.95,12.81 7.95,10.5 11.5,10.5C13.45,10.5 15.23,11.22 16.62,12.38L13,16H22V7L18.4,10.6Z" />
```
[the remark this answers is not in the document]
YvesRijckaert: I heard there is currently an initiative going on to change the underlying icon library. I will add a todo so we can change it to those in the future 👍

packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -136,6 +138,9 @@ const Toolbar = ({ isDisabled }: ToolbarProps) => {
       <div className={styles.formattingOptionsWrapper}>
         <ToolbarHeadingButton isDisabled={isDisabled || !canInsertBlocks} />
 
+        <span className={styles.divider} />
+        <ButtonUndo />
```
z0al, line 142: Any reasoning behind the placement? for markdown they are appended to the end for example 

Screenshot 2024-02-22 at 10 45 53

YvesRijckaert: I think it was like that in the ticket 🙈
But I will ask

YvesRijckaert: From Fidan: I don't mind having it at the front, undo/redo are usually on top in the software (design and text editing) I used to work on

cypress/component/rich-text/RichTextEditor.Tables.spec.ts
```diff
@@ -168,7 +168,7 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
         BLOCKS.HEADING_4,
         BLOCKS.HEADING_5,
         BLOCKS.HEADING_6,
-      ].forEach((type) => getDropdownItem(type).get('button').should('not.be.disabled'));
```
YvesRijckaert, line 171: @colomolo I removed the `.get('button')` here from this test, from what I saw the `getDropdownItem` already selects the button, or did I miss something?

PR #1599, opened 2024-02-01
fix: fixing custom app rich text dnd [TOL-1742]

aodhagan-cf:
image

Error being throw when dragging and dropping with custom rich text editor.
- Related to this history change & the new dndkit library

Todo
- [ ] Understand why change was introduced in the first place
- [ ] Remove or refactor as required

EDIT

Because we push a new undo as an empty array this condition doesn't short-circuit and the operations doesn't exist on the empty array. I'm not sure why this is only a problem for the App SDK and hasn't been happening the regular web app Rich Text 🤔

```
const lastOp =
lastBatch && lastBatch.operations[lastBatch.operations.length - 1]
```

https://github.com/ianstormtaylor/slate/blob/cd93871ae69721f7361e17e8b28feab004a9c855/packages/slate-history/src/with-history.ts#L72C5-L74C73

### DIFF
--- a/packages/rich-text/package.json
+++ b/packages/rich-text/package.json
@@ -67,7 +67,7 @@
     "is-plain-obj": "^3.0.0",
     "react-popper": "^2.3.0",
     "slate": "0.94.1",
-    "slate-history": "0.66.0",
+    "slate-history": "0.100.0",
     "slate-hyperscript": "0.77.0",
     "slate-react": "0.98.3"
   },
--- a/packages/rich-text/src/plugins/DragAndDrop/index.tsx
+++ b/packages/rich-text/src/plugins/DragAndDrop/index.tsx
@@ -45,11 +45,11 @@ export function createDragAndDropPlugin(): PlatePlugin {
             : false;
         });
 
-        if (!dropDisallowed) {
-          // Move the drop event to a new undo batch mitigating the bug where undo not only moves it back,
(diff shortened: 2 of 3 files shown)


packages/rich-text/src/plugins/DragAndDrop/index.tsx
```diff
@@ -48,7 +49,10 @@ export function createDragAndDropPlugin(): PlatePlugin {
         if (!dropDisallowed) {
           // Move the drop event to a new undo batch mitigating the bug where undo not only moves it back,
           // but also undoes a previous action: https://github.com/ianstormtaylor/slate/issues/4694
-          editor.history.undos.push([]);
+          (editor as unknown as HistoryEditor).history.undos.push({
```
aodhagan-cf, line 52: slate-history behind the scenes got refactored to use batch objects with operations instead of an array of arrays

packages/rich-text/package.json
```diff
@@ -67,7 +67,7 @@
     "is-plain-obj": "^3.0.0",
     "react-popper": "^2.3.0",
     "slate": "0.94.1",
-    "slate-history": "0.66.0",
+    "slate-history": "0.100.0",
```
chrishelgert, line 70: that's a tiny bump 🙄

YvesRijckaert (approved):
LGTM

chrishelgert (approved):

PR #1600, opened 2024-02-02
chore: upgrade to TS version 5.3.2 []

Chaoste:
@MayaGillilan updated the TS version already on the top level [recently](https://github.com/contentful/field-editors/pull/1579). I'm updating it now for all packages inside the repo.

### DIFF
--- a/packages/location/tsconfig.json
+++ b/packages/location/tsconfig.json
@@ -2,7 +2,6 @@
   "extends": "../../tsconfig.json",
   "include": ["src", "types"],
   "compilerOptions": {
-    "module": "esnext",
     "moduleResolution": "bundler",
     "lib": ["dom", "esnext"],
     "rootDir": "./src",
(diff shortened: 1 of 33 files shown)


packages/location/tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "extends": "../../tsconfig.json",
   "include": ["src", "types"],
   "compilerOptions": {
-    "module": "esnext",
```
aodhagan-cf, line 5: wondering what was the motivation behind removing this?

andipaetzold (approved):

packages/location/tsconfig.json
aodhagan-cf: Fair enough I only ask because in the packages it's esnext and at the top level it's es2020. Just wanted to make sure this is a conscious change

aodhagan-cf (approved):

packages/location/tsconfig.json
Chaoste: The module config is actually `exnext` on the top level :) 
https://github.com/contentful/field-editors/blob/5bbca863106912d6db500943b9bf77983059f29d/tsconfig.json#L5

PR #1604, opened 2024-02-13
chore: upgrade storybook [TOL-1763]

2wce:
We have upgraded storybook to the latest version with the following caveat:

Storybook identifies our repository as a create-react-app setup & because of that it advises us to install [this preset](https://www.npmjs.com/package/@storybook/preset-create-react-app). However if we install it then the following issues arise

- [ ] If we don't add the preset as a addon storybook has a critical error & can't load the react-webpack5 addon
- [ ] If we add the preset as a addon storybook has another critical error 

Both scenarios mean we can't complete a build as long that preset is in our setup. We might have to keep an eye on this as version 8 approaches hopefully the issue will be fixed before then


package.json
```diff
@@ -44,22 +44,20 @@
     "@babel/plugin-syntax-dynamic-import": "7.8.3",
     "@babel/plugin-syntax-flow": "7.21.4",
     "@babel/plugin-transform-runtime": "7.9.6",
-    "@babel/preset-env": "7.12.11",
+    "@babel/preset-env": "^7.23.9",
```
chrishelgert, line 47: we should pin all devDependencies (again)

.storybook/main.ts
```diff
@@ -24,12 +25,20 @@ const config: StorybookConfig = {
...
+    },
   },
   docs: {
     autodocs: false,
   },
 };
 
 export default config;
+
+function getAbsolutePath(value: string): any {
```
aodhagan-cf, line 42: Why is this necessary?

aodhagan-cf (approved):

PR #1565, opened 2023-12-20
test: [TOL-1361] Rich Text editor component tests

colomolo:
Moving Rich Text e2e test suites to component tests.

All e2e cleanups and test adjustments would be done in a follow up PRs.

PR #1605, opened 2024-02-13
feat: bumping plate to v30 [TOL-1834]

aodhagan-cf:
Bumping the plate plugin dependencies for rich text to v30


cypress/e2e/rich-text/RichTextEditor.Tables.spec.ts
```diff
@@ -177,10 +173,15 @@ describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
       );
     });
 
-    it('does not delete table header cells when selecting the whole table', () => {
+    it('does not delete table header cells when selecting the whole header row', () => {
```
aodhagan-cf, line 176: select all will delete the table so selecting the first row instead

cypress/e2e/rich-text/RichTextEditor.Tracking.spec.ts
```diff
@@ -129,7 +129,7 @@ describe('Rich Text Editor - Tracking', { viewportHeight: 2000 }, () => {
 
       richText.expectTrackingValue([
         action('paste', 'shortcut-or-viewport', {
-          characterCountAfter: 14,
+          characterCountAfter: 11,
```
aodhagan-cf, line 132: I don't know why this was 14 before...

cypress/e2e/rich-text/document-mocks/paragraphWithoutFormattings.js
```diff
@@ -34,7 +34,7 @@ export default {
           marks: [],
           nodeType: 'text',
           value:
-            " is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. ",
+            " is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. ",
```
aodhagan-cf, line 37: 🤷🏻‍♂️

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -64,10 +64,11 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
     <SdkProvider sdk={sdk}>
       <ContentfulEditorIdProvider value={id}>
         <div className={styles.root} data-test-id="rich-text-editor">
-          <PlateProvider
+          <Plate
             id={id}
             initialValue={initialValue}
-            plugins={plugins}
+            // FIXME plugins type is not working as expected
```
aodhagan-cf, line 70: FIXME

packages/rich-text/src/SyncEditorChanges.tsx
```diff
@@ -39,19 +40,18 @@ const useOnValueChanged = (onChange?: (doc: Contentful.Document) => unknown) =>
   React.useEffect(() => {
     const cb = createOnChangeCallback(onChange);
 
-    setEditorOnChange({
-      fn: (document) => {
-        // Skip irrelevant events e.g. mouse selection
-        const operations = editor?.operations.filter((op) => {
-          return op.type !== 'set_selection';
-        });
+    // FIXME double-check this is working
```
aodhagan-cf, line 43: This should be fine just need to double-check it

packages/rich-text/src/plugins/List/withList.ts
```diff
@@ -15,7 +15,8 @@ export const withList: WithOverride = (editor) => {
   const { deleteForward, deleteFragment } = editor;
 
   editor.deleteForward = (unit) => {
-    if (deleteForwardList(editor)) return;
+    // FIXME figure out how to handle this
```
aodhagan-cf, line 18: Not sure what needs to be passed here

packages/rich-text/package.json
```diff
@@ -49,18 +49,18 @@
...
-    "@udecode/plate-select": "23.7.0",
+    "@udecode/plate-basic-marks": "30.1.2",
+    "@udecode/plate-break": "30.1.2",
+    "@udecode/plate-common": "30.1.2",
+    "@udecode/plate-core": "30.1.2",
+    "@udecode/plate-list": "30.1.2",
+    "@udecode/plate-paragraph": "30.1.2",
+    "@udecode/plate-reset-node": "30.1.2",
+    "@udecode/plate-select": "30.1.2",
     "@udecode/plate-serializer-docx": "23.7.0",
```
aodhagan-cf, line 60: A lot of the pasting from word online and google doc break with this update. Probably better to handle in another ticket

cypress/e2e/rich-text/document-mocks/paragraphWithoutFormattings.js
aodhagan-cf: used to be a `&nbsp;`

packages/rich-text/src/plugins/Table/withInsertFragmentTable.ts
```diff
@@ -0,0 +1,166 @@
...
+                anchor: getStartPoint(editor, startCellPath),
+                focus: getEndPoint(editor, lastCellPath),
+              });
+            }
+          });
+
+          return;
+        }
+      } else if (fragment.length === 1 && fragment[0].type === ELEMENT_TABLE) {
+        // NOTE adjusting this from insertNode to insertFragment to removing preceding empty paragraph
```
aodhagan-cf, line 156: I changed this instead of inserting as node in the original implementation inserting fragment

cypress/e2e/rich-text/document-mocks/paragraphWithoutFormattings.js
aodhagan-cf: Since upgrading `&nbsp;` have been stripped out when pasting in microsoft word text

aodhagan-cf: https://github.com/udecode/plate/blob/18d99ddfde960744cc19d812cdf0690a378caaee/packages/core/src/plugins/html-deserializer/utils/cleanHtmlTextNodes.ts#L32C9-L35C10

packages/rich-text/src/plugins/Table/createTablePlugin.ts
```diff
@@ -38,7 +45,14 @@ export const createTablePlugin = (): PlatePlugin =>
     withOverrides: (editor, plugin) => {
       const { normalizeNode } = editor;
       // injects important fixes from plate's original table plugin
-      withTable(editor, plugin);
+      editor = withNormalizeTable(editor);
```
z0al, line 48: We reset this on line 59 below. So maybe drop it?

cypress/e2e/rich-text/RichTextEditor.Tables.spec.ts
```diff
@@ -12,10 +12,6 @@ import { RichTextPage } from './RichTextPage';
 describe('Rich Text Editor', { viewportHeight: 2000 }, () => {
   let richText: RichTextPage;
 
-  // copied from the 'is-hotkey' library we use for RichText shortcuts
-  const IS_MAC =
-    typeof window != 'undefined' && /Mac|iPod|iPhone|iPad/.test(window.navigator.platform);
-  const mod = IS_MAC ? 'meta' : 'control';
```
z0al, line 18: It was unused?

packages/rich-text/src/plugins/Table/createTablePlugin.ts
aodhagan-cf: done!

packages/rich-text/src/SyncEditorChanges.tsx
```diff
@@ -25,7 +25,8 @@ const useAcceptIncomingChanges = (incomingValue?: Value) => {
     }
 
     lastIncomingValue.current = incomingValue;
-    setEditorValue(editor, incomingValue);
+    // FIXME see why this casting is needed
```
chrishelgert, line 28: something we should do before the merging?

aodhagan-cf: We add the tracking property to our editor, with the new way things are structured at the top level we don't get the same type here unfortunately. I move the casting to where the editor is declared

cypress/e2e/rich-text/RichTextEditor.Tables.spec.ts
aodhagan-cf: It is not used anymore, highlighting table content in a different way now as the test was breaking due to the CMD+A logic

packages/rich-text/src/SyncEditorChanges.tsx
chrishelgert: aaaaand have you double checked? :D

packages/rich-text/src/plugins/Table/withInsertFragmentTable.ts
```diff
@@ -0,0 +1,165 @@
...
+  WithPlatePlugin,
+} from '@udecode/plate-common';
+import { ELEMENT_TABLE, getTableAbove, getTableGridAbove } from '@udecode/plate-table';
+import cloneDeep from 'lodash/cloneDeep.js';
+import { Path } from 'slate';
+
+import { insertNodes, PlateEditor } from '../../internal';
+
+/**
+ * If inserting a table,
```
chrishelgert, line 21: mh I don't catch up with the documentation, can you explain please?

packages/rich-text/src/plugins/Table/withInsertFragmentTable.ts
```diff
@@ -0,0 +1,165 @@
...
+          select(editor, {
+            anchor: getStartPoint(editor, cellEntries[0][1]),
+            focus: getEndPoint(editor, cellEntries.at(-1)![1]),
+          });
+          return;
+        }
+      }
+    }
+
+    if (insertedTable) {
```
chrishelgert, line 68: this would be the else block?

packages/rich-text/src/plugins/Table/withInsertFragmentTable.ts
```diff
@@ -0,0 +1,165 @@
...
+                focus: getEndPoint(editor, lastCellPath),
+              });
+            }
+          });
+
+          return;
+        }
+      } else if (fragment.length === 1 && fragment[0].type === ELEMENT_TABLE) {
+        // needed to insert as node, otherwise it will be inserted as text
+        insertNodes(editor, fragment, {
```
aodhagan-cf, line 151: this

packages/rich-text/src/plugins/Table/withInsertFragmentTable.ts
```diff
@@ -0,0 +1,165 @@
...
+
+import { insertNodes, PlateEditor } from '../../internal';
+
+/**
+ * If inserting a table,
+ * If block above anchor is a table,
+ * - Replace each cell above by the inserted table until out of bounds.
+ * - Select the inserted cells.
+ */
+export const withInsertFragmentTable = (editor: PlateEditor, { options }: WithPlatePlugin) => {
```
z0al, line 28: Is there a way to simplify this logic make use of the upstream function as much as possible? something like this pseudocode:

```javascript
import {withInsertFragmentTable as upstream} from '@plate/table'

export const withInsertFragmentTable = (editor, { options }) => {
	myEditor.insertFragment = (fragment) => {
	    if(condition){
         upstream(...)
      }else {
        // our logic
       }
	}
}
```

Or 

```javascript
import {withInsertFragmentTable as upstream} from '@plate/table'

export const withInsertFragmentTable = (editor, { options }) => {
	myEditor.insertFragment = (fragment) => {
	    upstream(...)

      if(condition){
        // undo upstream and override with our element
       }
	}
}
```

aodhagan-cf: I don't want to adjust the logic of this file as it's copied from plates implementation the only part that was adjusted is below

z0al: I know it's copied from plate which is why I want to eliminate as much as possible of it to:
1. get latest fixes when we upgrade
2. have less code to maintain

colomolo:
Wonder, how we're gonna merge these 🫠
https://github.com/contentful/field-editors/pull/1565

packages/rich-text/src/SyncEditorChanges.tsx
aodhagan-cf: I have haha

cypress/e2e/rich-text/RichTextEditor.Tables.spec.ts
z0al, line 176: I'd argue this is now a completely different test and probably worth a follow up to see if it's even needed.

z0al (approved):
LGTM 👍🏻 . Left a comment that can be handled in a follow up

cypress/e2e/rich-text/RichTextEditor.Tables.spec.ts
aodhagan-cf: Totally agree, will sync with @colomolo on this tomorrow

PR #1607, opened 2024-02-20
fix: removing npmrc file []

aodhagan-cf:
The `.npmrc` file looks like it broke our releases, removing to hopefully fix


chrishelgert (approved):
Let's try, if it works, maybe worth putting a comment why we do this :D

PR #1620, opened 2024-03-04
fix: editor focus [ZEND-4737]

YvesRijckaert:
For some unknown reason (probably an internal slate bug), the provided `useFocused();` from slate-react is always false.
Instead of relying on this hook to determine if the editor is in focus, we check it ourselves.
I removed the hook for now since we only used it for the hyperlink modal

### DIFF
--- a/packages/rich-text/src/plugins/Hyperlink/components/useHyperlinkCommon.tsx
+++ b/packages/rich-text/src/plugins/Hyperlink/components/useHyperlinkCommon.tsx
@@ -1,3 +1,5 @@
+import { useEffect, useState } from 'react';
+
 import { FieldAppSDK } from '@contentful/app-sdk';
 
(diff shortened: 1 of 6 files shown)


packages/rich-text/src/plugins/Hyperlink/components/useHyperlinkCommon.tsx
```diff
@@ -10,6 +12,28 @@ export function useHyperlinkCommon(element) {
...
+
+    if (editorElement) {
+      // Initially check if the editor is focused
+      setIsEditorFocused(document.activeElement === editorElement);
+
+      editorElement.addEventListener('focus', handleFocus);
+      editorElement.addEventListener('blur', handleBlur);
+    }
+
+    return () => {
```
colomolo, line 30: Should we move this return statement inside of the `if` above?

colomolo (approved):

PR #1621, opened 2024-03-06
feat: support external resources in resource links

veu:
Add support for external resources to the reference field editor.

## Details
* The biggest change is fetching _resource_ and _resource type_ entities as we’re extending the existing resource link logic and have already introduced the external resource card component in https://github.com/contentful/field-editors/pull/1602.
* Rich text doesn’t support external reference yet so the support is explicitly disabled. The changes here are only due to updated types.


Cyberxon (approved):
LGTM, nice refactoring work!

aodhagan-cf (approved):
LGTM 👍

PR #1622, opened 2024-03-07
feat: add card subtitle and Provider ID to the external card [DANTE-1407]

Cyberxon:
Before we have the ideal state for the product card we should be able to show some crucial information that doesn’t fit with the ideal card UI. 

In the title section of the card, we need to make sure to show the Provider name (in our case `Shopify`) before the Product type name (`Product`/`Product Variant`/`Collection`).

In the description section of the externalCard we should show the Product ID as a subtitle then show the description if there is any.

### DIFF
--- a/packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
+++ b/packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
@@ -40,7 +40,6 @@ const defaultProps = {
 const styles = {
   subtitle: css({
     color: tokens.gray600,
-    marginBottom: 'none',
   }),
   description: css({
     color: tokens.gray900,
@@ -71,7 +70,7 @@ function ExternalResourceCardDescription({
     );
   }
 
-  const truncatedDescription = truncate(description, 500, {});
+  const truncatedDescription = truncate(description, 180, {});
 
   return (
     <>
--- a/packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
+++ b/packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
@@ -1,9 +1,20 @@
 import * as React from 'react';
 
-import { Badge, EntryCard, MenuItem, MenuDivider } from '@contentful/f36-components';
+import {
+  Badge,
+  EntryCard,
+  MenuItem,
+  MenuDivider,
+  Paragraph,
+  Caption,
+} from '@contentful/f36-components';
+import tokens from '@contentful/f36-tokens';
+import { css } from 'emotion';
+import truncate from 'truncate';
 
 import { ExternalResourceInfo } from '../../common/EntityStore';
 import { ExternalResource, RenderDragFn } from '../../types';
+
 
 export interface ExternalResourceCardProps {
   info: ExternalResourceInfo;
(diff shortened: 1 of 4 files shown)


packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
```diff
@@ -27,10 +38,55 @@ const defaultProps = {
...
 };
 
+const styles = {
+  subtitle: css({
+    color: tokens.gray600,
+    marginBottom: 'none',
+  }),
+  description: css({
+    color: tokens.gray900,
+    marginBottom: 'none',
```
sbezludny, line 48: is that legal? I thought it was a Forma feature not css

Cyberxon: Doesn't it get translated to that same css property? 🤔

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
```diff
@@ -27,10 +38,55 @@ const defaultProps = {
...
+
+  if (!description) {
+    return (
+      <Paragraph className={styles.description} isWordBreak>
+        {subtitle}
+      </Paragraph>
+    );
+  }
+
+  const truncatedDescription = truncate(description, 500, {});
```
sbezludny, line 76: can we use <Paragraph isTruncated /> instead?

sbezludny (changes requested):

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
```diff
@@ -109,7 +167,12 @@ export function ExternalResourceCard({
             }
           : undefined
       }
-    />
+    >
+      <ExternalResourceCardDescription
+        subtitle={entity.fields.subtitle || `Product ID: ${entity.sys.id}`}
```
sbezludny, line 172: `Product ID: ${entity.sys.id}` this must not belong here. We must not couple with ProductId label anywhere

sbezludny (changes requested):

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
sbezludny: no, afaik it does not.

sbezludny (dismissed):

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
Cyberxon: When I talked about it with Hande she wanted a max of 2 ligns on the description and a max width of `642px` that's why I set it like that. using that option will make it way shorter, max of one lign I guess.

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
Cyberxon: This was added because we don't send the subtitles from the Backend just yet and according to hande it is a necessary information to have.

veu (approved):

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
sbezludny: it still feels weird not using existing apis (`isTruncated`). but not blocking

packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
```diff
@@ -27,10 +37,54 @@ const defaultProps = {
   hasCardRemoveActions: true,
 };
 
+const styles = {
+  subtitle: css({
+    color: tokens.gray600,
+    marginBottom: 'none',
```
sbezludny, line 43: this is still invalid

sbezludny (changes requested):

sbezludny (approved):
lgtm

PR #1624, opened 2024-03-14
fix(rich-text): respect sdk.field.locale text direction in text nodes [ZEND-4773]

z0al:
Slate detects text direction on a block level e.g. paragraph which is fine in most cases but its implementation is naive. For example, if you have an English word at the beginning of a paragraph written mostly in Arabic the whole paragraph would be marked as `ltr` instead of `rtl`. 

At Contentful, we don't need to guess the direction, we already have the locale information exposed in the Field SDK which is respected in other field editors e.g. SingleLineEditor. This brings some of that to RT but it's no where near complete. To fully support RTL we have to adjust the styling of some the elements e.g. blockquotes, tables ..etc.

This fixes the problem at hand and serves as a first step in the right direction.

### DIFF
--- a/packages/rich-text/src/plugins/Paragraph/Paragraph.tsx
+++ b/packages/rich-text/src/plugins/Paragraph/Paragraph.tsx
@@ -10,6 +10,7 @@ const styles = {
   [BLOCKS.PARAGRAPH]: css`
     line-height: ${tokens.lineHeightDefault};
     margin-bottom: 1.5em;
+    direction: inherit;
   `,
 };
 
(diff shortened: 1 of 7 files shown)


packages/rich-text/src/plugins/Paragraph/Paragraph.tsx
```diff
@@ -10,6 +10,7 @@ const styles = {
   [BLOCKS.PARAGRAPH]: css`
     line-height: ${tokens.lineHeightDefault};
     margin-bottom: 1.5em;
+    direction: inherit;
```
z0al, line 13: All these `inherit`s are here to force direction in Slate. AFAIK, there is no way to disable Slate's internal direction detection

colomolo (approved):
lgtm

PR #1627, opened 2024-03-19
test(richtext): use embed dialogs for testing the correct behaviour []

chrishelgert:


.circleci/config.yml
```diff
@@ -77,6 +77,14 @@ jobs:
               --reporter junit \
               --reporter-options "mochaFile=./cypress/reports/component/test-results.[hash].xml" \
               --browser chrome
+      - vault/get-secrets:
+          template-preset: packages-read
+      - run: |
+          echo "//npm.pkg.github.com/:_authToken=${GITHUB_PACKAGES_READ_TOKEN}" > ~/.npmrc
+          echo "@contentful:registry=https://npm.pkg.github.com" >> ~/.npmrc
+      - run:
```
chrishelgert, line 85: After this the splitting into wrokers should work correctly

z0al (approved):
LGTM. Btw, there is a `test` commit type for these kind of changes :)

cypress.config.ts
```diff
@@ -23,27 +23,6 @@ export default defineConfig({
     openMode: 0,
   },
   numTestsKeptInMemory: 1,
-  e2e: {
```
chrishelgert, line 26: We don't have e2e test anymore 😈

cypress/component/rich-text/RichTextEditor.Commands.spec.ts
```diff
@@ -8,7 +8,7 @@ import { mountRichTextEditor } from './utils';
 // the sticky toolbar gets in the way of some of the tests, therefore
 // we increase the viewport height to fit the whole page on the screen
 
-describe('Rich Text Editor - Commands', { viewportHeight: 2000 }, () => {
+describe('Rich Text Editor - Commands', { viewportHeight: 2000, viewportWidth: 1000 }, () => {
```
chrishelgert, line 11: All RichText texts are using the same width and height

PR #1633, opened 2024-04-08
feat: update design of error cards for resources [DANTE-1340]

msieroslawska:
The goal of this PR is to replace the custom Card that we were using for non-resolvable link with a variant of Entry Card that gives more information on the reason of error and the source of the data.

### Before for all resources (same space, xspace and external resources)
Screenshot 2024-04-08 at 17 46 15

### After for same space
01-same-space

### After for cross-space and external resources
02-xspace-external

### Before
Screenshot 2024-04-05 at 16 35 15

### After
03-rich-text

## Unsupported entities
Here we also updated the design to use the same layout for unsupported entities.

### Before
04-not-supported-before

### After
05-not-supported-after

## Notes
- Design for missing assets was not touched in the scope of this work (will be addressed later). Old implementation of the `MissingEntityCard` component that was previously serving all use cases has been extracted to a separate `MissingAssetCard` component.


packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
```diff
@@ -1,37 +1,42 @@
...
+  isDisabled?: boolean;
   isSelected?: boolean;
-  isDisabled: boolean;
   onRemove?: Function;
+  providerName?: string;
 }) {
+  const providerName = props.providerName ?? 'Source';
+  const description =
+    props.providerName === 'Contentful'
+      ? "Content was deleted or archived or you don't have access rights."
```
z0al, line 15: Archived entries don't result in this error, it's either deletion or access right. please update the copy

For example: 
Screenshot 2024-04-08 at 17 09 23

sbezludny:
Looking at the message:
`Entry is missing or inaccessible` vs `Content was deleted or archived or you don't have access rights.`

We no longer indicate a specific entity type that is missing -- entry or asset. What is the reason for this change, do we intentionally give less actionable information?

sbezludny:
Are we changing anything for Assets?

packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
```diff
@@ -1,37 +1,42 @@
...
+      : `Content was deleted or archived or you don't have access rights from the external provider. Contact your technical team.`;
```
sbezludny, line 16: > Contact your technical team

Do you think it helps them to get more help?

sbezludny, line 16: > you don't have access rights from the external provider

Can we indicate which one?

handekocabey86: The goal is for them to reach out to their support and not us because we know the reason for the error is them and not us. Happy to hear any suggestions you have on how we can convey this message.

packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
handekocabey86: Thank you for the info, we'll remove archived from the error messages.

packages/reference/src/components/MissingEntityCard/MissingEntityCard.tsx
sbezludny: the challenge I anticipate is actually finding someone who is in the position to help
1. How would anyone know to whom to reach out? In other words, who is the technical team for a content editor?
2. Would that be useful for a user of the Marketplace app (e.g. Shopify) that has an app installed on their space but does not manage the integration itself?

the alternative message could look like:
```
We were not able to find %YOUR_RESOURCE% with urn:%your_urn%. 

This may be due to various reasons beyond our control, such as changes or restrictions on the external provider's side. Verify the URN for accuracy.
```

msieroslawska:
The error copies have been adjusted to be less speculative and also to be suitable to all use cases.

Assets will be addressed in a separate ticket.

packages/reference/src/utils/getProviderName.ts
```diff
@@ -0,0 +1,8 @@
+export function getProviderName(resourceLink: string): string | undefined {
```
z0al, line 1: Can you please add JS doc here with examples of what's expected or some test cases? It's not so obvious for those of us not working directly with the cross-space feature.

z0al (approved):
LGTM. Left a small comment

packages/reference/src/utils/getProviderName.ts
msieroslawska: Test FTW! I added some simple unit tests as documentation.

PR #1637, opened 2024-04-15
style: make text field styles more consistent

MayaGillilan:


packages/single-line/src/SingleLineEditor.tsx
```diff
@@ -72,15 +72,23 @@ export function SingleLineEditor(props: SingleLineEditorProps) {
...
-            {props.withCharValidation === false && (
-              <div className={styles.validationRow}>
-                <CharCounter value={value || ''} checkConstraint={() => true} />
+            {!withCharValidation && (
+              <div className={styles.counterRow}>
+                <CharCounter
+                  value={value || ''}
+                  checkConstraint={checkConstraint}
+                  constraints={constraints}
+                />
```
z0al, line 91: Nit: Is just the git diff or are these two sections identical with the exception of the style class and `CharValidation`? 

Maybe it would look cleaner if do something like this?

```jsx
<div className={withCharValidation? styles.validationRow: styles.counterRow}>
    <CharValidation constraints={constraints} enabled={withCharValidation}/>

    <CharCounter
      value={value || ''}
      checkConstraint={checkConstraint}
      constraints={constraints}
    />
</div>
```

Don't feel strongly about it though.

packages/single-line/src/styles.ts
```diff
@@ -6,7 +6,15 @@ export const validationRow = css({
   justifyContent: 'space-between',
   fontSize: tokens.fontSizeM,
   marginTop: tokens.spacingXs,
-  color: tokens.gray700,
+  color: tokens.gray500,
+});
+
+export const counterRow = css({
```
z0al, line 12: Nit: this is also identical but one attribute. Can we avoid this duplication somehow?

e.g. 
```javascript
export const validationRow = (align) => css({
  display: 'flex',
  justifyContent: align === 'value'? 'space-between': 'flex-end',
  fontSize: tokens.fontSizeM,
  marginTop: tokens.spacingXs,
  color: tokens.gray500,
});
```

z0al (approved):

packages/single-line/src/SingleLineEditor.tsx
MayaGillilan: They are, I think that is indeed cleaner

packages/single-line/src/styles.ts
MayaGillilan: Actually we could get around needing the two styles/align flag by doing
`<span>{constraints.isDefaultConstraint || !enabled ? '' : `Max. ${constraints.max} characters`}</span>` in CharValidation. Any preference?

PR #1639, opened 2024-04-16
fix: passing failed to fetched entry card when inaccessible []

aodhagan-cf:
Previously we were seeing an infinite loader when the entries are deleted. Now showing the missing entry card instead

**before**

Screenshot 2024-04-16 at 14 59 56

**after**

Screenshot 2024-04-16 at 15 00 06


chrishelgert (approved):
Thought we would have fixed this already in the past 🤔 
Would you like to do the same for FetchingWrappedAssetCard please?

z0al (approved):

PR #1640, opened 2024-04-18
style: Remove CharValidation component and further simplify text editors

MayaGillilan:


packages/_shared/src/types.ts
```diff
@@ -2,6 +2,7 @@ export type ValidationType =
   | {
       type: 'max';
       max: number;
+      isDefaultConstraint?: boolean;
```
z0al, line 5: Where can I read more about this?

MayaGillilan: What do you mean? The idea is we already have an existing default limit of 256 chars for Symbols and 50k for Text, so I just added this flag to be able to distinguish between it and a custom one, but it's not relevant anymore if we remove the CharValidation component

z0al: > The idea is we already have an existing default limit of 256 chars for Symbols and 50k for Text, so I just added this flag to be able to distinguish between it and a custom one, but it's not relevant anymore if we remove the CharValidation component so actually it should be removed anyway

This :) . It was not obvious and I would even argue it's not needed. But since we are removing it anyway 👍🏻

z0al (approved):

2wce (approved):

PR #1641, opened 2024-04-23
fix(richtext-editor): handle failed inline entries [ZEND-4919,ZEND-4913]

chrishelgert:
follow up on https://github.com/contentful/field-editors/pull/1639

Currently, the rich text editor fails as we try to read 'contentType' from 'failed'.sys.
Instead patching the entry to 'failed', it uses now the status to check for errors

### DIFF
--- a/packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
+++ b/packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
@@ -7,11 +7,13 @@ import {
   useEntityLoader,
   MissingEntityCard,
   WrappedAssetCard,
+  type UseEntityStatus,
 } from '@contentful/field-editor-reference';
 import areEqual from 'fast-deep-equal';
 
 interface InternalAssetCardProps {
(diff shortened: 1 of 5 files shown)


aodhagan-cf (approved):
LGTM

packages/rich-text/src/plugins/shared/FetchingWrappedAssetCard.tsx
```diff
@@ -36,6 +34,10 @@ const InternalAssetCard = React.memo((props: InternalAssetCardProps) => {
     );
   }
 
+  if (!props.asset) {
```
z0al, line 37: Shouldn't we explicitly check for `loading`?

chrishelgert: would need some more refactorings, can propose that in a follow up PR 👍

chrishelgert: @z0al what do you think: https://github.com/contentful/field-editors/pull/1643

PR #1643, opened 2024-04-23
refactor(rich-text): better handling for status []

chrishelgert:
follow up on https://github.com/contentful/field-editors/pull/1641

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
@@ -21,6 +21,83 @@ const styles = {
   }),
 };
 
+type InternalFetchingWrappedInlineEntryCardProps = Pick<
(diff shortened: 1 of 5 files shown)


packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
```diff
@@ -21,6 +21,83 @@ const styles = {
...
+        entry,
+        contentType,
+        localeCode: locale,
+        defaultLocaleCode: defaultLocale,
+        defaultTitle: 'Untitled',
+      }),
+    [entry, contentType, locale, defaultLocale]
+  );
+
+  const contentTypeName = contentType ? contentType.name : '';
```
z0al, line 72: This feels wrong but can be fixed in a follow (already there in the existing code)

chrishelgert: I just it to also not show the `:` if the contentType would not be available 👍

PR #1646, opened 2024-05-02
fix: [] Consider text direction when positioning cell actions inside of Rich Text Editor

colomolo:
With LTR text actions dropdown goes to the right:
image

With RTL text actions dropdown goes to the left:
image

### DIFF
--- a/packages/rich-text/src/plugins/Table/components/TableActions.tsx
+++ b/packages/rich-text/src/plugins/Table/components/TableActions.tsx
@@ -19,8 +19,8 @@ import { isTableHeaderEnabled } from '../helpers';
 export const styles = {
   topRight: css({
     position: 'absolute',
-    top: '6px',
-    right: '5px',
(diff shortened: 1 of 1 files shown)


packages/rich-text/src/plugins/Table/components/TableActions.tsx
```diff
@@ -19,8 +19,8 @@ import { isTableHeaderEnabled } from '../helpers';
 export const styles = {
   topRight: css({
     position: 'absolute',
-    top: '6px',
-    right: '5px',
+    insetBlockStart: '6px',
```
z0al, line 22: TIL

z0al (approved):

packages/rich-text/src/plugins/Table/components/TableActions.tsx
aodhagan-cf: also TIL

PR #1650, opened 2024-05-15
feat: add locales to entry status calculations

MayaGillilan:


packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -199,10 +199,53 @@ export function getEntryTitle({
...
+export function getEntryStatus(
+  //TODO: remove union after fieldStatus is added to App SDK
+  sys: Entry['sys'] & { fieldStatus?: FieldStatus },
+  locales?: LocaleProps[]
+) {
   if (!sys || (sys.type !== 'Entry' && sys.type !== 'Asset')) {
     throw new TypeError('Invalid entity metadata object');
   }
+
+  if (sys.fieldStatus) {
```
chrishelgert, line 217: I think it should be after `deletedVersion` and `archivedVersion`

packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -199,10 +199,53 @@ export function getEntryTitle({
...
+type AsyncPublishStatus = 'draft' | 'published' | 'changed';
+
+type FieldStatus = {
+  '*': Record<string, AsyncPublishStatus>;
+};
+
+export function getEntryStatus(
+  //TODO: remove union after fieldStatus is added to App SDK
+  sys: Entry['sys'] & { fieldStatus?: FieldStatus },
+  locales?: LocaleProps[]
```
chrishelgert, line 211: NIT: could make it more flexible if you only want to provide one locale
```suggestion
  locale?: LocaleProps | LocaleProps[]
```

packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -199,10 +199,53 @@ export function getEntryTitle({
...
+    let status: AsyncPublishStatus = 'draft';
+    if (locales) {
+      if (locales.length === 1) {
+        const localeCode = locales[0].code;
+        return sys.fieldStatus['*'][localeCode] || 'draft';
+      }
+      const filteredStatusList = Object.entries(sys.fieldStatus['*']).filter(([localeCode, _]) =>
+        locales.some((locale) => locale.code === localeCode)
+      );
+      filteredStatusList.forEach(([_, fieldStatus]) => {
```
chrishelgert, line 227: NIT: we could combine this into one loop, so we don't have to iterate twice :)

```
for (const [localeCode, fieldStatus] of sys.fieldStatus['*']) {
  if (locales.some(locale => locale.code === localeCode) {
    if (fieldStatus === 'changed') {
      status = fieldStatus
      break
    }
    
    if (fieldStatus === 'published') {
      status === 'published'
     }
  }
}
```

packages/rich-text/src/plugins/EmbeddedEntityInline/FetchingWrappedInlineEntryCard.tsx
```diff
@@ -109,6 +110,7 @@ interface FetchingWrappedInlineEntryCardProps {
 export function FetchingWrappedInlineEntryCard(props: FetchingWrappedInlineEntryCardProps) {
   const { data: entry, status: requestStatus } = useEntity<Entry>('Entry', props.entryId);
   const { getEntityScheduledActions } = useEntityLoader();
+  const locales = useLocalesContext();
```
chrishelgert, line 113: Not sure why we need this, we should have the correct locale already from `props.sdk.field.locale` or is that a different one?

MayaGillilan: That only gives us the locale code, although currently it's only used to match against the fieldStatus so we could change that to just accept `string | string[]` instead

chrishelgert: yeah sounds good and then rename it to localeCode(s) for the entityHelper 👍

chrishelgert: then we also don't need contentful-management as dependency :)

packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -199,14 +199,66 @@ export function getEntryTitle({
...
+      Object.values(sys.fieldStatus['*']).forEach((fieldStatus) => {
+        if (fieldStatus === 'changed') {
+          status = fieldStatus;
+          return;
+        }
+        if (fieldStatus === 'published') {
+          status = fieldStatus;
+        }
+      });
+    }
```
chrishelgert, line 260: What do you think about something like this?
Could make it a little bit tinier and easier to read?

```
const condition = (locale) => {
  if (Array.isArray(localeCodes)) {
    return localeCodes.includes(locale)
  }
  
  return localeCodes ? localeCodes === locale : true
}
    
Oject.entries(sys.fieldStatus['*']).forEach(([localeCode, fieldStatus]) => {
  if (condition(localeCode)) {
    if (fieldStatus === 'changed') {
      status = fieldStatus;
      return;
    }
    if (fieldStatus === 'published') {
      status = fieldStatus;
    }
  }
})

chrishelgert (approved):

PR #1654, opened 2024-05-17
fix: action popup opacity [TOL-2131]

YvesRijckaert:
Due to a recent change in f36 the opacity was wrong for the action menu popup for assets

### DIFF
--- a/packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
+++ b/packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
@@ -14,7 +14,9 @@ const styles = {
   }),
   fileInformation: {
     menuItem: css({
-      opacity: 1,
+      '&:disabled': {
+        opacity: 1,
+      },
     }),
     dl: css({
       backgroundColor: tokens.gray100,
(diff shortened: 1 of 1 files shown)


packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
```diff
@@ -56,7 +58,12 @@ export function renderAssetInfo(props: { entityFile: File }) {
 
   return [
     <Menu.SectionTitle key="file-section">File info</Menu.SectionTitle>,
-    <Menu.Item key="file-information" className={styles.fileInformation.menuItem} isDisabled>
+    <Menu.Item
+      key="file-information"
+      className={styles.fileInformation.menuItem}
+      isDisabled
+      disabled
```
colomolo, line 65: Why is it needed? isDisabled doesn't add actual disabled attribute?

YvesRijckaert: Yes 😅

YvesRijckaert: It just sets tabIndex of -1

colomolo: So annoying. Same was with NavList.Item 😞

colomolo: Maybe worth pointing this out to the F36 team.

colomolo (approved):

PR #1655, opened 2024-05-17
feat: allow onBlur and onFocus on SingleLineEditor []

nerdess:
When building a Contentful App I was missing onBlur and onFocus events on SingleLineEditor. Added with this PR as optional props.


chrishelgert (approved):
Thanks for your contribution :)

PR #1657, opened 2024-05-17
feat: undo redo rich text tracking [TOL-1987]

YvesRijckaert:
Add tracking for undo and redo buttons
- via toolbar
- via shortcut


chrishelgert (approved):
LGTM

PR #1660, opened 2024-05-21
fix: spaces removed for non english text [ZEND-5011]

YvesRijckaert:
## Summary
This pull request addresses an issue where spaces were removed from Korean text pasted into the rich text field. The problem was caused by the `cleanHtmlEmptyElements` plugin from Plate, which removes all empty elements, including `<span>` elements used by MS Word to represent spaces in Korean text.
Content support ticket: https://contentful.atlassian.net/browse/ZEND-5011

## Problem
MS Word adds `<span>` elements for each Korean word, and spaces are represented by empty `<span>` elements. The `cleanHtmlEmptyElements` plugin removed these empty `<span>` elements, causing spaces to be removed in Korean text when pasted into the rich text field. 

## Solution
We have customized the plugin to ignore empty `<span>` elements, ensuring that spaces in Korean text are preserved. This change means we no longer rely on the `cleanHtmlEmptyElements` plugin from Plate.

## Additional Changes
Added a test to ensure this behavior is maintained in the future.

### DIFF
--- a/packages/rich-text/src/plugins/DeserializeDocx/cleanHtmlEmptyElements.tsx
+++ b/packages/rich-text/src/plugins/DeserializeDocx/cleanHtmlEmptyElements.tsx
@@ -1,0 +1,32 @@
+/*
+ The cleanHtmlEmptyElements from Plate is overly aggressive in removing elements, removing those that are necessary for maintaining proper spacing (span).
+ We are using a slightly adjusted version of the cleanHtmlEmptyElements function to prevent the removal of span elements.
+ */
+import { traverseHtmlElements } from '@udecode/plate-common';
+
+const ALLOWED_EMPTY_ELEMENTS = new Set(['BR', 'IMG', 'TH', 'TD', 'SPAN']);
+
+const isEmpty = (element: Element): boolean => {
+  return !ALLOWED_EMPTY_ELEMENTS.has(element.nodeName) && !element.innerHTML.trim();
+};
+
+const removeIfEmpty = (element: Element): void => {
+  if (isEmpty(element)) {
+    const { parentElement } = element;
+
+    element.remove();
+
+    if (parentElement) {
+      removeIfEmpty(parentElement);
+    }
+  }
+};
+
+/** Remove empty elements from rootNode. Allowed empty elements: BR, IMG. */
+export const cleanHtmlEmptyElements = (rootNode: Node): void => {
+  traverseHtmlElements(rootNode, (element) => {
+    removeIfEmpty(element);
+
+    return true;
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/plugins/DeserializeDocx/cleanHtmlEmptyElements.tsx
```diff
@@ -0,0 +1,33 @@
...
+    if (parentElement) {
+      removeIfEmpty(parentElement);
+    }
+  }
+};
+
+/** Remove empty elements from rootNode. Allowed empty elements: BR, IMG. */
+export const cleanHtmlEmptyElements = (rootNode: Node): void => {
+  traverseHtmlElements(rootNode, (element) => {
+    console.log('element', element.nodeName, element.innerHTML);
```
chrishelgert, line 28: 👀

YvesRijckaert: Oopsie, removed 🙇

packages/rich-text/src/plugins/DeserializeDocx/cleanHtmlEmptyElements.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    element.remove();
+
+    if (parentElement) {
+      removeIfEmpty(parentElement);
+    }
+  }
+};
+
+/** Remove empty elements from rootNode. Allowed empty elements: BR, IMG. */
```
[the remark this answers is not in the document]
YvesRijckaert: This is inside the DeserialzeDocx plugin, wouldn't it be already scoped to MS Word documents then?

packages/rich-text/src/plugins/DeserializeDocx/cleanHtmlEmptyElements.tsx
```diff
@@ -0,0 +1,32 @@
+/*
+ The cleanHtmlEmptyElements from Plate is overly aggressive in removing elements, removing those that are necessary for maintaining proper spacing (span).
```
z0al, line 2: Is there an open upstream issue we can link here? are maintainers aware of the behavior?

z0al (approved):

packages/rich-text/src/plugins/DeserializeDocx/cleanHtmlEmptyElements.tsx
YvesRijckaert: Didn't find one, I'll create one 👍

PR #1662, opened 2024-05-23
feat: allow MultipleEntryReferenceEditor to render children prop

nerdess:
The commit message is a bit misleading. This is about offering the ability to extend the body of an MultipleReferenceEditor entry with custom React elements.

Check storyboard > Multiple Entry > Custom Children.

The content "Some custom children..." has been added using the new optional `renderCustomChildren` prop.

Bildschirmfoto 2024-05-23 um 14 54 52

```
<MultipleEntryReferenceEditor
    renderCustomChildren={() => {
      return <>Some custom children. Anything can be rendered here!</>;
    }}
    viewType="link"
    sdk={sdk}
    isInitiallyDisabled={isInitiallyDisabled}
    parameters={{
      instance: instanceParams || {
        showCreateEntityAction: true,
        showLinkEntityAction: true,
      },
  }}
/>
```

I was missing this feature when developing a custom Contentul app. Might be useful for others.


colomolo:
Hi @nerdess. `MultipleEntryReferenceEditor` component have `renderCustomCard` prop that has similar functionality. I believe you can achive the same result with it.

Please refer to the Storybook:
https://contentful-field-editors.netlify.app/?path=/story/editors-multiple-entry--custom-card

nerdess:
@colomolo Yeah I am aware of `renderCustomCard` and I started with it but faced limitations. It's been half a year since I did that amendment, if I remember correctly it was like this: with just popping in the children I did have the correct header out-of-the-box and also the drag/drop functionality of the list. I did not have to rebuild all that.

nerdess:
@colomolo It's been a while but would there be an interest in this PR if I fix the tests?

colomolo:
Hi @nerdess. Thanks for getting back to this. We discussed this today and as there were no other requests for this functionality, we agreed to avoid adding it as for now. Sorry for a long turnaround.

colomolo:
I will close this now. Please feel free to reach out with any other questions.

PR #1665, opened 2024-05-24
feat: support strikethrough

YvesRijckaert:
Adding support for strikethrough based on the PR for superscript / subscript https://github.com/contentful/field-editors/pull/1280/files

Screenshot 2024-05-27 at 09 25 38

### Todos
- [x] Update @contentful/rich-text-types after this PR is merged: https://github.com/contentful/rich-text/pull/562

### DIFF
--- a/cypress/component/rich-text/RichTextEditor.Marks.spec.ts
+++ b/cypress/component/rich-text/RichTextEditor.Marks.spec.ts
@@ -21,7 +21,7 @@ describe('Rich Text Editor - Marks', { viewportHeight: 2000, viewportWidth: 1000
   });
 
   const findMarkViaToolbar = (mark: string) => {
-    if (mark === 'code' || mark === 'superscript' || mark === 'subscript') {
+    if (['code', 'superscript', 'subscript', 'strikethrough'].includes(mark)) {
       cy.findByTestId('dropdown-toolbar-button').click();
       return cy.findByTestId(`${mark}-toolbar-button`);
     } else {
@@ -30,7 +30,7 @@ describe('Rich Text Editor - Marks', { viewportHeight: 2000, viewportWidth: 1000
   };
 
   const toggleMarkViaToolbar = (mark: string) => {
-    if (mark === 'code' || mark === 'superscript' || mark === 'subscript') {
+    if (['code', 'superscript', 'subscript', 'strikethrough'].includes(mark)) {
       cy.findByTestId('dropdown-toolbar-button').click();
       cy.findByTestId(`${mark}-toolbar-button`).click();
     } else {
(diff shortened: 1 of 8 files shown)


cypress/component/rich-text/RichTextEditor.Marks.spec.ts
```diff
@@ -21,7 +21,12 @@ describe('Rich Text Editor - Marks', { viewportHeight: 2000, viewportWidth: 1000
   });
 
   const findMarkViaToolbar = (mark: string) => {
-    if (mark === 'code' || mark === 'superscript' || mark === 'subscript') {
+    if (
+      mark === 'code' ||
```
chrishelgert, line 25: what do you think about?
`['code', 'superscript', 'subscript', 'strikethrough'].includes(mark)`

you could also extract it into a constant outside of the function and re-use it in `toggleMarkViaToolbar`

packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -129,7 +137,8 @@ const Toolbar = ({ isDisabled }: ToolbarProps) => {
   const dropdownItemsAvailable =
     isMarkEnabled(sdk.field, MARKS.SUPERSCRIPT) ||
     isMarkEnabled(sdk.field, MARKS.SUBSCRIPT) ||
-    isMarkEnabled(sdk.field, MARKS.CODE);
+    isMarkEnabled(sdk.field, MARKS.CODE) ||
+    isMarkEnabled(sdk.field, MARKS.STRIKETHROUGH);
```
chrishelgert, line 141:
```suggestion
  const dropdownItemsAvailable = [MARKS.SUPERSCRIPT, MARKS.SUBSCRIPT, MARKS.CODE, MARKS.STRIKETHROUGH].some(mark => isMarkEnabled(sdk.field, mark)
```
less duplication and easier to extend if there should be even more marks 😅

packages/rich-text/src/plugins/Marks/Strikethrough.tsx
```diff
@@ -0,0 +1,82 @@
...
+  createDefaultStrikethroughPlugin({
+    type: MARKS.STRIKETHROUGH,
+    component: Strikethrough,
+    handlers: {
+      onKeyDown: buildMarkEventHandler(MARKS.STRIKETHROUGH),
+    },
+    deserializeHtml: {
+      rules: [
+        {
+          validNodeName: ['S'],
```
chrishelgert, line 78: ❓ why is that one uppercase? 🤔

YvesRijckaert: Not sure, they are all uppercase (italic, bold, etc). Maybe something plate / slate specific? 😖

packages/rich-text/src/Toolbar/index.tsx
```diff
@@ -126,10 +134,12 @@ const Toolbar = ({ isDisabled }: ToolbarProps) => {
     isMarkEnabled(sdk.field, MARKS.BOLD) ||
     isMarkEnabled(sdk.field, MARKS.ITALIC) ||
     isMarkEnabled(sdk.field, MARKS.UNDERLINE);
-  const dropdownItemsAvailable =
-    isMarkEnabled(sdk.field, MARKS.SUPERSCRIPT) ||
-    isMarkEnabled(sdk.field, MARKS.SUBSCRIPT) ||
-    isMarkEnabled(sdk.field, MARKS.CODE);
+  const dropdownItemsAvailable = [
```
chrishelgert, line 137: We currently do the same check on line 81, could we combine it?

chrishelgert (approved):

PR #1670, opened 2024-06-06
fix: do new status calculation only conditionally [TOL-2143]

chrishelgert:
Restores the commented status calculation by locale, but only do it if the feature is enabled.
For the beta that comes as an instance parameter.

disabled here: https://github.com/contentful/field-editors/pull/1666/files

### DIFF
--- a/packages/rich-text/src/plugins/EmbeddedResourceInline/FetchingWrappedResourceInlineCard.tsx
+++ b/packages/rich-text/src/plugins/EmbeddedResourceInline/FetchingWrappedResourceInlineCard.tsx
@@ -9,7 +9,7 @@ import { INLINES } from '@contentful/rich-text-types';
 
 import { truncateTitle } from '../../plugins/shared/utils';
 
-const { getEntryTitle, getEntryStatus } = entityHelpers;
+const { getEntryTitle, getEntityStatus } = entityHelpers;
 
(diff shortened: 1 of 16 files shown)


packages/rich-text/src/plugins/EmbeddedResourceInline/FetchingWrappedResourceInlineCard.tsx
```diff
@@ -54,7 +54,10 @@ export function FetchingWrappedResourceInlineCard(props: FetchingWrappedResource
     defaultTitle: 'Untitled',
   });
   const truncatedTitle = truncateTitle(title, 40);
-  const status = getEntryStatus(entry.sys, props.sdk.field.locale);
+  const status = getEntityStatus(
+    entry.sys,
+    props.sdk.parameters.instance.useLocalizedEntityStatus ? props.sdk.field.locale : undefined
```
z0al, line 59: Are we still planning to later introduce something in the spirit of `sdk.entity.getStatus(sys: EntrySys)` that houses the conditional logic?

chrishelgert: yes, still waiting for an answer and will then create a follow up ticket for this 👍

PR #1688, opened 2024-06-20
feat: [TOL-2175] Pass isLocalePublishingEnabled flag via instance param

colomolo:
Pass flag for further use dependant components.

### DIFF
--- a/packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
+++ b/packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
@@ -92,12 +92,10 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
       renderDragHandle: props.renderDragHandle,
       onEdit,
       onRemove,
-      useLocalizedEntityStatus:
-        props.sdk.parameters.instance.useLocalizedEntityStatus &&
-        props.sdk.parameters.instance.isLocalePublishingEnabled,
+      useLocalizedEntityStatus: props.sdk.parameters.instance.useLocalizedEntityStatus,
       isLocalized:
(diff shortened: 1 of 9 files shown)


packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
```diff
@@ -92,8 +92,12 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
       renderDragHandle: props.renderDragHandle,
       onEdit,
       onRemove,
-      useLocalizedEntityStatus: props.sdk.parameters.instance.useLocalizedEntityStatus,
-      isLocalized: !!('localized' in props.sdk.field && props.sdk.field.localized), // missing in types :(
+      useLocalizedEntityStatus:
+        props.sdk.parameters.instance.useLocalizedEntityStatus &&
+        props.sdk.parameters.instance.isLocalePublishingEnabled,
```
chrishelgert, line 97: `useLocalizedEntityStatus` should be the same as `isLocalePublishingEnabled`?

useLocalizedEntityStatus resolves to the feature flag / entitlement check

colomolo: Where is that being set?

chrishelgert: web app: src/javascripts/app/entity_editor/EntityField/useWidgetApi.ts#L226

chrishelgert:
`props.sdk.parameters.instance.isLocalePublishingEnabled` should the remaining ones be `props.sdk.parameters.instance.useLocalizedEntityStatus` too?

PR #1690, opened 2024-06-25
fix(entityHelpers): adjust getEntityStatus and add tests [TOL-2219]

chrishelgert:


packages/_shared/src/utils/entityHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+              type,
+              version: 1,
+            }),
+            ['en-US', 'de-DE', 'fr-FR']
+          );
+
+          expect(result).toEqual('changed');
+        });
+
+        test('returns changed if at least one selected locale has a fieldStatus which is changed', () => {
```
z0al, line 63: Isn't this the same as the one above it 🤔

packages/_shared/src/utils/entityHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+              type,
+              version: 1,
+            }),
+            ['en-US', 'de-DE', 'fr-FR']
+          );
+
+          expect(result).toEqual('published');
+        });
+
+        test('returns published if at least one selected locale has a fieldStatus which is changed  (and none changed)', () => {
```
z0al, line 101: Um..but why? "changed" status represents having a mix of draft + published content. In this test, the german is published but the french in draft so I'd expect the overall status to be "changed" 🤔

chrishelgert: Correct for outside of locale based publishing.
inside of locale-based publishing the most advanced state wins and it's not aggregated correctly. (This is the current assumption that this brings the most benefit for the editors)

chrishelgert: can update the description of the test, to reflect this better

packages/_shared/src/utils/entityHelpers.test.ts
chrishelgert: you're right, will remove it 👍

YvesRijckaert (approved):

PR #1696, opened 2024-07-08
feat: enabling lists instead table cells []

aodhagan-cf:
WIP: testing enabling lists inside of table cells

- toolbar list icon is disabled when inside a table header cell


packages/rich-text/src/plugins/List/transforms/unwrapList.ts
```diff
@@ -1,36 +1,97 @@
...
-    } while (!hasUnliftedListItems(editor, at));
 
-    // finally unwrap all lifted items
-    unwrapNodes(editor, {
-      at,
-      match: { type: BLOCKS.LIST_ITEM },
-      split: false,
-    });
+      if (licEntry) {
+        // Special case for embedded entry and asset
```
aodhagan-cf, line 63: Special handling for embedded assets and entries, aside from this everything else is pretty much the same as the source

PR #1704, opened 2024-07-22
feat: 🎸 fetch all resource types

anho:
We want to have all available resource types in memory hence we fetch all available pages. Depends on https://github.com/contentful/contentful-management.js/pull/2365.


package.json
```diff
@@ -184,5 +184,8 @@
   },
   "publishConfig": {
     "registry": "https://npm.pkg.github.com/"
+  },
+  "dependencies": {
```
chrishelgert, line 188: why do we need it there as a dependency too?

anho: I don't have an answer for you, was added by yarn 🤷🏻‍♂️

chrishelgert (approved):
LGTM

PR #1717, opened 2024-08-06
fix: add a missing handler for ERL click [DANTE-1735]

msieroslawska:
External Resource Cards had no `onEdit` handler provided in the parent component, therefore clicking on the card had no effect. This PR fixes this issue and opens the URL provided in `externalUrl` (if existing).

https://contentful.atlassian.net/browse/DANTE-1734


packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx
```diff
@@ -86,7 +86,6 @@ ExternalResourceCardDescription.displayName = 'ExternalResourceCardDescription';
 export function ExternalResourceCard({
   info,
   isClickable,
-  onEdit,
```
mayakarabula, line 89: should we remove it from `ExternalResourceCardProps` as we are not using it any more?

mayakarabula (approved):

PR #1718, opened 2024-08-07
feat: wrapping connected rte instead of root one with entity provider []

aodhagan-cf:
To ensure that the reference package used is the same version as the rich text editor


chrishelgert:
NIT: wonder if it would more sense to use the default richtext one and deprecate the export of `ConnectedRichTextEditor`.
The last time we used the default one, we had an issue if multiple locales were selected in the snapshot compare, maybe that's just a quickfix too?

chrishelgert (approved):
Approved if we want to go with this approach

PR #1719, opened 2024-08-09
fix(rich-text): changing scroll position with \r [TOL-2365]

chrishelgert:
- Normalize the content before rendering it
- This will not trigger a change event for the editor directly
- The next time the content is adjusted and saved, it will be cleaned up internally

### DIFF
--- a/packages/rich-text/src/plugins/Normalizer/baseRules.ts
+++ b/packages/rich-text/src/plugins/Normalizer/baseRules.ts
@@ -3,6 +3,7 @@ import { BLOCKS, TEXT_CONTAINERS } from '@contentful/rich-text-types';
 import { INLINE_TYPES } from '../../helpers/editor';
 import { transformWrapIn } from '../../helpers/transformers';
 import { getParentNode, isText } from '../../internal/queries';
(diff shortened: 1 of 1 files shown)


2wce (approved):

packages/rich-text/src/plugins/Normalizer/baseRules.ts
```diff
@@ -34,4 +35,14 @@ export const baseRules: Required<NormalizerRule>[] = [
     },
     transform: transformWrapIn(BLOCKS.PARAGRAPH),
   },
+  {
+    // Replaces `\r` as it causes issues with scroll position and focus
+    match: isText,
+    validNode: (_editor, [node]) => typeof node.text === 'string' && !node.text.includes('\r'),
+    transform: (editor, [node, path]) => {
+      if (typeof node.text === 'string') {
```
z0al, line 43: Should we remove this if since the check is handled in the predicate? otherwise the callback here theoretically doesn't always return a value.

chrishelgert: in theory yes, we could then to do a type cast to satisfy typescript

PR #1725, opened 2024-08-21
chore: upgrade type-fest[TOL-1741]

2wce:


packages/reference/package.json
```diff
@@ -51,7 +51,7 @@
     "moment": "^2.20.0",
     "p-queue": "^4.0.0",
     "react-intersection-observer": "9.4.0",
-    "type-fest": "2.17.0"
+    "type-fest": "4.25.0"
```
chrishelgert, line 54: please don't pin dependencies on libraries

is type-fest even a dependency or a dev-dependency?

chrishelgert: if we can move it to devDependency, do we even need it here or can we remove it as it's on the root package.json

2wce: i think it was added as a dependency here but it's probably correct to have it as a dev dependency

chrishelgert (approved):

Issue #1236, opened 2022-09-14
RichTextEditor value and onChange props


vitalbone:
Hi, I've been trying to understand if it's possible to add an initial value and change the value of the `RichTextEditor` field. 

There doesn't seem to be any documentation so I'm not sure how it's supposed to function. Pardon my naiveté but I was hoping it would be as simple as the following however this doesn't work.

```ts
import 'codemirror/lib/codemirror.css'
import React from 'react'
import { BLOCKS } from '@contentful/rich-text-types'
import { FieldExtensionSDK } from '@contentful/app-sdk'
import { useSDK, useAutoResizer } from '@contentful/react-apps-toolkit'
import { RichTextEditor } from '@contentful/field-editor-rich-text'

const INITIAL_DOCUMENT = {
  nodeType: BLOCKS.DOCUMENT,
  data: {},
  content: [
    {
      nodeType: BLOCKS.PARAGRAPH,
      data: {},
      content: [
        {
          nodeType: 'text',
          value: 'Initial value',
          marks: [],
          data: {},
        },
      ],
    },
  ],
}

const Field = () => {
  const sdk = useSDK<FieldExtensionSDK>()
  const [value, setValue] = React.useState(INITIAL_DOCUMENT)
  
  useAutoResizer()

  function handleTextChange(doc: any) {
    doc.content[0].content[0].value = 'Updated value'
    setValue(doc)
  }

  return (
    <RichTextEditor
      sdk={sdk}
      isInitiallyDisabled={false}
      value={value}
      onChange={handleTextChange}
    />
  )
}

export default Field
``` 

Could you please provide an example of the correct way to add an initial value in the text in the editor, as well as the mechanism to update the text in the editor. Thank you.

JoelWakefield:
I've also been looking into this - I wanted to add a character counter below to allow my team to keep track of character count without being strictly limited by a max or min. 

It's not a big feature, but it's a small quality of life improvement that would allow them to avoid jumping from character-counting tool to contentful.

JoelWakefield:
@vitalbone, I found something which might be helpful. 

While I couldn't get the `RichTextEditor`'s `onChange` function to fire, there is a `ConnectedRichTextEditor` exported from `'@contentful/field-editor-rich-text'` as well - it also uses the same props (minus the `isInitiallyDisabled` prop)! 

The `onChange` function is fired for this component will fire; here's a small example of why I used it for:

```
import { useEffect, useState } from 'react';
[...]

PR #1740, opened 2024-09-03
fix: broken rich text onChange prop

MayaGillilan:
Closes https://github.com/contentful/field-editors/issues/1236

### DIFF
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -113,10 +113,10 @@ const RichTextEditor = (props: RichTextProps) => {
     []
   );
   React.useEffect(() => {
-    if (onChange) {
-      return sdk.field.onValueChanged(onChange);
+    if (!onChange) {
+      return;
     }
-    return;
+    return sdk.field.onValueChanged(onChange);
   }, [onChange, sdk.field]);
 
   const id = getContentfulEditorId(props.sdk);
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -20,20 +20,36 @@ import { SyncEditorChanges } from './SyncEditorChanges';
 import Toolbar from './Toolbar';
 import StickyToolbarWrapper from './Toolbar/components/StickyToolbarWrapper';
 
-type ConnectedProps = {
+type RichTextProps = {
   sdk: FieldAppSDK;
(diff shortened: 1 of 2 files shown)


packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -88,14 +104,19 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
...
+  const { sdk, isInitiallyDisabled, onAction, restrictedMarks, onChange, ...otherProps } = props;
   const isEmptyValue = React.useCallback(
     (value) => !value || deepEquals(value, Contentful.EMPTY_DOCUMENT),
     []
   );
+  // TODO: Deprecated use of onChange, should be removed in next major version
+  // To prevent breaking change, assigning onChange to SDK handler
+  React.useEffect(() => {
+    if (onChange) {
+      sdk.field.onValueChanged(onChange);
```
z0al, line 117: we should unsubscribe here on unmount

packages/rich-text/src/RichTextEditor.tsx
```diff
@@ -88,14 +104,19 @@ export const ConnectedRichTextEditor = (props: ConnectedProps) => {
...
-
-const RichTextEditor = (props: Props) => {
-  const { sdk, isInitiallyDisabled, onAction, restrictedMarks, ...otherProps } = props;
+const RichTextEditor = (props: RichTextProps) => {
+  const { sdk, isInitiallyDisabled, onAction, restrictedMarks, onChange, ...otherProps } = props;
   const isEmptyValue = React.useCallback(
     (value) => !value || deepEquals(value, Contentful.EMPTY_DOCUMENT),
     []
   );
+  // TODO: Deprecated use of onChange, should be removed in next major version
```
z0al, line 113: Users won't see this. You should move the notice next to the prop declaration and make it a JS doc e.g.

```typescript
/** 
* @deprecated use `sdk.field.onValueChanged` instead.
*/
onChange?: ...
```

packages/rich-text/src/RichTextEditor.tsx
z0al: `onValueChanged` returns an unsubscribe function already

MayaGillilan: True 😅  forgot about that

z0al (approved):
Thanks

PR #1756, opened 2024-09-30
chore(dependencies): bumping plate packages and react version to 18 [TOL-2415]

aodhagan-cf:
- Bumping plate packages
- Updating to react v18, required as the cypress component test webpack build fails without it after bumping some of the plate packages
- Fixing cypress component test parallelism


.circleci/config.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       - run:
           name: Run cypress component tests
           command: |
-            TESTFILES=$(circleci tests glob cypress/component/**/*.ts | circleci tests split --split-by=timings --timings-type=filename | tr '\n' ',')
+            TESTFILES=$(circleci tests glob cypress/component/**/*.ts | circleci tests split --split-by=timings --timings-type=filename | awk '{if (NR>1) printf ","; printf "%s", $0} END {if (NR>0) printf " "}')
```
aodhagan-cf, line 74: This is a known issue with cypress, if you end a CSV with a comma the cypress command ignores it and runs everything, we are actually running all tests but now in 3 containers instead of splitting

colomolo (approved):

.circleci/config.yml
chrishelgert: wonder if we have this issue in other projects too? (experience-packages, ..)

packages/_shared/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "peerDependencies": {
     "@contentful/app-sdk": "^4.29.0",
-    "react": ">=16.8.0"
+    "react": ">=17.0.2"
```
chrishelgert, line 50:
```suggestion
    "react": "16 | 17 | 18"
```

otherwise, we would drop react 16 support ;)

apps/entry-app-collapsible/package.json
```diff
@@ -13,8 +13,8 @@
     "@contentful/field-editor-single-line": "^0.15.6",
     "emotion": "^10.0.17",
     "prop-types": "^15.7.2",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "^18.3.1",
```
chrishelgert, line 16: Wonder if it should be a dependency or more a peer + devDependency?

packages/_test/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "peerDependencies": {
     "@contentful/app-sdk": "^4.29.0",
-    "react": ">=16.8.0"
+    "react": ">=17.0.2"
```
chrishelgert, line 49: here too :D  (and in the other ones too :D)

packages/_shared/package.json
aodhagan-cf: Thanks, good spot!

packages/rich-text/package.json
```diff
@@ -72,16 +72,16 @@
...
+    "react-dom": ">=17.0.2"
   },
   "devDependencies": {
     "@contentful/field-editor-test-utils": "^1.5.1",
     "@contentful/rich-text-react-renderer": "^15.16.4",
     "@types/is-hotkey": "^0.1.6",
     "@udecode/plate-test-utils": "^3.2.0",
     "prism-react-renderer": "2.4.0",
-    "react": ">=16.14.0"
+    "react": "17.0.2"
```
chrishelgert, line 84: dev should be react 18 right?

.circleci/config.yml
aodhagan-cf: experience-packages is not parallelised but yeah it's possible with the combo of test field splitting in circleci

chrishelgert (approved):

PR #1761, opened 2024-10-02
chore(dependencies): bumping forma everywhere []

aodhagan-cf:
Bumping forma packages


package.json
```diff
@@ -48,8 +48,8 @@
     "@babel/preset-react": "7.13.13",
     "@commitlint/cli": "19.3.0",
     "@commitlint/config-conventional": "18.4.4",
-    "@contentful/f36-components": "^4.62.0",
-    "@contentful/f36-tokens": "4.0.1",
+    "@contentful/f36-components": "^4.70.0",
+    "@contentful/f36-tokens": "4.0.5",
```
YvesRijckaert, line 52: any reason why we have this one pinned?

YvesRijckaert (approved):
LGTM

package.json
aodhagan-cf: I think in general it's better to pin devDependencies, I think f36 components probably should also be pinned since all other dev dependencies are pinned here

PR #1769, opened 2024-10-08
fix: use proper methods to fetch NER entities

veu:
Use proper types and methods from contentful-management for fetching [NER](https://www.contentful.com/help/orchestration/native-external-references/) entities.

Removing the default props in 49be1667ebce35dab86b31e46549d50779f5c4a2 is technically unrelated but it made sense to clean up while I was at it.


2wce (approved):
LGTM

PR #1773, opened 2024-10-10
feat: implement localized entity status badge in references [TOL-2454]

YvesRijckaert:
## Description
- localised entity status popover badge for reference cards (asset and entry)
- remove 'default' badge for references that are not localised. We don't need this anymore since here we will also show the stacked badge

## Approach

- We pass `privateLocales` and `activeLocales` to the widget renderer using the widget API sdk props (internal).
- We also moved the `useAsyncLocalePublishStatus` fn to the `shared` package, so we can use it in the webapp and in the `reference` package to determine the locale status for each reference

#### Single reference
Screenshot 2024-10-10 at 14 13 17

#### Multi reference
Screenshot 2024-10-10 at 14 13 58

#### Asset link
Screenshot 2024-10-10 at 14 46 12

#### Asset Gallery
Screenshot 2024-10-10 at 14 46 20

### DIFF
--- a/packages/reference/src/assets/WrappedAssetCard/WrappedAssetLink.tsx
+++ b/packages/reference/src/assets/WrappedAssetCard/WrappedAssetLink.tsx
@@ -1,7 +1,13 @@
 import * as React from 'react';
 
 import { EntryCard } from '@contentful/f36-components';
-import { entityHelpers, isValidImage, SpaceAPI } from '@contentful/field-editor-shared';
+import {
+  entityHelpers,
+  isValidImage,
+  LocalePublishStatusMap,
+  SpaceAPI,
+} from '@contentful/field-editor-shared';
+import { LocaleProps } from 'contentful-management';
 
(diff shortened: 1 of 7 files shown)


packages/reference/src/assets/WrappedAssetCard/WrappedAssetLink.tsx
```diff
@@ -58,6 +66,10 @@ export const WrappedAssetLink = (props: WrappedAssetLinkProps) => {
           entityType="Asset"
           entityId={props.asset.sys.id}
           status={status}
+          useLocalizedEntityStatus={props.useLocalizedEntityStatus}
+          entity={props.asset}
```
aodhagan-cf, line 70: Seems redundant to pass the entity and the entityId here. Maybe remove the entityId as a prop?

YvesRijckaert: Ah good catch, will remove it 👍

aodhagan-cf (approved):

PR #1774, opened 2024-10-10
feat: bring useAsyncLocalePublishStatus function [TOL-2454]

YvesRijckaert:
Bring in `useAsyncLocalePublishStatus` that can calculate the locale publishing status map for an entity. We will use this hook in the webapp and in other packages in field editors. Later on we might also use it in the entity search package.

⚠️ This is mostly copy-paste code from webapp


aodhagan-cf (approved):
LGTM

PR #1782, opened 2024-10-17
feat(Rich Text Editor): add sticky offset property

cf-remylenoir:
Having the sticky entity editor tabs in the web app requires having a customizable offset for the Rich Text Editor sticky toolbar to ensure it stays visible when scrolling and not hidden behind the tabs.

**Before**

The toolbar is sticky behind the tabs.

https://github.com/user-attachments/assets/0935d43e-8582-42ab-bf84-2e0965126f82

**After**

Adding the offset value matching the tabs' height ensures the toolbar stays visible.

https://github.com/user-attachments/assets/b190189e-bc10-4da7-8b13-30bd5f49eb09

### DIFF
--- a/packages/rich-text/src/RichTextEditor.tsx
+++ b/packages/rich-text/src/RichTextEditor.tsx
@@ -49,6 +49,7 @@ type ConnectedRichTextProps = {
   isDisabled?: boolean;
   isToolbarHidden?: boolean;
   actionsDisabled?: boolean;
+  stickyToolbarOffset?: number;
 };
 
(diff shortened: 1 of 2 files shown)


YvesRijckaert (approved):

DanweDE:
@cf-remylenoir Nit, but to my eyes it feels like we're off by one pixel as we have both horizontal lines on top now :thinking: 
image

cf-remylenoir:
Yes, I am controlling the offset value in the web app and will bring it 1px less 🙂 

> @cf-remylenoir Nit, but to my eyes it feels like we're off by one pixel as we have both horizontal lines on top now 🤔 image

PR #1791, opened 2024-10-29
feat(rich-text): bump validation and renderer to latest version []

chrishelgert:
- Remove copied validation and use the one from rich-text-types now
- Bump renderer packages to benefit from size reduction


YvesRijckaert (approved):
nice 🥳

PR #1792, opened 2024-11-04
feat(shared): remove the "async" useAsyncLocalePublishStatus

z0al:


packages/_shared/src/hooks/useLocalePublishStatus.ts
```diff
@@ -37,15 +42,15 @@ function getLocalePublishStatusMap(entity: AssetProps | EntryProps, locales: Loc
 /**
  * Get the publish status for each locale
  */
-export function useAsyncLocalePublishStatus(
+export function useLocalePublishStatus(
```
chrishelgert, line 45: This is also used where we don't have the localesAPI (entry list, entry editor), they would no longer work would they?

chrishelgert: Can we support both, providing a list of Locale[] or LocaleCodes[] and the Locales API?

z0al: Ah, so we want to keep both variants?

chrishelgert: Yeah I think we have to yes
This are all the places where it's used in user_interface and we don't have the LocalesAPI everywhere
https://github.com/search?q=repo:contentful/user_interface%20useAsyncLocalePublishStatus&type=code

z0al: > providing a list of Locale[] or LocaleCodes[] and the Locales API?

Why LocaleCodes[]?

z0al: Reverted back support for `LocaleProps[]`

chrishelgert: > Why LocaleCodes[]?
As only the `locale.code` would be needed maybe that's more a simpler then 🤷

chrishelgert (approved):
Thanks for taking care of this

packages/_shared/src/hooks/useLocalePublishStatus.ts
z0al: Not really, I don't know why but our status map is returned with `status` & `locale` as a body where the locale needs to at least have a `code`, `default` and `name` which is later used to render the status badge. 

I don't like it but I'm also not too familiar with where these things are used so didn't want to explode the scope. Would probably create a separate refactoring ticket, wdyt?

PR #1796, opened 2024-11-05
feat(shared): show stacked badge when status diff [TOL-2502]

YvesRijckaert:
Exports a util function to calculate if the statuses differ. Controlled by `shouldRetainLocaleHistory`

### DIFF
--- a/packages/_shared/src/utils/entityHelpers.test.ts
+++ b/packages/_shared/src/utils/entityHelpers.test.ts
@@ -1,4 +1,6 @@
-import { getEntityStatus, type EntitySys } from './entityHelpers';
+import { LocalePublishStatus } from 'hooks/useLocalePublishStatus';
+
+import { getEntityStatus, hasDifferentLocaleStatuses, type EntitySys } from './entityHelpers';
 
 describe('getEntityStatus', () => {
   function createEntity(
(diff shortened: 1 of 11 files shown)


packages/_shared/src/utils/entityHelpers.test.ts
```diff
@@ -175,4 +177,26 @@ describe('getEntityStatus', () => {
...
+
+  describe('hasDifferentLocaleStatuses', () => {
+    test('returns false if all locales have the same status', () => {
+      const statusMap = new Map<string, LocalePublishStatus>([
+        ['en-US', { status: 'published', locale: { code: 'en-US' } as any }],
+        ['fr-FR', { status: 'published', locale: { code: 'fr-FR' } as any }],
+        ['de-DE', { status: 'published', locale: { code: 'de-DE' } as any }],
+      ]);
+
+      expect(hasDifferentLocaleStatuses(statusMap)).toBe(false);
```
chrishelgert, line 189: NIT jest has fancy utils for false and true

`toBeTruthy()` and `toBeFalsy()`

packages/reference/src/components/EntityStatusBadge/EntityStatusBadge.tsx
```diff
@@ -34,7 +38,12 @@ export function EntityStatusBadge({
     getEntityScheduledActions,
   });
 
-  if (useLocalizedEntityStatus && activeLocales && localesStatusMap) {
+  const hasMultipleStatuses = useMemo(
+    () => entityHelpers.hasDifferentLocaleStatuses(localesStatusMap),
+    [localesStatusMap]
+  );
+
+  if (useLocalizedEntityStatus || hasMultipleStatuses) {
```
chrishelgert, line 46: wonder if we should have a second "feature flag" to enable the `hasMultipleStatuses` check, not that we confuse customers already with a new UX (in case the fieldStatus has different information than expected)

YvesRijckaert: Do we already have customers that have enabled and then disabled locale publishing though? The publishMode switch is still behind a FF right?

chrishelgert: I think we saw already something in the past when we rolled the `getEntryStatus` out without a feature flag, that the status did change for some customers -  for example if a locale was added after the entry was published

chrishelgert (approved):
🥇

PR #1803, opened 2024-11-25
feat: show the error and function tab link in the entity card [FUS-97]

Seth-Carter:
This PR adds error handling for cases when external resources retrieved via function invocation result in an error.

The error message also presents a link to the relevant function invocation log page in the web app so users can debug any issues caused by errors inside their functions.

image


Seth-Carter:
@contentful/team-tolkien Are there flakey component tests, or have I actually somehow changed rich text rendering despite not touching any of that code?

chrishelgert:
> @contentful/team-tolkien Are there flakey component tests, or have I actually somehow changed rich text rendering despite not touching any of that code?

Mh I saw the Richtext.HR text failing sometimes, so should be flacky :/

YvesRijckaert:
> > @contentful/team-tolkien Are there flakey component tests, or have I actually somehow changed rich text rendering despite not touching any of that code?
> 
> Mh I saw the Richtext.HR text failing sometimes, so should be flacky :/

We have a task for it in the backlog: https://contentful.atlassian.net/browse/TOL-2583

Seth-Carter:
> > > @contentful/team-tolkien Are there flakey component tests, or have I actually somehow changed rich text rendering despite not touching any of that code?
> > 
> > 
> > Mh I saw the Richtext.HR text failing sometimes, so should be flacky :/
> 
> We have a task for it in the backlog: https://contentful.atlassian.net/browse/TOL-2583

Thanks for confirming that I'm not insane (at least not when it comes to this PR).

packages/reference/src/components/ResourceEntityErrorCard/FunctionInvocationErrorCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+        function logs
+      </TextLink>
+    </Text>
+  );
+
+  return (
+    <Flex justifyContent="left" alignItems="center">
+      <ErrorCircleOutlineIcon variant="negative" />
+      <Text fontColor="colorNegative">&nbsp;Function invocation error.</Text>
+      {props.status === 'success' && props.functionId && children}
```
Seth-Carter, line 34: This is a potential edge case where we can't get the function id to successfully crate the function log link.

packages/reference/src/components/ResourceEntityErrorCard/ResourceEntityErrorCard.tsx
```diff
@@ -14,11 +15,24 @@ type ResourceEntityErrorCardProps = {
 };
 
 export function ResourceEntityErrorCard(props: ResourceEntityErrorCardProps) {
+  const providerName = getProviderName(props.linkType);
```
chrishelgert, line 18: we could move this after the `isUnsupportedError` check, as it's not used there :)

packages/reference/src/components/ResourceEntityErrorCard/FunctionInvocationErrorCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+type FunctionInvocationErrorCardProps = {
+  isSelected?: boolean;
+  isDisabled?: boolean;
+  organizationId: string;
+  appDefinitionId: string;
+  onRemove?: Function;
+  providerName?: string;
+};
+
+export function FunctionInvocationErrorCard(props: FunctionInvocationErrorCardProps) {
```
chrishelgert, line 48: NIT: we can destructure the props directly here, then the following code would get easier (no more `prop.`)
And you can use a default value for the `providerName`

packages/reference/src/components/ResourceEntityErrorCard/FunctionInvocationErrorCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+  organizationId: string;
+  appDefinitionId: string;
+  functionId?: string;
+  status: 'error' | 'success' | 'loading';
+};
+
+function FunctionInvocationErrorLogLink(props: FunctionInvocationErrorLogLinkProps) {
+  const functionLink = `/account/organizations/${props.organizationId}/apps/definitions/${props.appDefinitionId}/functions/${props.functionId}/logs`;
+
+  const children = (
```
chrishelgert, line 21: As it's only used once, you can add it directly below ;)
Then it's easier to follow the code 😅

packages/reference/src/common/EntityStore.tsx
```diff
@@ -204,7 +259,17 @@ async function fetchExternalResource({
...
+          .then(({ items }) => {
+            return items[0] ?? null;
+          })
+          .catch((e) => {
+            const parsedError = JSON.parse(e.message);
+            if (isFunctionInvocationErrorResponse(parsedError)) {
+              resourceFetchError = parsedError;
+            }
+
+            return null;
```
chrishelgert, line 271: Previously the error was not handled and therefore thrown to the parent, now we would hide maybe some errors, is this on purpose?

chrishelgert (commented):
Some tests for the new logic would be great :)

packages/reference/src/common/EntityStore.tsx
Seth-Carter: We can't rethrow the original error here because we need the `resourceTypes` call to finish in order to get the correct appDefinition id and organization id that is required to generate the function log link in the FunctionInvocationErrorCard component.

I considered making these two calls to the CMA client sequential, but that would cause a degradation in performance.

I could store the original error in the `resourceFetchError` variable declared in the outer scope and just rethrow it if it's not a `FunctionInvocationResponseError` after the CMA client calls are complete.

I'm not sure if that's much better.

Seth-Carter: I changed this to rethrow the original error if it's not related to function invocation.

packages/reference/src/resources/Cards/ResourceCard.spec.tsx
```diff
@@ -198,4 +213,14 @@ describe('ResourceCard', () => {
...
     ).toBeDefined();
   });
+
+  it('renders function invocation error card when an external resource request fails', async () => {
+    const { getByTestId } = renderResourceCard({
+      linkType: resolvableExternalResourceType,
+      entityUrn: unresolvableExternalEntityUrn,
+    });
+
+    await waitFor(() => expect(getByTestId('cf-ui-function-invocation-error-card')).toBeDefined());
```
Seth-Carter, line 223: I think this is how you're meant to test things in here?

packages/reference/src/common/EntityStore.tsx
```diff
@@ -92,8 +92,43 @@ type EntityQueryKey = [
...
+): response is FunctionInvocationErrorResponse {
+  const functionInvocationErrorMessages = [
+    'An error occurred while executing the Contentful Function code',
+    'Response payload of the Contentful Function is invalid',
+  ];
+  return (
+    response !== null &&
+    typeof response === 'object' &&
+    'message' in response &&
+    typeof response.message === 'string' &&
```
colomolo, line 127: Why not just:
```ts
return (typeof response?.message === 'string' && functionInvocationErrorMessages.includes(response.message));
```

packages/reference/src/common/EntityStore.tsx
```diff
@@ -108,6 +143,43 @@ export function isUnsupportedError(value: unknown): value is UnsupportedError {
...
+    this.isFunctionInvocationError = true;
+    this.organizationId = organizationId;
+    this.appDefinitionId = appDefinitionId;
+  }
+}
+
+export function isFunctionInvocationError(value: unknown): value is FunctionInvocationError {
+  return (
+    typeof value === 'object' &&
+    (value as FunctionInvocationError | null)?.isFunctionInvocationError === true
```
colomolo, line 161: Again, could this be simply:
```ts
return value?.isFunctionInvocationError === true;
```
?

chrishelgert (approved):

packages/reference/src/common/EntityStore.tsx
Seth-Carter: The `response` param was `unknown` in a previous version of this code, so I had to do all those checks. I'll change that to make it simpler now that we're doing an `instanceof Error` check before this function is called.

PR #1805, opened 2024-11-26
feat: single-line-editor add an optional prop to hide all validation messages [LUMOS-312]

ethan-ozelius-contentful:
### Description
Add an optional prop to render all validation messages.  This prop defaults to true, to ensure backward compatibility.

### Necessary use case
In studio, users need to be able to bind component properties to entry/asset fields.
- example 1: binding a heading to an entry title
- example 2: binding a background image to an asset file.
However, we do **not** allow users to make changes to the underlying entry/asset while they are in the process of binding (if they need to edit the underlying entry/asset, there is a separate form).  Therefore, it doesn't make sense for studio to render the validation messages when binding.

Screenshot 2024-11-26 at 2 42 07 PM

###  with validation messages enabled
Screenshot 2024-11-26 at 2 14 09 PM

### with withCharValidation = false
Screenshot 2024-11-26 at 2 13 46 PM

###  with validation messages disabled
Screenshot 2024-11-26 at 2 14 39 PM


packages/single-line/src/SingleLineEditor.tsx
```diff
@@ -16,17 +16,22 @@ export interface SingleLineEditorProps {
...
 
   /**
    * is the field manually disabled
    */
   isDisabled?: boolean;
 
+  /**
+   * whether validations should be rendered or not.
+   */
+  validationsAreVisable?: boolean;
```
chrishelgert, line 29: should we name it `withValidations`? to stay consistent in the naming

or `withCharInformation`? as the char counter would be no validation 🙈

chasepoirier: +1 to renaming to `withCharInformation` to match naming conventions of this component and it matches the purpose of the boolean a bit better since to Chris' point it's not just for validations but also the character counter.

packages/single-line/src/SingleLineEditor.tsx
```diff
@@ -84,25 +95,24 @@ export function SingleLineEditor(props: SingleLineEditorProps) {
...
+                    <CharCounter value={value || ''} checkConstraint={checkConstraint} />
+                    <CharValidation constraints={constraints} />
+                  </div>
+                )}
+                {withCharValidation === false && (
+                  <div className={styles.validationRow}>
+                    <CharCounter value={value || ''} checkConstraint={() => true} />
+                  </div>
+                )}
+              </>
```
chasepoirier, line 111: I think we can clean up these conditionals a bit to something like:

```tsx
// new const for checkConstraint above the return of the component
const handleCheckConstraint = withCharValidation ? checkConstraint : () => true

// updated return
{withCharInformation && (
    <div className={styles.validationRow}>
      <CharCounter value={value || ''} checkConstraint={handleCheckConstraint} />
      {withCharValidation && <CharValidation constraints={constraints} />}
    </div>
)}
```

ethan-ozelius-contentful: Nice, yeah this change definitely cleans the code up.

- [ ] clean up conditionals.

chasepoirier (approved):
Awesome, looks good!

packages/single-line/src/SingleLineEditor.test.tsx
```diff
@@ -229,4 +229,35 @@ describe('SingleLineEditor', () => {
     expect(getByText('0 characters')).toBeInTheDocument();
     expect(queryByText('Requires between 100 and 1000 characters')).not.toBeInTheDocument();
   });
+
+  it('renders no validation message or counter if validationsAreVisable is false', () => {
```
chrishelgert, line 233: NIT the test case is no longer correct after renaming the props 😅

chrishelgert (approved):

PR #1806, opened 2024-11-28
feat(date): allow arbitrary date range [ZEND-5700]

z0al:
In a follow up we will consider respecting the content model validation settings


aodhagan-cf (approved):
LGTM

nit: maybe feels more like a feature than a fix

PR #1814, opened 2024-12-04
fix: rerender rich text when active locales change [TOL-2606]

YvesRijckaert:
Rerender rich text editor when active locales change

### DIFF
--- a/packages/rich-text/src/SdkProvider.tsx
+++ b/packages/rich-text/src/SdkProvider.tsx
@@ -8,7 +8,7 @@ interface SdkProviderProps {
 }
 
 function useSdk({ sdk }: SdkProviderProps) {
-  const sdkMemo = React.useMemo<FieldAppSDK>(() => sdk, []); // eslint-disable-line -- TODO: explain this disable
+  const sdkMemo = React.useMemo<FieldAppSDK>(() => sdk, [sdk.parameters.instance.activeLocales]); // eslint-disable-line -- TODO: explain this disable
 
   return sdkMemo;
 }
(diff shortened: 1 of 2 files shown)


packages/rich-text/src/SdkProvider.tsx
```diff
@@ -8,7 +8,7 @@ interface SdkProviderProps {
 }
 
 function useSdk({ sdk }: SdkProviderProps) {
-  const sdkMemo = React.useMemo<FieldAppSDK>(() => sdk, []); // eslint-disable-line -- TODO: explain this disable
+  const sdkMemo = React.useMemo<FieldAppSDK>(() => sdk, [sdk.parameters.instance.activeLocales]); // eslint-disable-line -- TODO: explain this disable
```
chrishelgert, line 11: are we still providing this? 🤔 
I thought it was removed in favor of the localesAPI

YvesRijckaert: No we didn't go with the localesAPI. We only made a change in the `EntitySelectorSDK` to pass the editor locale props and then rely on the `getLocaleSettings` method to calculate the active locales. That was only relevant for the entity selector.
This PR is about the rich text editor. The editor was not updating when the active locales change because the SDK is memoized. 

But maybe we could take a look at removing this value from the instance param as well? 🤔

chrishelgert: I think we no longer provide it in the user_interface 🤔

chrishelgert (approved):

PR #1818, opened 2025-01-08
feat: permissions check for cross space reference editor [FUS-115]

marcolink:
# Description
Checks permission when rendering to the Reference Editor. This check exists on the service, but we also want to check for access in the frontend for better UX. The same logic exists for "normal" same space entries.

### DIFF
--- a/packages/reference/src/resources/SingleResourceReferenceEditor.tsx
+++ b/packages/reference/src/resources/SingleResourceReferenceEditor.tsx
@@ -4,6 +4,7 @@ import { FieldConnector } from '@contentful/field-editor-shared';
 
 import { EntityProvider } from '../common/EntityStore';
 import { ReferenceEditorProps } from '../common/ReferenceEditor';
+import { useEditorPermissions } from '../common/useEditorPermissions';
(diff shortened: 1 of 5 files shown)


packages/reference/src/resources/SingleResourceReferenceEditor.tsx
```diff
@@ -17,10 +18,18 @@ export function SingleResourceReferenceEditor(
   }
 ) {
   const { dialogs, field } = props.sdk;
+
+  const editorPermissions = useEditorPermissions({
```
chrishelgert, line 22: we could move `useEditorPermissions` inside of `useResourceLinkActions` as it's the same for single nad multiple resources :)

+ NIT: instead of passing all props down, just provide the necessary ones :)

marcolink: @chrishelgert I wasn't sure where else this is used, changed it 👍

marcolink:
I completely removed the `apiUrl` prop as this is not used anywhere

chrishelgert:
BTW do you plan to have the checks in RichText editor too?

Cyberxon (approved):
LGTM 🚀

PR #1821, opened 2025-01-15
fix(reference): infinite loop on permisson check for cross space references []

chrishelgert:


packages/reference/src/common/useContentTypePermissions.ts
```diff
@@ -56,12 +60,17 @@ export function useContentTypePermissions(
...
+      }
     }
 
-    void checkContentTypeAccess();
+    if (availableContentTypes.length > 0) {
+      void checkContentTypeAccess();
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: Evaluate the dependencies
-  }, [availableContentTypes]);
+  }, [availableContentTypes, creatableContentTypes]);
```
Lelith, line 73: I think there is a Typo and the var should be called createableContentTypes ? Or was that there before

chrishelgert: it's everywhere in this file without `e` after creat 🙈

chrishelgert: let me fix the typos :D

chrishelgert: we have the same working everywhere :/ so maybe not working fixing it now

Lelith: 💔 i see, then we will live with it!

Lelith (approved):

PR #1822, opened 2025-01-15
feat(field-dialog): filter by locale [FUS-199]

BasKiers:
Adds `locale` filter for reference search

### DIFF
--- a/packages/reference/src/common/EntityStore.tsx
+++ b/packages/reference/src/common/EntityStore.tsx
@@ -461,7 +461,7 @@ const [InternalServiceProvider, useFetch, useEntityLoader, useCurrentIds] = cons
         urn: string,
         options?: GetResourceOptions
       ): QueryResourceResult<R> {
-        const queryKey: ResourceQueryKey = ['Resource', resourceType, urn, locale];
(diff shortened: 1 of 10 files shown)


packages/reference/src/common/EntityStore.tsx
```diff
@@ -188,7 +188,12 @@ const isEntityQueryKey = (queryKey: QueryKey): queryKey is EntityQueryKey => {
   );
 };
 
-type ResourceQueryKey = [ident: 'Resource', resourceType: string, urn: string];
+type ResourceQueryKey = [
+  ident: 'Resource',
+  resourceType: string,
+  urn: string,
+  locale: string | undefined
```
Cyberxon, line 195: Was this done to make sure we always have the locale in the key? Wouldn't it be better to have `locale?: string` ? Or do we need a fixed shape for the key?

Cyberxon (approved):
LGTM, one minor question.

packages/reference/src/common/EntityStore.tsx
marcolink: I think this is required to make the typed array eventually extendable 🤔

PR #1825, opened 2025-01-20
test: [] Fix flaky tests

colomolo:
Attempt to fix flaky RTE tests.

### DIFF
--- a/cypress/component/rich-text/RichTextEditor.HR.spec.ts
+++ b/cypress/component/rich-text/RichTextEditor.HR.spec.ts
@@ -3,7 +3,7 @@ import { BLOCKS } from '@contentful/rich-text-types';
 import { block, document as doc, text } from '../../../packages/rich-text/src/helpers/nodeFactory';
 import { emptyParagraph, paragraphWithText } from './helpers';
 import { RichTextPage } from './RichTextPage';
-import { mountRichTextEditor } from './utils';
+import { focusEditorAndType, mountRichTextEditor } from './utils';
 
 // the sticky toolbar gets in the way of some of the tests, therefore
 // we increase the viewport height to fit the whole page on the screen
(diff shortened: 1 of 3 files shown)


cypress/component/rich-text/RichTextEditor.HR.spec.ts
```diff
@@ -13,9 +13,12 @@ describe('Rich Text Editor - HR', { viewportHeight: 2000, viewportWidth: 1000 },
   const expectDocumentToBeEmpty = () => richText.expectValue(undefined);
 
   function addBlockquote(content = '') {
-    richText.editor.click().type(content);
+    richText.editor.click();
+    richText.editor.should('be.focused');
```
MayaGillilan, line 17: What do you think about adding a helper function that does these three steps and you just pass in the text to be typed?

cypress/component/rich-text/RichTextEditor.HR.spec.ts
```diff
@@ -39,7 +42,8 @@ describe('Rich Text Editor - HR', { viewportHeight: 2000, viewportWidth: 1000 },
     });
 
     it('should add a new line when clicking', () => {
-      richText.editor.click().type('some text');
+      richText.editor.click();
+      richText.editor.type('some text');
```
MayaGillilan, line 46: Maybe let's add the should be focused to these other areas too

cypress/component/rich-text/RichTextEditor.HR.spec.ts
MayaGillilan: 🤔 fair point, although I think being focused is a natural assumption if you're typing. Maybe something like "focusAndType"?

MayaGillilan: We may also want to do something like this for markdown and JSON editors, not sure if those have been flakey though

colomolo: Yep, actually, used almost the same name before reading your comment 😀 Also, added few (potentially) redundant checks, but should help to make it more stable.

MayaGillilan: I feel like with cypress you can never have too much redundancy :D

MayaGillilan (approved):
LGTM, only thing is I'm not sure we need to check focus again after the focusAndType? but extra redundancy also doesn't hurt so maybe it's safer

PR #1828, opened 2025-02-03
chore: use updated types from contentful-management

MayaGillilan:


packages/reference/src/__fixtures__/FakeSdk.ts
```diff
@@ -72,13 +78,13 @@ export function newReferenceEditorFakeSdk(props?: ReferenceEditorSdkProps): [Fie
             await delay(props.fetchDelay);
           }
           if (entryId === entries.empty.sys.id) {
-            return entries.empty;
+            return entries.empty as EntryProps;
```
z0al, line 81: why not as this in the entries fixture file so it's set once?

packages/reference/src/__fixtures__/FakeSdk.ts
```diff
@@ -89,13 +95,13 @@ export function newReferenceEditorFakeSdk(props?: ReferenceEditorSdkProps): [Fie
             await delay(props.fetchDelay);
           }
           if (assetId === assets.empty.sys.id) {
-            return assets.empty;
+            return assets.empty as unknown as AssetProps;
```
z0al, line 98: ditto

z0al (approved):

PR #1838, opened 2025-03-07
feat: use useActiveLocales hook instead of instance params [TOL-2887]

YvesRijckaert:
- Extract `useActiveLocales` from experience packages to field editors shared, and change the type so it can accept both `EntitySelectorExtensionSDK` and `FieldAppSDK`
- Use this new `useActiveLocales` hook to determine the active locales instead of passing it as a instance param in field editors from the webapp

Later after this PR is merged: Use the now extracted `useActiveLocales` from field editors shared in experience-packages entity-search

## Testing:
Screenshot 2025-03-07 at 14 07 02

### DIFF
--- a/packages/_shared/src/hooks/useActiveLocales.tsx
+++ b/packages/_shared/src/hooks/useActiveLocales.tsx
@@ -1,0 +1,56 @@
+import { useEffect, useState } from 'react';
+
+import { BaseAppSDK, SharedEditorSDK } from '@contentful/app-sdk';
+
+type LocaleSettings = {
+  mode: 'single' | 'multi';
+  active?: string[];
+  focused?: string;
+};
+
+export type SDK = {
+  editor?: SharedEditorSDK['editor'];
+  locales: Pick<BaseAppSDK['locales'], 'available'>;
+};
+
+export function useActiveLocales(sdk: SDK) {
+  const [activeLocales, setActiveLocales] = useState<Array<{ code: string }>>([]);
+
+  useEffect(() => {
+    // Set all available locales if editor is not available
+    if (!sdk.editor) {
+      setActiveLocales(sdk.locales.available.map((code) => ({ code })));
+      return;
+    }
+
+    const availableLocales = new Set(sdk.locales.available);
+
+    const updateLocales = (settings: LocaleSettings) => {
+      let localeCodes: string[] = [];
+
+      if (settings.mode === 'multi' && settings.active) {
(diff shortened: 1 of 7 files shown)


packages/rich-text/src/SdkProvider.tsx
```diff
@@ -8,7 +8,7 @@ interface SdkProviderProps {
 }
 
 function useSdk({ sdk }: SdkProviderProps) {
-  const sdkMemo = React.useMemo<FieldAppSDK>(() => sdk, [sdk.parameters.instance.activeLocales]); // eslint-disable-line -- TODO: explain this disable
```
YvesRijckaert, line 11: This line was necessary to rerender the rich text editor when the activeLocales (previously passed as instance params) would change. However now that we calculate them based on the `useActiveLocales` we don't seem to need this anymore

packages/rich-text/src/SdkProvider.tsx
```diff
@@ -8,7 +8,7 @@ interface SdkProviderProps {
...
+  const sdkMemo = React.useMemo<FieldAppSDK>(() => sdk, []); // eslint-disable-line -- TODO: explain this disable
```
aodhagan-cf, line 11: `TODO: explain this disable` I guess the reason for this is we don't want to trigger a load of rerenders and that it's okay to just create the memo when this hook gets mounted and the side effects are managed by the even listeners in each of the sdk properties themselves

aodhagan-cf (approved):
Looks good to me 🚀

packages/rich-text/src/SdkProvider.tsx
aodhagan-cf: I guess the new hook now deals with the rerenders instead

packages/_shared/src/hooks/useActiveLocales.tsx
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useState } from 'react';
+
+import { BaseAppSDK, SharedEditorSDK } from '@contentful/app-sdk';
+
+type LocaleSettings = {
+  mode: 'single' | 'multi';
+  active?: string[];
+  focused?: string;
```
chrishelgert, line 8: We could ignore single and focused here, as with the new translation widget is stays consistent :)

chrishelgert (approved):

z0al:
@YvesRijckaert `refactor:` is unlikely to trigger a release which you need here

z0al (approved):

PR #1840, opened 2025-03-10
fix: dnd embedded entities in firefox [TOL-2460]

aodhagan-cf:
Adding drag handle to embedded entities, if drag handle is provided then adding `draggable` to the `EntryCard` or `AssetCard` respectively, this makes them draggable in firefox.


YvesRijckaert (approved):
LGTM! Thank you

PR #1849, opened 2025-03-28
feat: update markdown editor force tab to open [BBEE-1138]

seanf-ctfl:
In this PR I update the `MarkdownEditor` component.

I add a new `forceTab` optional parameter. 

When given, this will force a specific tab to open and not allow the other tab to open. The disabled tab will also show the `not-allowed` cursor.

I am adding this, as the UI needed to for the '_editor_' tab to open when the AI Action is loading.

https://contentful.atlassian.net/browse/BBEE-1138

### DIFF
--- a/packages/markdown/src/MarkdownEditor.tsx
+++ b/packages/markdown/src/MarkdownEditor.tsx
@@ -43,6 +43,8 @@ export interface MarkdownEditorProps {
   previewComponents?: PreviewComponents;
   onReady?: Function;
   isDisabled?: boolean;
+  /** Forces a specific tab to be active, and disables the other */
+  forceTab?: MarkdownTab;
 }
 
 export function MarkdownEditor(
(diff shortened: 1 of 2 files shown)


chrishelgert:
Would it make sense to provide a configuration `enabledTabs` and the first one of this would be selected and the other ones not defined are disabled

Could be a little bit more explanatory instead of `forceTab`? 🤔

packages/markdown/src/MarkdownEditor.tsx
```diff
@@ -107,10 +109,12 @@ export function MarkdownEditor(
   return (
     <div className={styles.container} data-test-id="markdown-editor">
       <MarkdownTabs
-        active={selectedTab}
+        active={props.enableTab || selectedTab}
         onSelect={(tab) => {
+          if (props.enableTab) return;
           setSelectedTab(tab);
         }}
+        forceTab={props.enableTab}
```
seanf-ctfl, line 117:
```suggestion
        enableTab={props.enableTab}
```

chrishelgert (approved):
💅

PR #1856, opened 2025-04-22
feat: Improve Accessibility of field editors by enabling associating them with labels [UFO-1737]

Lelith:
### Problem:

our field editors did not accept or forward an ID prop, which is a anti pattern for accessibility. Associating a label with an input field enables autofocusing of the associated input field when selecting / clicking on the label. 

### Risks

fields could get, on accident, the same ID (when e.g. using an array index for id) which worsens the accessibility and violates HTML5 rules.

### DIFF
--- a/packages/boolean/stories/BooleanEditor.stories.tsx
+++ b/packages/boolean/stories/BooleanEditor.stories.tsx
@@ -1,18 +1,19 @@
 import * as React from 'react';
 
+import { FormControl } from '@contentful/f36-components';
 import { ActionsPlayground, createFakeFieldAPI } from '@contentful/field-editor-test-utils';
 import type { Meta, StoryObj } from '@storybook/react';
 
-import * as Editor from '../src/BooleanEditor';
+import { BooleanEditor } from '../src/BooleanEditor';
 
-const meta: Meta<typeof Editor.BooleanEditor> = {
+const meta: Meta<typeof BooleanEditor> = {
   title: 'editors/Boolean',
-  component: Editor.BooleanEditor,
+  component: BooleanEditor,
 };
 
 export default meta;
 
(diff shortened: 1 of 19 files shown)


MayaGillilan (dismissed):

packages/multiple-line/src/MultipleLineEditor.tsx
cf-remylenoir, line 1: Let's add the JSDoc to the addition of the `id` prop, wdyt?

packages/boolean/stories/BooleanEditor.stories.tsx
cf-remylenoir, line 1: I was trying to get my hands on the Storybook deployment with these changes, but it is not setup to have it deployed, only when the changes are merged to the `master` branch 😬

cf-remylenoir: Ahhh nope, here it is https://deploy-preview-1856--contentful-field-editors.netlify.app/

cdun (commented):
Are there any other editors that we should think about supporting before merging? Some of them might require more thought (e.g. `RichText` being contenteditable rather than an input is more complex), but maybe `List`, `Slug`, `Tags` etc?

Lelith:
@cdun 
I added it for Slug and Url editors as well. The Tag and List editor slipped me, i will add them as well :) thanks for the reminder

About the RichTextEditor, as far as i understood the WCAG recommendation, we should focus on inputs and textareas, so i would say we can leave it out for now. But when we hire the Accessibility Auditor i am sure we will talk about it again and they maybe have a recommendation on how we can make it more acceessible.

cdun (approved):

MayaGillilan (approved):

PR #1869, opened 2025-05-20
fix(slug-editor): [TOL-3065] Do not autogenerate slug if user has no access

colomolo:
Pass down access param to prevent slug autogeneration in cases when the user doesn't have permissions to edit the field.

### DIFF
--- a/packages/slug/src/SlugEditor.tsx
+++ b/packages/slug/src/SlugEditor.tsx
@@ -29,6 +29,8 @@ export interface SlugEditorProps {
       trackingFieldId?: string;
     };
   };
+
+  access?: boolean;
 }
 
 function isSupportedFieldTypes(val: string): val is 'Symbol' {
@@ -47,6 +49,7 @@ function FieldConnectorCallback({
   createdAt,
   performUniqueCheck,
   id,
+  access = true,
 }: {
   Component: typeof SlugEditorFieldStatic | typeof SlugEditorField;
   value: string | null | undefined;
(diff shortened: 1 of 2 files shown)


z0al (changes requested):
Questions:
- Is product okay with this?
- "access" is vague. It should be clear that this is edit access
- Why not use `sdk.access`?

colomolo:
> * Is product okay with this?

Okay/not okay with not showing fake value?

> * "access" is vague. It should be clear that this is edit access
> * Why not use `sdk.access`?

We pass this prop around in other field editors, so for consistency:
https://github.com/contentful/user_interface/blob/master/src/javascripts/widgets/BuiltinWidgets.tsx#L660

It comes from here:
https://github.com/contentful/user_interface/blob/master/src/javascripts/app/entity_editor/EntityField/EntityFieldControl.tsx#L167

YvesRijckaert (approved):

z0al:
> Okay/not okay with not showing fake value?

Ok with slug generation not working without a clear indicator for the "why".

> We pass this prop around in other field editors, so for consistency

@colomolo I don't think that is a good example to follow. All field editors already have access to `sdk.access`. If the `access` attribute is supposed to be the new standard prop then it should be done in the [FieldConnector](https://github.com/contentful/field-editors/blob/80efc635cc467c4505b1cfe0397b3a8246040ebe/packages/_shared/src/FieldConnector.ts#L150-L156) where we provide common props to all field editors.

PR #1872, opened 2025-05-30
feat: enable read-only assets and reference cards [SPA-2702]

kizer-cf:
This PR modifies several reference cards to allow rendering truly non-interactive cards. This is a prerequisite for [SPA-2702](https://contentful.atlassian.net/browse/SPA-2702)

[SPA-2702]: https://contentful.atlassian.net/browse/SPA-2702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
+++ b/packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
@@ -12,7 +12,7 @@ import { useEntity, useEntityLoader } from '../../common/EntityStore';
 import { LinkActionsProps, MissingAssetCard } from '../../components';
 import { Action, Asset, FieldAppSDK, ViewType, RenderDragFn } from '../../types';
 import { WrappedAssetCard, WrappedAssetCardProps } from './WrappedAssetCard';
-import { WrappedAssetLink } from './WrappedAssetLink';
+import { WrappedAssetLink, WrappedAssetLinkProps } from './WrappedAssetLink';
 
 type FetchingWrappedAssetCardProps = {
   assetId: string;
@@ -32,7 +32,7 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
   const { getEntityScheduledActions } = useEntityLoader();
   const loadEntityScheduledActions = React.useCallback(
     () => getEntityScheduledActions('Asset', props.assetId),
-    [getEntityScheduledActions, props.assetId]
+    [getEntityScheduledActions, props.assetId],
   );
   const localesStatusMap = useLocalePublishStatus(asset, props.sdk.locales);
   const activeLocales = useActiveLocales(props.sdk);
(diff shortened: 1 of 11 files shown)


z0al:
@kizer-cf the PR introduces two changes:
1. Treats `sdk.navigator.*` as possibly undefined
2. Uses `isClickable` to determine whether to render the card as a link

My comment above is about (1). The second change is fine IMHO.

> For example, not checking that onEdit is set before calling it is causing actual errors in our codebase, and I would argue it's not reasonable to expect implementers to define a noop for it when it's marked as optional.

I would agree with that if `onEdit` is optional, but it's not? `onEdit` relies on `sdk.navigator` which isn't optional & typing of the prop also reflects that as you had to change it in the PR 🤔 

The SDK serves as a contract for not only field editors but also custom apps. All APIs that the editor uses should be present. We don't have any optional methods as per [docs](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/). It's the responsibility of the host app to initialize a proper SDK. In that sense, the noop implementation isn't a workaround, it's a requirement so that field editors don't need to double-check that every SDK method exists before calling it. 

Happy to jump on a call btw to discuss this

z0al:
@kizer-cf to give another perspective. While we have more flexibility with our own field editors. Many customers use custom field editor apps. Not passing a proper SDK e.g. `sdk.navigator.openAsset` is not implemented, is likely to crash their app, which isn't good. Using a noop implementation prevents that while still giving the host app control over the functionality

kizer-cf:
Thanks for clarifying on the SDK, it must be our error then for not providing a `navigator.openAsset` function, thus the errors. I could revert that part of my change.

However, we still need `isClickable` functionality, as there is no easy way for us to fix the accessibility and UX issue otherwise.

z0al:
> However, we still need isClickable functionality

@kizer-cf sounds reasonable. Happy to review that after reverting the sdk changes

kizer-cf:
@z0al I have made some minor changes to revert some of the changes I made to the way `sdk.navigator.openAsset` is assigned to `onEdit`. However, I have left the type definition changes to `onEdit` and `onRemove` intact, as these were throwing all sorts of errors otherwise. This is also consistent with how these were defined in other components (this file was the only exception before now). For example, 
https://github.com/contentful/field-editors/blob/b54d8502e85f24a8c97b0219b656d9c4f0703206/packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx#L37
https://github.com/contentful/field-editors/blob/83b9250c2be6ca3fa1df24fff4fd9ae316c4958b/packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx#L22

Also, every other place onEdit and onRemove were called, they were called conditionally, consistent with my changes. Examples:
https://github.com/contentful/field-editors/blob/83b9250c2be6ca3fa1df24fff4fd9ae316c4958b/packages/reference/src/resources/ExternalResourceCard/ExternalResourceCard.tsx#L119
https://github.com/contentful/field-editors/blob/b54d8502e85f24a8c97b0219b656d9c4f0703206/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx#L216
https://github.com/contentful/field-editors/blob/72600722cc012078f465211b3e8e6b46882176a4/packages/reference/src/assets/WrappedAssetCard/WrappedAssetLink.tsx#L89
https://github.com/contentful/field-editors/blob/7bd8433855f7dae63a6f4441ff97e5afc57e75b1/packages/reference/src/components/MissingAssetCard/MissingAssetCard.tsx#L26

packages/reference/src/assets/WrappedAssetCard/FetchingWrappedAssetCard.tsx
```diff
@@ -130,6 +141,14 @@ export function FetchingWrappedAssetCard(props: FetchingWrappedAssetCardProps) {
...
     }
 
+    const viewType = props.viewType;
+
+    function renderDefaultCard(
+      props?: Partial<CustomEntityCardProps> | Partial<WrappedAssetLinkProps>,
+    ) {
+      return viewType === 'link' ? _renderDefaultLink(props) : _renderDefaultCard(props);
+    }
+
```
z0al, line 138: Nit: it looks to me that `_renderDefaultLink` & `_renderDefaultLink` are not needed here and we could just inline the rendering of link vs card inside the `renderDefaultCard` something like:

```typescript
    function renderDefaultCard(props?: CustomEntityCardProps) {
      const cardProps = {
        ...commonProps,
        ....props,
        ...blabla
      };

      if (viewType === 'link') {
        return (
          <WrappedAssetLink
            {...cardProps}
            href={commonProps.entityUrl}
            getEntityScheduledActions={loadEntityScheduledActions}
          />
        );
      } else {
        return <WrappedAssetCard {...cardProps} getAssetUrl={...} />;
      }
    }
```

kizer-cf: Thanks @z0al. Made the sugegsted refactor. Will need your approval again.

z0al (approved):

PR #1875, opened 2025-06-04
feat: bring release status components [TOL-3103]

YvesRijckaert:
Migrate release status component from user interface to field editors _shared to be able to display them in reference and rich text package in follow up PRs

In a follow up ticket we could potentially avoid the duplication between `LocalePublishingEntityStatusBadge` and `TimelineEntityStatus`


chrishelgert (approved):
Maybe worth to have a look at the naming - is it ReleaseEntityStatus or TimelineEntityStatus?

in the types and constants we have Release and the components with Timeline - could be great to use the same pattern.

Would think it's more the status in the release, as the timeline is the grouping of multiple releases?

PR #1876, opened 2025-06-05
feat: release status in references [TOL-3103]

YvesRijckaert:
We want to display the latest release status in the reference status (e.g. "to publish", "to become draft", etc).
For this I migrated the `useActiveReleaseLocalesStatuses` hook to field editors shared, so it can be used by both the reference package, the rich text editor and inside the webapp.

As a temporary solution, we pass all release information through the widget SDK using the instance parameters. This will change later when we have decided on Release SDK context specifics

Blocked by: https://github.com/contentful/field-editors/pull/1875

Screenshot 2025-06-06 at 09 51 52


packages/_shared/src/hooks/useActiveReleaseLocalesStatuses.spec.ts
```diff
@@ -0,0 +1,305 @@
+import { renderHook } from '@testing-library/react';
+import type { CollectionProp, EntryProps, LocaleProps } from 'contentful-management';
+import type { ReleaseV2Entity, ReleaseV2EntityWithLocales, ReleaseV2Props } from 'types';
```
chrishelgert: shouldn't it be a relative path?

packages/_shared/src/hooks/useActiveReleaseLocalesStatuses.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+
+import { getPreviousReleaseEntryVersion } from '../utils/getPreviousReleaseEntryVersion';
+import { useActiveReleaseLocalesStatuses } from './useActiveReleaseLocalesStatuses';
+
+const buildEntryDraft = (status: 'draft' | 'published' | 'changed') =>
```
chrishelgert: buildEntryDraft with `published` status? 🤔 🙈

packages/_shared/src/hooks/useActiveReleaseLocalesStatuses.ts
```diff
@@ -0,0 +1,132 @@
...
+  releases,
+}: {
+  currentEntryDraft: EntryProps;
+  entryId: string;
+  releaseVersionMap: Map<string, Map<string, ReleaseAction>>;
+  locales: LocaleProps[];
+  activeRelease: ReleaseV2Props;
+  releases: CollectionProp<ReleaseV2Props>;
+}) => {
+  const { previousReleaseEntity } = getPreviousReleaseEntryVersion({
```
chrishelgert: should we wrapt it in `useMemo`?

packages/_shared/src/types.ts
```diff
@@ -1,3 +1,7 @@
+import { BadgeVariant } from '@contentful/f36-components';
+import { CollectionProp, ReleaseProps } from 'contentful-management/types';
+import { LocaleProps } from 'contentful-management/types';
```
chrishelgert:
```suggestion
import type { CollectionProp, ReleaseProps } from 'contentful-management/types';
import type { LocaleProps } from 'contentful-management/types';
```

packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -61,13 +66,35 @@ async function openEntry(
...
+    activeRelease,
+    isActiveReleaseLoading,
+    releases,
+  }: {
+    releaseVersionMap: Map<string, Map<string, ReleaseAction>>;
+    locales: LocaleProps[];
+    activeRelease: ReleaseV2Props;
+    isActiveReleaseLoading: boolean;
+    releases: CollectionProp<ReleaseV2Props>;
+  } = JSON.parse(props.sdk.parameters.instance.release);
```
chrishelgert: maybe worth to extract into an utils and trycatch the parsing?

chrishelgert (approved):
Added some tiny nit comments, otherwise looks good 👍

PR #1920, opened 2025-09-02
chore: update f36 icons to use Forma v5

missating:
This updates all Forma dependencies to Forma V5. it also updates all icons to the new v5 icons using the icon codemod

### DIFF
--- a/packages/rich-text/src/plugins/shared/EmbeddedInlineToolbarIcon.tsx
+++ b/packages/rich-text/src/plugins/shared/EmbeddedInlineToolbarIcon.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { FieldAppSDK } from '@contentful/app-sdk';
 import { Menu, Flex } from '@contentful/f36-components';
-import { EmbeddedEntryInlineIcon } from '@contentful/f36-icons';
+import { EmbeddedLineIcon } from '@contentful/f36-icons';
 import tokens from '@contentful/f36-tokens';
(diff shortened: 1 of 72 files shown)


packages/rich-text/stories/RichTextEditor.stories.tsx
```diff
@@ -23,6 +24,8 @@ import { assets, contentTypes, entries, locales, spaces } from '../src/__fixture
 import RichTextEditor from '../src/RichTextEditor';
 import { RichTextPreview } from './RichTextPreview';
 
+
```
colomolo, line 27: ✂️

packages/rich-text/src/plugins/shared/EmbeddedInlineToolbarIcon.tsx
```diff
@@ -66,8 +67,8 @@ export function EmbeddedInlineToolbarIcon({
       onClick={handleClick}
     >
       <Flex alignItems="center" flexDirection="row">
-        <EmbeddedEntryInlineIcon
-          variant="secondary"
+        <EmbeddedLineIcon
+          color={tokens.gray900}
```
colomolo, line 71: is it intentional?

missating: yes. Forma v5 now uses `colors` for `icons`:  https://github.com/contentful/forma-36/blob/main/MIGRATION.md#icon

PR #1921, opened 2025-09-03
chore: update forma []

YvesRijckaert:


Lelith:
@YvesRijckaert  is this a duplicate for what Nkoyo is already doing? https://github.com/contentful/field-editors/pull/1922

PR #1922, opened 2025-09-03
chore: update f36 components to use v5

missating:
This updates all Forma components to V5. it also updates all icons to the new v5 icons using the icon codemod

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,3 +102,21 @@ All field editor packages are open source software [licensed as MIT](./LICENSE).
 [prettier]: https://prettier.io/
 [eslint]: https://eslint.org/
 [lingui]: https://lingui.dev/
+
+## Internationalization
+
+The latest versions of field editors are internationalized with [lingui](https://lingui.dev/). Any app using the latest field editors needs to include the `@lingui/core` and `@lingui/message-utils` packages and initialize `lingui` like this before rendering field editors:
+
+```javascript
+import { i18n } from '@lingui/core';
+import { compileMessage } from '@lingui/message-utils/compileMessage';
+
+...
+
+i18n.setMessagesCompiler(compileMessage);
+i18n.activate('en-US');  // use desired locale code
+
+const App = () => {
+  ...
+};
+```
--- a/packages/default-field-editors/CHANGELOG.md
+++ b/packages/default-field-editors/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.8.2](https://github.com/contentful/field-editors/compare/@contentful/default-field-editors@1.8.1...@contentful/default-field-editors@1.8.2) (2025-09-05)
+
+**Note:** Version bump only for package @contentful/default-field-editors
+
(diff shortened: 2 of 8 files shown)


2wce (approved):

apps/entry-app-collapsible/src/FieldGroupsEditor.tsx
```diff
@@ -69,7 +71,7 @@ const SortableFieldList = SortableContainer(
         <SortableFieldItem groupId={groupId} key={`item-${field.id}`} index={index} field={field} />
       ))}
     </ul>
-  )
+  ),
```
Lelith: Have these files been prettified? there are a lot of added commas everywhere 👀

package.json
```diff
@@ -49,8 +49,9 @@
     "@babel/preset-react": "7.27.1",
     "@commitlint/cli": "19.8.0",
     "@commitlint/config-conventional": "18.4.4",
-    "@contentful/f36-components": "4.70.0",
-    "@contentful/f36-tokens": "4.2.0",
+    "@contentful/f36-components": "5.1.0",
+    "@contentful/f36-tokens": "5.1.0",
+    "@phosphor-icons/react": "2.1.8",
```
Lelith: Why the extra dependency here? the phosphor icons should come via the package itself

packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
```diff
@@ -95,7 +95,7 @@ export function renderAssetInfo(props: { entityFile: File }) {
         )}
       </dl>
     </Menu.Item>,
-  ];
+  ].map((item, i) => <Menu key={i}>{item}</Menu>);
```
Lelith, line 98: is this change on purpose or is it related to something else?

packages/rich-text/stories/RichTextEditor.stories.tsx
```diff
@@ -23,6 +24,8 @@ import { assets, contentTypes, entries, locales, spaces } from '../src/__fixture
 import RichTextEditor from '../src/RichTextEditor';
 import { RichTextPreview } from './RichTextPreview';
 
+
+
```
Lelith:
```suggestion
```

apps/entry-app-collapsible/src/FieldGroupsEditor.tsx
missating: Yes, but I'll run it again to check. it may also be from my settings.

package.json
missating: Yes, but we had a [cjs/esm](https://contentful.slack.com/archives/C961BF4RX/p1756811802159099) issue with Phosphor that [this](https://github.com/contentful/forma-36/pull/3170) should fix, when its merged. So I'll clean this after I can test it against that.

packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
missating: yes, it comes from Team Tolkien. it fixes a broken test when using the MenuItem without the Menu component which has been updated cc @2wce

PR #1925, opened 2025-09-05
fix: [] File type import

colomolo:
Fix import of the `File` type in shared package.


packages/_shared/src/index.tsx
```diff
@@ -30,7 +30,7 @@ export { entityHelpers };
 export { FieldConnector } from './FieldConnector';
 export { ModalDialogLauncher };
 export { PredefinedValuesError } from './PredefinedValuesError';
-export { Asset, Entry, File } from './typesEntity';
+export { Asset, Entry, type File } from './typesEntity';
```
chrishelgert:
```suggestion
export type { Asset, Entry, File } from './typesEntity';
```

PR #1928, opened 2025-09-08
feat: add asset support for release entity status calculation [TOL-3363]

MayaGillilan:
Adds asset support for release entity status calculation, makes status badges for assets in reference and rich text editors release-aware. Also fixes a localization error that was breaking rich text

### DIFF
--- a/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
@@ -78,6 +78,8 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
   const localesStatusMap = useLocalePublishStatus(entry, props.sdk.locales);
   const { releaseLocalesStatusMap } = useActiveReleaseLocalesStatuses({
     currentEntityDraft: entry,
+    entityId: props.entryId,
+    entityType: 'Entry',
     releaseVersionMap,
     locales,
     activeRelease,


packages/_shared/src/hooks/useActiveReleaseLocalesStatuses.ts
```diff
@@ -1,39 +1,44 @@
...
         activeRelease,
         releases,
       }).previousReleaseEntity,
-    [entryId, releaseVersionMap, activeRelease, releases],
+    [entityId, releaseVersionMap, activeRelease, releases],
   );
   const activeReleaseReleaseEntity = useMemo(
-    () => activeRelease?.entities.items.find((entity) => entity.entity.sys.id === entryId),
-    [activeRelease?.entities.items, entryId],
+    () => activeRelease?.entities.items.find((entity) => entity.entity.sys.id === entityId),
```
chrishelgert: we might also want to check for entity.sys.type (the same id could be theoretically used for entry and asset)

packages/_shared/src/utils/getPreviousReleaseEntity.ts
```diff
@@ -36,10 +36,10 @@ export function getPreviousReleaseEntryVersion({
   if (indexOfActive && indexOfActive > 0) {
     for (let i = indexOfActive - 1; i >= 0; i--) {
       const release = orderedScheduledReleases![i];
-      const action = releaseVersionMap.get(entryId)?.get(release.sys.id);
+      const action = releaseVersionMap.get(entityId)?.get(release.sys.id);
       if (action !== 'not-in-release') {
         previousRelease = release;
-        previousReleaseEntity = release.entities.items.find((e) => e.entity.sys.id === entryId);
+        previousReleaseEntity = release.entities.items.find((e) => e.entity.sys.id === entityId);
```
chrishelgert: see entity.sys.type

packages/_shared/src/utils/getPreviousReleaseEntity.ts
```diff
@@ -36,10 +36,10 @@ export function getPreviousReleaseEntryVersion({
   if (indexOfActive && indexOfActive > 0) {
     for (let i = indexOfActive - 1; i >= 0; i--) {
       const release = orderedScheduledReleases![i];
-      const action = releaseVersionMap.get(entryId)?.get(release.sys.id);
+      const action = releaseVersionMap.get(entityId)?.get(release.sys.id);
```
chrishelgert, line 41: good that we remove this with the app sdk cleanup, otherwise we would need to introduce the type too 😅

packages/_shared/src/hooks/useActiveReleaseLocalesStatuses.ts
MayaGillilan: Good call, although it also raises the question for me of why we're even passing in the entityId explicitly when we also pass in `currentEntityDraft` (which is a confusing name IMO)

chrishelgert (approved):

PR #1929, opened 2025-09-09
fix(_shared): restore the icon for the entity status badge []

chrishelgert:
with the forma update the icon was replaced which is not the one indented by UX - reverted it back to the desired one.

### DIFF
--- a/packages/_shared/src/LocalePublishingEntityStatusBadge/Icon.tsx
+++ b/packages/_shared/src/LocalePublishingEntityStatusBadge/Icon.tsx
@@ -1,0 +1,15 @@
+import React from 'react';
+
+import { generateIconComponent } from '@contentful/f36-icon';
+
+// TODO: use from forma icons, once it's changed there (currently a custom one here: https://github.com/contentful/forma-36/blob/main/packages/components/navbar/src/icons/CaretIcon.tsx)
+export const ArrowDownIcon = generateIconComponent({
+  name: 'ArrowDownIcon',
+  viewBox: '0 0 12 20',
+  path: (
+    <path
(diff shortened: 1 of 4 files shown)


packages/_shared/src/LocalePublishingEntityStatusBadge/Icon.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import { generateIconComponent } from '@contentful/f36-icon';
+
+// TODO: use from forma icons, once it's changed there (currently a custom one here: https://github.com/contentful/forma-36/blob/main/packages/components/navbar/src/icons/CaretIcon.tsx)
+export const ArrowDownIcon = generateIconComponent({
```
missating: The main point of the Forma v5 migration is that we can all use the new icons provided by [Phosphor](https://phosphoricons.com/) and be able to control the use of customs icons by adding those ourselves. 

The icon codemod replaced this icon in the migration with the `ArrowDownIcon` but from the `todo` comment this should be replaced with the `CaretDownIcon` 

Please check here and see if that works for your use case, if not we need to add a custom icon directly to Forma: https://f36.contentful.com/components/icon#examples cc @fidanismiri

fidanismiri: hey Nkoyo, I see. Can we replace the original with the `CaretDownIcon` , size tiny?  @missating

missating: yes, `CaretDownIcon` is available

chrishelgert: Screenshot 2025-09-09 at 17 26 24

updated will look like then like this (LBP and release)

PR #1930, opened 2025-09-12
fix: [TOL-3361] Allows bulk editing only for non active releases

joseliacosta:
## Description
This PR includes a small logic to check if there is an `activeRelease`  while editing a reference field. If yes, bulk editing will be disabled and it shows slide as regular (and disabled ) entry instead.

**Why?**
```
// WIP
```

### DIFF
--- a/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
+++ b/packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
@@ -93,7 +93,7 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
 
   const onEdit = async () => {
     const slide = await openEntry(props.sdk, props.entryId, {
-      bulkEditing: props.parameters.instance.bulkEditing,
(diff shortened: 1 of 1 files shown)


packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
```diff
@@ -93,7 +93,7 @@ export function FetchingWrappedEntryCard(props: EntryCardReferenceEditorProps) {
 
   const onEdit = async () => {
     const slide = await openEntry(props.sdk, props.entryId, {
-      bulkEditing: props.parameters.instance.bulkEditing,
```
chrishelgert, line 96: We could set the bulkEditing flag to false in the web app directly: https://github.com/contentful/user_interface/blob/1098e71cae01e9c4887ae2a9f20b5c1f40a015d9/src/javascripts/app/entity_editor/EntityField/useWidgetApi.ts#L254, through that customer wouldnt need to have the latest field editors :)

joseliacosta: Thanks for the suggestion, I'll close this PR then!

PR #1934, opened 2025-09-18
feat: fix wrong menu item component from forma

missating:
- Fixed issues with menu not closing after trigger
- Fixed failing test cause by v4 components in apps folder found in lock file

### DIFF
--- a/packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
+++ b/packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { Menu, MenuItem, Text } from '@contentful/f36-components';
+import { MenuItem, Text, MenuSectionTitle } from '@contentful/f36-components';
 import tokens from '@contentful/f36-tokens';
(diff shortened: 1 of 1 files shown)


packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
```diff
@@ -58,7 +58,7 @@ export function renderAssetInfo(props: { entityFile: File }) {
 
   return [
     <Menu.SectionTitle key="file-section">File info</Menu.SectionTitle>,
```
chrishelgert: is this also a standalone export? looks wrong when mixing Menu. and standalone menu imports 🙈

chrishelgert (approved):

packages/reference/src/assets/WrappedAssetCard/AssetCardActions.tsx
missating: You know, I was thinking this too, Let me check

missating: updated

PR #1935, opened 2025-09-19
feat: support release aware references [TOL-3360]

chrishelgert:
- Use sdk.release instead of temporary instanceParameters
- Update badges
- support for release in the entitystore


cypress/fixtures/FakeSdk.ts
```diff
@@ -70,8 +69,8 @@ export function createReferenceEditorTestSdk(props?: ReferenceEditorFakeSdkProps
   const localSdk = {
     field,
     locales,
-    cmaAdapter: createFakeCMAAdapter({
```
YvesRijckaert, line 73: just wondering, why did we remove `createFakeCMAAdapter`?

packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusPopover.tsx
```diff
@@ -170,25 +169,13 @@ export function ReleaseEntityStatusPopover({
     }, 300);
   }, []);
 
-  const status = determineBadgeStatus(releaseLocalesStatusMap, activeLocales);
+  const status = determineBadgeStatus(releaseStatusMap, activeLocales);
   const ariaLabel = status.secondary ? 'Multiple statuses' : status.primary;
   const wrapperClass = generateDynamicStyles(status);
 
-  if (isLoading) {
```
YvesRijckaert, line 177: No more loading state?

cypress/fixtures/FakeSdk.ts
chrishelgert: because we use the cma client now directly instead of creating a new one - through that we get some 🪄 from the web app (through that the adapter is no longer needed)

packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusPopover.tsx
chrishelgert: no as we no longer load anything at this point :)

packages/_shared/src/hooks/useReleaseStatus.ts
```diff
@@ -0,0 +1,178 @@
...
+  ReleaseV2Props,
+} from '../types';
+import { getEntityStatus } from '../utils/entityHelpers';
+import { sanitizeLocales } from '../utils/sanitizeLocales';
+
+function createReleaseLocaleStatus(
+  locale: Pick<LocaleProps, 'code' | 'default' | 'name'>,
+  status: ReleaseEntityStatus,
+): ReleaseLocalesStatus {
+  switch (status) {
```
YvesRijckaert, line 26: `willUnpublish` is not part of this switch case, is that expected?

chrishelgert: that is `becomesDraft` 🤔

packages/rich-text/stories/RichTextEditor.stories.tsx
```diff
@@ -301,9 +300,9 @@ const DemoRichTextEditor = () => {
 
 export const Default: Story = {
   parameters: {
-    controls: { hideNoControlsWarning: true }
+    controls: { hideNoControlsWarning: true },
   },
   render: () => {
     return <DemoRichTextEditor />;
-  }
+  },
```
YvesRijckaert, line 307: unrelated but we should have some auto formatting rules for this repo, a lot of the code changes are comma's everywhere 🙈

YvesRijckaert (approved):
LGTM !!! thank you 🚀

packages/_shared/src/hooks/useReleaseStatus.ts
YvesRijckaert: On line 167 `willUnpublish` is mentioned, that's why I was confused

chrishelgert: fixed that 😓

PR #1942, opened 2025-09-26
chore: add i18n to ReleaseEntityStatusBadge and ReleaseEntityStatusPopover [TOL-3432]

joseliacosta:
Make the following components internationalized:
- [x] ReleaseEntityStatusBadge
- [x] ReleaseEntityStatusPopover
- [x] ReleaseEntityStatusLocale
- [x] ReleaseEntityStatusLocalesList

Jira ticket: https://contentful.atlassian.net/browse/TOL-3432

✌🏼 Quick win: merge `RELEASE_BADGES` properties into `getReleaseStatusBadgeConfig` to avoid duplication

### DIFF
--- a/packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusBadge.tsx
+++ b/packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusBadge.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 
 import { Badge } from '@contentful/f36-components';
-import { getReleaseStatusBadgeConfig } from 'utils/getReleaseStatusBadgeConfig';
 
 import type { ReleaseEntityStatus } from '../types';
+import { getReleaseStatusBadgeConfig } from '../utils/getReleaseStatusBadgeConfig';
 
 type ReleaseEntityActionBadgeProps = {
   status: ReleaseEntityStatus;
--- a/packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusBadge.tsx
+++ b/packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusBadge.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { Badge, BadgeVariant } from '@contentful/f36-components';
+import { t } from '@lingui/core/macro';
 
 import type { ReleaseEntityStatus } from '../types';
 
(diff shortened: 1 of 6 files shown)


packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusBadge.tsx
```diff
@@ -34,14 +35,19 @@ const config: Record<ReleaseEntityStatus, { label: string; variant: BadgeVariant
 
 export function ReleaseEntityStatusBadge({ className, status }: ReleaseEntityActionBadgeProps) {
   const badgeConfig = config[status];
+  const statusLabel = badgeConfig.label;
```
chrishelgert: you have to change it in the config:
the config would need to be a function to generate it
inside you can then use `t` to build the messages with static messages
in the componetn here yo uwould say something like this:
const badeConfig = getConfig(status) // getConfig()[status]

joseliacosta: Thanks for the suggestion, @chrishelgert!
Like this 4943ce4493b565e09a78aabb969b9b78d8d51b7b?

chrishelgert: yes exactly

packages/_shared/src/utils/getReleaseStatusBadgeConfig.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function getReleaseStatusBadgeConfig(status: ReleaseEntityStatus): {
+  label: string;
+  variant: BadgeVariant;
+  default: string;
+  hover: string;
+  icon: string;
+} {
+  switch (status) {
+    case 'willPublish':
```
colomolo, line 14: Does it make sense to extract these statuses to some enum? Feels like we will need them in the future in other places as well.

joseliacosta: I don't know, @colomolo 🤔 
Given we use `ReleaseEntityStatus` from `types`, is still worth to create an enum?

colomolo: Ah, yeah, it's fine then 👍

colomolo (approved):

PR #1945, opened 2025-09-30
fix: improve rating field usability with trackpad [ZEND-6961]

MayaGillilan:
The onClick event for some reason was buggy with my trackpad and may have been causing users issues, so I changed it to a more reliable onMouseDown as well as fixed a very minor warning about props

### DIFF
--- a/packages/rating/src/RatingRibbon.tsx
+++ b/packages/rating/src/RatingRibbon.tsx
@@ -68,7 +68,7 @@ export class RatingRibbon extends React.Component<RatingRibbonProps, RatingRibbo
             testId="rating-editor-star"
             isDisabled={this.props.disabled}
             key={num}
-            onClick={() => {
+            onMouseDown={() => {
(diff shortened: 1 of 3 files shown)


packages/rating/src/RatingRibbon.tsx
```diff
@@ -68,7 +68,7 @@ export class RatingRibbon extends React.Component<RatingRibbonProps, RatingRibbo
             testId="rating-editor-star"
             isDisabled={this.props.disabled}
             key={num}
-            onClick={() => {
```
chrishelgert, line 71: onMouseDown will be triggered by right and middle clicks too, should we filter those out? https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value

MayaGillilan: True, will filter that and fix the tests quickly

PR #1947, opened 2025-10-02
fix: use entry and asset functions from cma instead of sdk.space [TOL-1684]

chrishelgert:
Migrate all sdk.space usage for getting an entry/asset to sdk.cma instead to benefit from batching, caching and release handling.
We then only have sdk.space left for content types which will be handled in another ticket


YvesRijckaert (approved):
amazing

packages/reference/src/common/EntityStore.tsx
```diff
@@ -603,12 +604,14 @@ const [InternalServiceProvider, useFetch, useEntityLoader, useCurrentIds] = cons
       }
 
       const subscribeQuery = ({ queryKey, queryHash }: Query) => {
-        const [entityType, entityId] = queryKey;
```
z0al, line 606: Every time we touch the entity store something breaks. Let's hope this won't cause any issues 😅

z0al (approved):
LGTM

packages/reference/src/common/EntityStore.tsx
chrishelgert: 🙈 we should get rid of it :D

PR #1948, opened 2025-10-02
fix: [TOL-3475] Handle no tertiary status

joseliacosta:
As TS already handle all possible ReleaseEntityStatus, this PR removes `switch/case` `default` statement. It was also throwing an exception unnecessarily when navigating from Create Assets to Releases

UPD by @colomolo: the problem was with setting css styles for tertiary badge even when there is no tertiary status, which resulted in getColor call with `{ secondary: undefined, tertiary: undefined }`. Now conditionally adding these styles only when tertiary status is needed.

### DIFF
--- a/packages/_shared/src/utils/getReleaseStatusBadgeConfig.ts
+++ b/packages/_shared/src/utils/getReleaseStatusBadgeConfig.ts
@@ -66,5 +66,7 @@ export function getReleaseStatusBadgeConfig(status: ReleaseEntityStatus): {
         hover: tokens.green400,
         icon: tokens.green400,
       };
+    default:
+      throw new Error(`Unknown release entity status: ${status}`);
   }
 }


packages/_shared/src/utils/getReleaseStatusBadgeConfig.ts
```diff
@@ -66,7 +66,5 @@ export function getReleaseStatusBadgeConfig(status: ReleaseEntityStatus): {
         hover: tokens.green400,
         icon: tokens.green400,
       };
-    default:
```
z0al, line 69: Why would it matter if the cases are all handled by the previous conditions and we exit ?

colomolo: I think throwing also a bit radical for a status badge to crash the whole page. Should we have 'Unknown' or 'Error' status or something like that for unforeseen errors?

chrishelgert: as it should never be called with the invalid status, it should not run to this point at all. Maybe we miss a check somewhere 🤔

joseliacosta: Folks, I think I found it! [10283d9](https://github.com/contentful/field-editors/pull/1948/commits/10283d9bfdd28fad12cddd2c2a14b646489f8588)
From y previous PR, to remove a small duplication when implementing `i18n` I've called `getReleaseStatusBadgeConfig` inside `getColor = ({ secondary, tertiary, isHover }: BadgeSVGType)`, but in this case, `status` was not `ReleaseEntityStatus` was a bit diff one:
```ts
type Status = {
  primary: ReleaseEntityStatus;
  secondary?: ReleaseEntityStatus;
  tertiary?: ReleaseEntityStatus;
};
```
To avoid the issue, even removing the default statement from `switch`, I put an early `return` to prevent it to call this config function unnecessarily.

If possible, I would like to understand why or which cases we actually need to get the colors from `status.secondary` or `status.tertiary` because in all my debugging cases I saw it returning as `undefined` but perhaps I did not test all possibilities. cc.: @colomolo `&&` @YvesRijckaert

colomolo: @joseliacosta I think the whole logic is essentially described here:
https://github.com/contentful/field-editors/blob/TOL-3475/packages/_shared/src/ReleaseEntityStatusBadge/ReleaseEntityStatusPopover.tsx#L132-L149

If there are three different statuses, we show triple stack with different colors. If there are only two varying statuses, we show double stack (no tertiary). But when it's the same status for all locales, we still show double stack to signal that publishing is per locale (LBP enabled). So you always have at least double-stacked badge with LBP.

PR #1949, opened 2025-10-02
feat: migrate remaining space.getEntries

chrishelgert:
- refactor(slug): use cma client instead of space for resolving entries
- refactor(validation-errors): use cma client instead of space for resolving
BREAKING CHANGE: Needs cma as property now for validation-errors


MayaGillilan:
If there's a breaking change should we add it to the commit message?

PR #1952, opened 2025-10-07
fix: useReleaseStatus respects Published status

MayaGillilan:
The refactor of useReleaseStatus was missing a case for published status in the returned `releaseEntityStatus`, this restores it so we get the correct status instead of the default "not in release'

### DIFF
--- a/packages/_shared/src/hooks/useReleaseStatus.ts
+++ b/packages/_shared/src/hooks/useReleaseStatus.ts
@@ -130,6 +130,10 @@ export function useReleaseStatus({
 
   const releaseEntityStatus: ReleaseEntityStatus = useMemo(() => {
     const releaseArray = Array.from(releaseStatusMap.values());
+    if (releaseArray.find(({ status }) => status === 'published')) {
+      return 'published';
+    }
+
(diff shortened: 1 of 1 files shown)


packages/_shared/src/hooks/useReleaseStatus.ts
```diff
@@ -130,6 +130,10 @@ export function useReleaseStatus({
 
   const releaseEntityStatus: ReleaseEntityStatus = useMemo(() => {
     const releaseArray = Array.from(releaseStatusMap.values());
+    if (releaseArray.find(({ status }) => status === 'published')) {
+      return 'published';
```
colomolo, line 134: In which case do we have published status for release entities?

colomolo: This ticket?
https://contentful.atlassian.net/browse/TOL-3499

MayaGillilan: We return this also if the entity is not in a release, so if the release is published but not in the release we should still return published here

PR #1955, opened 2025-10-17
feat(rich-text): add character validation

z0al:
Screenshot 2025-10-17 at 12 34 32

### Todos

- [ ] unit test against rich-text-plain-text-renderer

### DIFF
--- a/packages/rich-text/src/CharConstraints.tsx
+++ b/packages/rich-text/src/CharConstraints.tsx
@@ -32,10 +32,10 @@ function CharCounter({ checkConstraints }: CharCounterProps) {
 
   return (
     <span
-      data-status-code={valid ? null : 'invalid-size'}
       className={cx({
         [styles.counterInvalid]: !valid,
-      })}>
+      })}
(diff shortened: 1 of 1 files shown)


YvesRijckaert (approved):

packages/rich-text/src/CharConstraints.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+function CharCounter({ checkConstraints }: CharCounterProps) {
+  const editor = usePlateEditorState();
+
+  const count = editor.getCharacterCount();
+  const valid = checkConstraints(count);
+
+  return (
+    <span
+      data-status-code={valid ? null : 'invalid-size'}
```
YvesRijckaert: where do we use this data attribute?

z0al: Nothing. Taken from the original CharValidation component in field-editors-shared but looking it up, it's not used anywhere. Removed it

PR #1956, opened 2025-10-20
fix: deploy playground to vercel [TOL-3428]

2wce:
## Changes

- The build output for _shared/src/index.tsx had the exports before the imports which broke the build for storybook. Since we don't need to import only to re-export I've combined the 2 into one step
- upgraded test util version since some apps had an older version which led to symlink issues locally and on CI eg https://vercel.com/contentful-apps/field-editors/FDHStNRwBad3oeXfAxTuppqLBvuR#L91

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,15 +4,19 @@ orbs:
   vault: contentful/vault@1
 
 executors:
-  linux-cypress:
-    docker:
-      - image: cypress/browsers:node-22.20.0-chrome-141.0.7390.107-1-ff-144.0-edge-141.0.3537.85-1
-    environment:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -7,13 +7,13 @@ executors:
   linux-node:
     resource_class: xlarge
     docker:
-      - image: cimg/node:20.14.0
+      - image: cimg/node:22.14.0
```
BobHemphill76: is this targeting the correct latest 22.x node img?

2wce: nice catch, i had replaced the version incorrectly 👌🏽

PR #1959, opened 2025-10-22
fix(rich-text): compute count for initial value

z0al:
Plate bypasses `editor.apply` when setting the initial value of the editor (happens after plugins are initialised), causing the initial editor count always to be `0` until the editor is focused.

Since the `usePlateEditorState` re-renders after the initial value is set, the `editor.getCharacterCount` was being called but giving a stale value. The fix is to delay computing the initial count until the first call to `getCharacterCount` to make sure we are working with the initial editor value.

### DIFF
--- a/packages/rich-text/src/plugins/CharCounter/withCharCounter.ts
+++ b/packages/rich-text/src/plugins/CharCounter/withCharCounter.ts
@@ -3,6 +3,8 @@ import debounce from 'p-debounce';
 import { PlateEditor } from '../../internal';
 import { getTextContent } from './utils';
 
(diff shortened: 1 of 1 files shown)


packages/rich-text/src/CharConstraints.tsx
```diff
@@ -24,12 +28,21 @@ interface CharCounterProps {
   checkConstraints: (value: number) => boolean;
 }
 
+const getCharacterCount = debounce((editor: PlateEditor) => {
+  return getTextContent(editor).length;
+}, 300);
```
joseliacosta: nit: is worth to have this debounce time as a constant somewhere?

PR #1961, opened 2025-10-27
fix(reference): share scheduled actions key [TOL-3575]

YvesRijckaert:


packages/reference/src/common/EntityStore.tsx
```diff
@@ -492,38 +496,31 @@ const [InternalServiceProvider, useFetch, useEntityLoader, useCurrentIds] = cons
           return new Promise<ScheduledAction[]>((resolve) => resolve([]));
         }
 
-        // This is fixed to force the cache to reuse previous results
-        const fixedEntityCacheId = 'scheduledActionEntityId';
-
-        // A space+environment combo can only have up to 500 scheduled actions
+        // A space+environment combo can only have up to 500 scheduled actions (1000 soon, see: https://github.com/contentful/user_interface/pull/28351)
```
chrishelgert: we may not want to reference it back to user_interface as it's internal repo, maybe link it the other way around

packages/reference/src/common/queryClient.tsx
```diff
@@ -10,12 +14,15 @@ const clientContext = React.createContext<QueryClient | undefined>(undefined);
 
 export function useQueryClient(): QueryClient {
   const client = React.useContext(clientContext);
+  const hostClient = useHostQueryClient();
```
chrishelgert: Isn't it the same as the line before? https://github.com/TanStack/query/blob/38b40081cedd207285ac0b4bb6818a2e84704b95/packages/react-query/src/QueryClientProvider.tsx#L10

YvesRijckaert: The way I understand it:
`const client = React.useContext(clientContext)` reads the field editors context, defined here: https://github.com/contentful/field-editors/blob/dc6497e4e2370ec54c13f904b7d45356e6a7356b/packages/reference/src/common/queryClient.tsx#L9
then this new one looks at React Query’s global context

is that right? 🤔

chrishelgert: mhh right, but if we would use `QueryClientContext` from React Query instead?
the `useQueryClient` would throw an error if it doesnt find an context

YvesRijckaert: '"@tanstack/react-query"' has no exported member named 'QueryClientContext' 😭

But I wrapped it with try catch?

chrishelgert: ah I didnt saw the try catch when I had a look before 🙈

PR #1962, opened 2025-10-27
feat: [TOL-3522] Add flag for handling status for references

colomolo:
Jira Ticket: [TOL-3522](https://contentful.atlassian.net/browse/TOL-3522) | Pairing w/ @joseliacosta 
# Description
Previously we fixed the Status badge mostly for reference entries/assets. 
When an entity is a reference and is **published** on **current** or somewhere else, it shows the badge `published` even if is not part of the release.
If an entity is not a reference, because we reuse `useReleaseStatus` hook, we display a wrong status badge and wrong footer color, specially for those that are published somewhere (e.g.; Current) but they are not part of the release. 

It's a problem because even before you add to release or not (from Campaign Modal) we already render the wrong footer _(see image)_
image-20251014-084004

For this reason, to determine the correct status to be shown we need to distinct if status is being used in Entity reference or just regular one. 

### Quick-win 🚀 
I've suggested a refactor on test file to include all the status and reuse some repeated code snippets. 

[TOL-3522]: https://contentful.atlassian.net/browse/TOL-3522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/_shared/src/hooks/useReleaseStatus.spec.ts
+++ b/packages/_shared/src/hooks/useReleaseStatus.spec.ts
@@ -135,462 +135,527 @@ const expectEntityStatus = (
 const ENTITY_TYPES: EntityType[] = ['Entry', 'Asset'];
 
 describe('useReleaseStatus', () => {
-  ENTITY_TYPES.forEach((entityType) => {
-    const defaultEntityId = entityType === 'Entry' ? 'entry-1' : 'asset-1';
-    const differentEntityId = entityType === 'Entry' ? 'entry-2' : 'asset-2';
+  const locales = createDefaultLocales();
+
+  describe('Guard clauses and invalid inputs', () => {
+    it('returns empty map when entity is undefined', () => {
+      const { result } = renderHook(() =>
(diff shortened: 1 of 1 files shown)


packages/reference/src/entries/WrappedEntryCard/FetchingWrappedEntryCard.tsx
chrishelgert, line 1: and AssetCard

joseliacosta: @chrishelgert, `FetchingWrappedEntryCard`| `...AssetCard` inside rich-text are considered references too?

chrishelgert: yes

YvesRijckaert (approved):

PR #1965, opened 2025-10-29
docs(richText): document overrides issue []

YvesRijckaert:
Add documentation note that users should use the overrides / resolutions to ensure compatible versions are installed

This is temporary until we have capacity to upgrade plate to latests version

### DIFF
--- a/packages/rich-text/README.md
+++ b/packages/rich-text/README.md
@@ -14,6 +14,7 @@ import { RichTextEditor } from '@contentful/field-editor-rich-text';
 ## ⚠️ Important: Package Configuration
 
 Due to peer dependency resolution in npm, you must add the following to your `package.json` to ensure compatible versions are installed:
+This is a temporary workaround until we have upgraded the underlying rich text packages.
 
 **For npm:**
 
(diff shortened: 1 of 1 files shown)


packages/rich-text/README.md
```diff
@@ -11,6 +11,39 @@ import 'codemirror/lib/codemirror.css';
...
+    "slate": "0.94.1",
+    "slate-react": "0.102.0"
+  }
+}
+```
+
+After adding the overrides/resolutions, reinstall dependencies:
+
+```bash
+rm -rf node_modules package-lock.json  # or yarn.lock for Yarn
```
chrishelgert: we should not recommend this, it might break even more things

packages/rich-text/README.md
```diff
@@ -11,6 +11,39 @@ import 'codemirror/lib/codemirror.css';
 import { RichTextEditor } from '@contentful/field-editor-rich-text';
 ```
 
+## ⚠️ Important: Package Configuration
```
chrishelgert, line 14: maybe call out that this is only temporarly until we solved the conflicts

PR #1967, opened 2025-11-03
chore: upgrade slate, slate-react and add slate-dom

joseliacosta:
Upgrade folowing versions:
Remove overrides that are no longer needed due the usage of latest versions of slate-*

### DIFF
--- a/cypress/component/rich-text/RichTextEditor.spec.ts
+++ b/cypress/component/rich-text/RichTextEditor.spec.ts
@@ -51,16 +51,16 @@ describe('Rich Text Editor', { viewportHeight: 2000, viewportWidth: 1000 }, () =
             BLOCKS.TABLE_ROW,
             {},
             block(BLOCKS.TABLE_HEADER_CELL, {}, paragraphWithText('heading 1')),
-            block(BLOCKS.TABLE_HEADER_CELL, {}, paragraphWithText('heading 2'))
+            block(BLOCKS.TABLE_HEADER_CELL, {}, paragraphWithText('heading 2')),
           ),
           block(
             BLOCKS.TABLE_ROW,
             {},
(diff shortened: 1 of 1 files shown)


cypress/component/rich-text/RichTextEditor.spec.ts
```diff
@@ -147,18 +147,8 @@ describe('Rich Text Editor', { viewportHeight: 2000, viewportWidth: 1000 }, () =
...
-      } else {
-        richText.expectValue(
-          doc(
-            block(BLOCKS.PARAGRAPH, {}, text('some')),
-            entryBlock(),
-            block(BLOCKS.PARAGRAPH, {}, text(' text.')),
-            emptyParagraph()
-          )
-        );
-      }
```
joseliacosta, line 161: Previously, when using chrome, why was it expected we drag an embed entry into the middle of a string? cc.: @z0al

joseliacosta: e.g.:

Image

z0al: That's how dragging blocks into text/inline elements should behave generally. It's supposed to split the inline node and tries to insert the block in between. What is the reason for the change in behavior?

z0al: That said, if that's the only breaking change it's probably Ok.

joseliacosta: > if that's the only breaking change it's probably Ok.

Thanks for checking 🙏🏼 

I've debugged a test that started failing after `slate` and `slate-react` upgrade and then realized this split was no longer happening. To be safe I checked if dragging and dropping inline embed entries was working as before and good news, the behavior is the same 🎉

PR #1975, opened 2025-11-11
fix: add useFocused hook from slate-react update to show TableActions

joseliacosta:
After bumping `slate-react`, is needed to check both `useSelect` AND `useFocused` to determine if an element should show interactive UI elements. e.g.: `<TableActions/>`


2wce (approved):
🤞🏽

PR #1977, opened 2025-11-11
feat: resolve transformed image urls [TOL-3598]

2wce:
we want to test and see if this change will help consistently transform thumbnails for large assets


packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx
```diff
@@ -114,6 +114,20 @@ export const WrappedAssetCard = ({
...
+    if (size === 'small') {
+      return entityHelpers.getResolvedImageUrl(entityFile.url, {
+        width: 150,
+        height: 150,
+        fit: 'thumb',
+      });
+    }
+
+    return entityHelpers.getResolvedImageUrl(entityFile.url, { height: 300 });
+  };
```
chrishelgert: why not const imageUrl = useMemo(...

PR #1983, opened 2025-11-13
feat: resolve transformed image urls [TOL-3598]

2wce:
this will help us consistently transform image urls for thumbnails. tested as part of canary release [in this PR](https://github.com/contentful/field-editors/pull/1977) which pinpointed the bug fix that has also been added to this PR

### DIFF
--- a/packages/_shared/src/utils/entityHelpers.ts
+++ b/packages/_shared/src/utils/entityHelpers.ts
@@ -306,15 +306,6 @@ export const getEntryImage = async (
   }
 };
 
-const TRANSFORMATIONS_ENDPOINTS = {
-  'downloads.ctfassets.net': 'images.ctfassets.net',
-  'downloads.flinkly.com': 'images.flinkly.com',
-};
-
-const DOWNLOADS_ENDPOINTS = Object.keys(TRANSFORMATIONS_ENDPOINTS);
-
-type TransformationEndpointKey = keyof typeof TRANSFORMATIONS_ENDPOINTS;
-
 export const getResolvedImageUrl = (
   url: string,
   params?: { w?: number; h?: number; fit?: string },
@@ -324,10 +315,9 @@ export const getResolvedImageUrl = (
     const urlToParse = url.startsWith('//') ? `https:${url}` : url;
     const parsedUrl = new URL(urlToParse);
 
-    // Replace downloads domain with images domain if it matches
-    if (DOWNLOADS_ENDPOINTS.includes(parsedUrl.hostname)) {
-      parsedUrl.hostname =
-        TRANSFORMATIONS_ENDPOINTS[parsedUrl.hostname as TransformationEndpointKey];
+    // Replace downloads.* domain with images.* domain
+    if (parsedUrl.hostname.startsWith('downloads.')) {
+      parsedUrl.hostname = parsedUrl.hostname.replace(/^downloads\./, 'images.');
     }
 
     if (params) {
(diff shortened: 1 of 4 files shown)


packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -305,3 +305,54 @@ export const getEntryImage = async (
     return null;
   }
 };
+
+const TRANSFORMATIONS_ENDPOINTS = {
+  'downloads.ctfassets.net': 'images.ctfassets.net',
+  'downloads.flinkly.com': 'images.flinkly.com',
```
colomolo: Not sure that we should expose domains here like this. Can we get away with `'downloads.': 'images.'` maybe?

2wce: the fear was we might end up changing a domain name that we shouldn't change. not sure if we can invert control & have the consumer provide the endpoints? it might mean some prop drilling but i can take a look

2wce: unless we are guaranteed to have full control over this domain then that would be trivial to do you suggestion

packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -305,3 +305,54 @@ export const getEntryImage = async (
...
+    // Replace downloads domain with images domain if it matches
+    if (DOWNLOADS_ENDPOINTS.includes(parsedUrl.hostname)) {
+      parsedUrl.hostname =
+        TRANSFORMATIONS_ENDPOINTS[parsedUrl.hostname as TransformationEndpointKey];
+    }
+
+    if (params) {
+      Object.entries(params).forEach(([key, value]) => {
+        if (value !== undefined) {
+          parsedUrl.searchParams.set(key, value.toString());
```
colomolo, line 326: I think `toString()` here is not needed as it's handled later already.
```suggestion
          parsedUrl.searchParams.set(key, value);
```

packages/_shared/src/utils/entityHelpers.ts
```diff
@@ -305,3 +305,54 @@ export const getEntryImage = async (
...
+    // If original URL was protocol-relative, restore it
+    return url.startsWith('//') ? result.replace(/^https:/, '') : result;
+  } catch {
+    // fallback to previous behaviour for relative URLs
+    if (!params) return url;
+
+    const searchParams = new URLSearchParams();
+    Object.entries(params).forEach(([key, value]) => {
+      if (value !== undefined) {
+        searchParams.set(key, value.toString());
```
colomolo, line 341: same here

packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx
```diff
@@ -97,6 +97,24 @@ export const WrappedAssetCard = ({
...
+  const entityFile = asset.fields.file
+    ? asset.fields.file[localeCode] || asset.fields.file[defaultLocaleCode]
+    : undefined;
+
+  const imageUrl = React.useMemo(() => {
+    if (!entityFile?.url) return '';
+
+    if (size === 'small') {
+      return entityHelpers.getResolvedImageUrl(entityFile.url, {
+        w: 150,
```
colomolo: nit: would be great to have these sizes as consts.

packages/_shared/src/utils/entityHelpers.ts
2wce: value will be either number or string at this point since we have mixed types so TS will complain if it's not a string here

PR #2007, opened 2025-12-02
chore: security cleanup []

2wce:
we don't need the workflow with the third party approval step


joseliacosta (approved):
thanks for taking care of it 🙏🏼

PR #2013, opened 2025-12-16
feat(rich-text): speed up initial normalization and reduce re-renders

z0al:
## Background

An initial normalization step is needed to handle a few edge cases where a document may be valid schema-wise but either doesn’t make sense to render in the editor or is incompatible with Slate.

The challenge is that we don't want this initial normalization step to trigger an onChange event hence it's done before initializing the editor.

## Current logic

We currently initialize a temporary Plate editor with all plugins, run normalization, and then discard it. This allows us to reuse all existing plugin-defined normalization rules, but it is expensive—especially when the editor re-renders.

## Proposal

I reimplemented a dedicated adapter that converts the API document to Slate and performs normalization in the same step. By design, **_this is not a 1:1 mapping of the previous logic_**, but it should be sufficient for our needs since many of the existing normalization rules handle cases that aren’t possible given the API schema. Additionally, the previous logic had bugs.

The perf gain is actually more than I expected:

```bash
# benchmarked using npm.im/benchmark
current implementation x 6.89 ops/sec ±26.27% (15 runs sampled)
new implementation x 5,812 ops/sec ±2.03% (97 runs sampled)
Fastest is: new implementation (843.74x faster)
```

## Risks
I may have missed a case or two that requires handling; the implementation is largely based on existing integration tests. If we encounter a missing case the editor may crash or result in an intended auto-save event upon visiting an entry

### DIFF
--- a/packages/rich-text/src/helpers/toSlateDoc.ts
+++ b/packages/rich-text/src/helpers/toSlateDoc.ts
@@ -7,7 +7,7 @@ import {
   type Hyperlink,
   BLOCKS,
 } from '@contentful/rich-text-types';
-import { Text as TextInterface } from 'slate';
+import { Text as TextInterface, Element as ElementInterface } from 'slate';
 
 import type { Text, Element } from '../internal';
 import { isText } from '../internal';
@@ -118,6 +118,12 @@ function transformNode(node: CfBlock | CfInline): Element {
     // can it be next to another inline node in the children array. If this is
     // the case, an empty text node will be added to correct this to be in
     // compliance with the constraint.
+    //
+    // Note: this only handles the case where an inline node is the first or
+    // after another inline node. Handling the case where an inline node is
+    // the last child of a parent block is implemented further down below
+    // outside the loop.
+    //
     // Ref: https://docs.slatejs.org/concepts/11-normalizing
     if (inlineTypes.has(child.nodeType) && !isText(lastChild)) {
       el.children.push({ text: '' });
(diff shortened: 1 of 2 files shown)


packages/rich-text/src/helpers/toSlateDoc.ts
```diff
@@ -0,0 +1,201 @@
...
+    // Empty hyperlinks should be removed
+    if (isEmptyHyperlink(child)) {
+      continue;
+    }
+
+    // Inline nodes cannot be the first or last child of a parent block, nor
+    // can it be next to another inline node in the children array. If this is
+    // the case, an empty text node will be added to correct this to be in
+    // compliance with the constraint.
+    // Ref: https://docs.slatejs.org/concepts/11-normalizing
```
z0al, line 127: I just realised I'm missing tests for this + handling for inlines appearing as last element of a block. I will fix that shortly

cypress/component/rich-text/document-mocks/normalizationWithoutValueChange.js
```diff
@@ -102,6 +102,12 @@ export default {
             },
           ],
         },
+      ],
+    },
+    {
+      nodeType: 'unordered-list',
+      data: {},
+      content: [
```
z0al, line 110: Previously an empty list was being dropped instead of being fixed (by adding an empty list item) like the other cases in the same test 🤷🏻

packages/rich-text/src/helpers/toSlateDoc.ts
z0al: Fixed

chrishelgert (approved):
🥇

joseliacosta (approved):
Awesome! 💯

PR #2038, opened 2026-01-08
fix: show updated entry reference title when in a specific release [TOL-3711]

joseliacosta:
## Description

When patching a reference entry inside a release it must also invalidate cache and get a fresh entities list from `GET /entries?release=`
It was working only for current release because:
- store was not up-to-date when fetching new data with release info;
- queryKey was explicitly checking the qtd of params (and potentially ignoring when `releaseId` is present)

## Before
image (3) (1)

## After
new-reference-entry

Jira: [TOL-3711](https://contentful.atlassian.net/browse/TOL-3711)

[TOL-3711]: https://contentful.atlassian.net/browse/TOL-3711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/reference/src/common/EntityStore.tsx
+++ b/packages/reference/src/common/EntityStore.tsx
@@ -201,8 +201,8 @@ const isEntityQueryKey = (queryKey: QueryKey): queryKey is EntityQueryKey => {
   return (
     Array.isArray(queryKey) &&
     (queryKey[0] === 'Entry' || queryKey[0] === 'Asset') &&
-    // length === 5 because releaseId is optional parameter
-    queryKey.length >= 4
+    // length === 4 when releaseId is not present, length === 5 when releaseId is present
+    (queryKey.length === 4 || queryKey.length === 5)
   );
 };
 


packages/reference/src/common/EntityStore.tsx
```diff
@@ -595,13 +596,17 @@ const [InternalServiceProvider, useFetch, useEntityLoader, useCurrentIds] = cons
     // @ts-expect-error ...
     const onEntityChanged = props.sdk.space.onEntityChanged;
     const onSlideInNavigation = props.sdk.navigator.onSlideInNavigation;
-
+    console.log('>>> FieldEditors onEntityChanged', onEntityChanged);
```
MayaGillilan: ;)

packages/reference/src/common/EntityStore.tsx
```diff
@@ -201,7 +201,8 @@ const isEntityQueryKey = (queryKey: QueryKey): queryKey is EntityQueryKey => {
   return (
     Array.isArray(queryKey) &&
     (queryKey[0] === 'Entry' || queryKey[0] === 'Asset') &&
-    queryKey.length === 4
+    // length === 5 because releaseId is optional parameter
+    queryKey.length === 5
```
chrishelgert: 4 or 5 or?
without the release information it hsould be still 4 🤔

packages/reference/src/common/EntityStore.tsx
chrishelgert: dont forget to remove your console.logs ;)

joseliacosta: Oops 🙈 
Thanks for catching this, just got to excited by the progress hahaha

packages/reference/src/common/EntityStore.tsx
joseliacosta: 🤦🏼‍♀️ Oh, yes, you are right. So I've adapted here to got both cases.

packages/reference/src/common/EntityStore.tsx
```diff
@@ -201,7 +201,8 @@ const isEntityQueryKey = (queryKey: QueryKey): queryKey is EntityQueryKey => {
   return (
     Array.isArray(queryKey) &&
     (queryKey[0] === 'Entry' || queryKey[0] === 'Asset') &&
-    queryKey.length === 4
+    // length === 5 because releaseId is optional parameter
```
chrishelgert: Can we make it more explicit?
If a releaseId is set, it should be 5; otherwise, it should be 4?

PR #2043, opened 2026-01-14
feat: Migrate field editors .space.getCachedContentTypes to Plain Client APIs [TOL-3472]

joseliacosta:
## Description

Migrates content type fetching from deprecated `space.getCachedContentTypes()` to `cma.contentType.getMany()` and consolidates duplicate logic into a shared `useContentTypes` hook.

## Problem

1. **Deprecated API**: `space.getCachedContentTypes()` was deprecated in SDK v4.0.0
2. **Code duplication**: 8+ files manually calling `cma.contentType.getMany()` with duplicate pagination logic
3. **Pagination bug**: Original implementation limited to 1000 content types - spaces with more would silently lose data

## 🎁 Bonus

- This migration actually **fixes a silent bug** 🎉  the original `getCachedContentTypes()` was limited to 1000 content types (verified in webapp source). This PR aims to properly paginates through all content types.
- sharing queryKeys between user interface and field-editors can lead to mismatches. From now it introduces a `queryKey` mapping to centralize the react query keys we reuse from user-interface;
- Before https://github.com/contentful/field-editors/pull/2045, only reference package implemented react-query. Now it's available on `_shared` level but the packages that uses `useContentType` must implement `SharedQueryClientProvider` as well.

[Jira Ticket](https://contentful.atlassian.net/browse/TOL-3472)

### DIFF
--- a/packages/_test/src/createFakeSpaceAPI.ts
+++ b/packages/_test/src/createFakeSpaceAPI.ts
@@ -64,9 +64,6 @@ const testContentTypes: ContentType[] = [
 
 export function createFakeSpaceAPI(customizeMock: CustomizeMockFn = identity): SpaceAPI {
   return customizeMock({
-    getCachedContentTypes() {
-      return testContentTypes;
-    },
     getContentTypes() {
       return Promise.resolve({
         items: testContentTypes,
--- a/packages/markdown/src/__fixtures__/FakeSdk.ts
+++ b/packages/markdown/src/__fixtures__/FakeSdk.ts
@@ -121,9 +121,6 @@ export function newReferenceEditorFakeSdk(props?: ReferenceEditorSdkProps): [Fie
     },
     space: {
       ...space,
-      getCachedContentTypes() {
-        return localizeContentTypes(space.getCachedContentTypes());
-      },
       getContentTypes() {
         return Promise.resolve(
           space.getContentTypes().then((response) => {
--- a/packages/rich-text-alpha/src/components/FetchingWrappedEntryCard.tsx
+++ b/packages/rich-text-alpha/src/components/FetchingWrappedEntryCard.tsx
@@ -14,10 +14,11 @@ import {
   useLocalePublishStatus,
   useActiveLocales,
   useReleaseStatus,
-  useContentTypes,
+  useContentType,
   type ReleaseStatusMap,
   type ReleaseV2Props,
   type ReleaseEntityStatus,
+  SharedQueryClientProvider,
 } from '@contentful/field-editor-shared';
 import areEqual from 'fast-deep-equal';
 
@@ -99,7 +100,7 @@ interface FetchingWrappedEntryCardProps {
   onRemove?: VoidFunction;
 }
 
-export const FetchingWrappedEntryCard = (props: FetchingWrappedEntryCardProps) => {
+const InternalFetchingWrappedEntryCard = (props: FetchingWrappedEntryCardProps) => {
   const { entryId, onEntityFetchComplete } = props;
   const { data: entry, status, currentEntity } = useEntity<Entry>('Entry', entryId);
   const { getEntityScheduledActions } = useEntityLoader();
@@ -116,11 +117,10 @@ export const FetchingWrappedEntryCard = (props: FetchingWrappedEntryCardProps) =
     isReference: true,
   });
 
-  const { contentTypes: allContentTypes } = useContentTypes(props.sdk);
-  const contentType = React.useMemo(
-    () => allContentTypes.find((ct) => entry && ct.sys.id === entry.sys.contentType.sys.id),
-    [allContentTypes, entry],
-  );
+  const contentTypeId = entry?.sys.contentType.sys.id;
+  const { data: contentType } = useContentType(props.sdk, contentTypeId || '', {
+    enabled: !!contentTypeId,
+  });
 
   React.useEffect(() => {
     if (status === 'success') {
@@ -158,5 +158,13 @@ export const FetchingWrappedEntryCard = (props: FetchingWrappedEntryCardProps) =
       release={props.sdk.release as ReleaseV2Props | undefined}
       releaseEntityStatus={releaseEntityStatus}
     />
+  );
+};
+
+export const FetchingWrappedEntryCard = (props: FetchingWrappedEntryCardProps) => {
+  return (
+    <SharedQueryClientProvider>
+      <InternalFetchingWrappedEntryCard {...props} />
+    </SharedQueryClientProvider>
   );
 };
--- a/packages/rich-text-alpha/src/plugins/embeds/components/FetchingWrappedInlineEntryCard.tsx
+++ b/packages/rich-text-alpha/src/plugins/embeds/components/FetchingWrappedInlineEntryCard.tsx
@@ -13,7 +13,8 @@ import {
   Entry,
   FieldAppSDK,
   entityHelpers,
-  useContentTypes,
+  useContentType,
+  SharedQueryClientProvider,
 } from '@contentful/field-editor-shared';
 import { INLINES } from '@contentful/rich-text-types';
 import { css } from 'emotion';
@@ -34,7 +35,7 @@ type InternalFetchingWrappedInlineEntryCardProps = Pick<
   locale: string;
   defaultLocale: string;
   entry: Entry;
-  allContentTypes: ContentType[];
+  contentType?: ContentType;
   getEntityScheduledActions: React.ComponentProps<
     typeof ScheduledIconWithTooltip
   >['getEntityScheduledActions'];
@@ -43,7 +44,7 @@ type InternalFetchingWrappedInlineEntryCardProps = Pick<
 
 function InternalFetchingWrappedInlineEntryCard({
   entry,
-  allContentTypes,
+  contentType,
   locale,
   defaultLocale,
   isSelected,
@@ -53,16 +54,6 @@ function InternalFetchingWrappedInlineEntryCard({
   onRemove,
   isDisabled,
 }: InternalFetchingWrappedInlineEntryCardProps) {
-  const contentType = React.useMemo(() => {
-    if (!allContentTypes) {
-      return undefined;
-    }
-
-    return allContentTypes.find(
-      (contentType) => contentType.sys.id === entry.sys.contentType.sys.id,
-    );
-  }, [allContentTypes, entry]);
-
   const title = React.useMemo(
     () =>
       getEntryTitle({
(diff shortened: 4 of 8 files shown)


packages/_shared/src/hooks/useContentTypes.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useState } from 'react';
+
+import { ContentType, FieldAppSDK } from '@contentful/app-sdk';
+
+type SDKWithCMA = Pick<FieldAppSDK, 'cma'>;
+type CMAClient = SDKWithCMA['cma'];
+
+export function useContentTypes(source: SDKWithCMA | CMAClient): ContentType[] {
+  const [contentTypes, setContentTypes] = useState<ContentType[]>([]);
```
chrishelgert: WDYT about using react-query?
with the shared React query client as we do for references

joseliacosta: Now ` useContentTypes` hook fully relies on React Query. 
I've learned a lot during this refactor, thanks for the suggestion, @chrishelgert !

cypress/component/rich-text/RichTextEditor.Commands.spec.ts
```diff
@@ -14,7 +14,8 @@ describe('Rich Text Editor - Commands', { viewportHeight: 2000, viewportWidth: 1
   beforeEach(() => {
     cy.viewport(1000, 2000);
     richText = new RichTextPage();
-    mountRichTextEditor();
+    // Pre-fetch content types to avoid async loading issues in keyboard navigation tests
+    cy.wrap(null).then(() => mountRichTextEditor({ prefetchContentTypes: true }));
```
chrishelgert: why do we need the `cy.wrap(null)`?

cypress/component/rich-text/RichTextEditor.Commands.spec.ts
```diff
@@ -108,66 +109,72 @@ describe('Rich Text Editor - Commands', { viewportHeight: 2000, viewportWidth: 1
 
     it('should navigate then embed on pressing enter', () => {
       richText.editor.click().type('/');
-      getCommandList().findByText('Embed Example Content Type').should('exist');
-      richText.editor.type('{enter}');
-      getCommandList().findByText('Embed Example Content Type').should('not.exist');
-      richText.editor.type('{enter}');
+      getCommandList().findByText('Embed Example Content Type').click();
```
chrishelgert: this tests now a differnt behaviour doesnt it?

chrishelgert: we should keep the behaviour where we test the keyboard input

packages/_shared/src/hooks/useContentTypes.ts
```diff
@@ -0,0 +1,70 @@
...
+  const { data: contentTypes = [] } = useQuery(
+    CONTENT_TYPES_QUERY_KEY,
+    async () => {
+      const allContentTypes: ContentType[] = [];
+      const limit = 1000;
+      let skip = 0;
+      let total = 0;
+
+      do {
+        const response = await cma.contentType.getMany({ query: { limit, skip } });
```
chrishelgert, line 77: Please check that we do the same query and and use the same query key as here: user_interface: src/javascripts/core/react-query/cma/contentType/useGetManyContentTypes.ts

then the data should be shared between user interface and field editors :)

packages/rich-text/src/plugins/Hyperlink/useEntityInfo.ts
```diff
@@ -41,7 +41,9 @@ async function fetchAllData({
     : sdk.cma.asset.get({ assetId: entityId }));
   if (entity.sys.contentType) {
     const contentTypeId = entity.sys.contentType.sys.id;
-    contentType = sdk.space.getCachedContentTypes().find((ct) => ct.sys.id === contentTypeId);
+    contentType = (await sdk.cma.contentType.getMany({})).items.find(
```
chrishelgert: we could only search for the one needed here 🤔 

or load all with the hook and only search then in this place (to use the cached data)

packages/_shared/src/hooks/useContentTypes.ts
joseliacosta: Good catch! Thanks for the suggestion and I think as a quick-win make sense to isolate the keys in a single place where we could avoid some mismatching between what we have on user_interface and field-editors. 
As we do have some requests on reference/EntityStore, I've tried to refactor them here: 21b35fe5b18eb6b6d9aef98b6072e1516717debb and 835755c2fd9aa91a92afad31ae217ca962127b50.
WDYT @chrishelgert?

packages/rich-text/src/plugins/Hyperlink/useEntityInfo.ts
joseliacosta: Yeah, following the strategy I shared on the previous comment, I've replaced it e18b0e16ea94357bd827399c9ed62b148b24379e
Thanks for the suggestion 🙏🏼

cypress/component/rich-text/RichTextEditor.Commands.spec.ts
joseliacosta: After I started using react-query to get the content-type, these tests stop working. I suspect there is some mismatching between the cypress synthetic keyboard events and async calls but I am not sure yet.
However, changing the behavior of the test just to make it pass is not so great 🙈. Thanks for pointing this during the code review 🙇🏼  

I am trying to understand how this test was passing because [from the playground](https://contentful-field-editors.colorfuldemo.com/?path=/docs/editors-rich-text-editor--docs), it is not :(

 I will time box and investigate a bit more before merging this PR 🙏🏼

cypress/component/rich-text/RichTextEditor.Commands.spec.ts
joseliacosta: I thought we needed because now `useContentTypes` became a promise but actually we don't. So, removed :) I am still trying to rollback the tests to the original assert (keyboard and so on).

cypress/component/rich-text/RichTextEditor.Commands.spec.ts
joseliacosta: The test is fixed! Actually I was surprised it was working before but I debugged it and made the test reproduce the same behavior we expected and have working on prod :)

packages/_shared/src/hooks/useContentTypes.ts
```diff
@@ -0,0 +1,119 @@
...
+      // This ensures the cache is refreshed if schema changes were made
+      if (oldSlideLevel > newSlideLevel) {
+        void queryClient.invalidateQueries(
+          createGetManyContentTypesKey(spaceId, environmentId, queryParams),
+        );
+      }
+    });
+
+    return unsubscribe as () => void;
+  }, [navigator, queryClient, spaceId, environmentId, queryParams]);
```
chrishelgert: theoretically only if it's not the shared client 🤔 
otherwise we clear too often as it will be cleared via PubSub normally

chrishelgert: (we dont have that for entries, thats why we do it there on slidein navigation)

packages/_shared/src/queryKeys.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Creates a query key for fetching a single asset.
+ * Matches the structure used in user_interface for cache sharing.
+ *
+ * Source: https://github.com/contentful/user_interface/blob/main/src/javascripts/core/react-query/cma/asset/useGetAsset.ts
+ *
+ * @param spaceId - The space ID
+ * @param environmentId - The environment ID
+ * @param assetId - The asset ID to fetch
+ * @returns Query key array for React Query
```
colomolo, line 25: Are these really needed since this is TS file?

packages/_shared/src/hooks/useContentTypes.ts
```diff
@@ -0,0 +1,119 @@
...
+ * Simple helper to fetch a single content type by ID.
+ * For use in async functions or one-off fetches.
+ * For repeated fetches with caching, use the useContentType hook instead.
+ */
+export async function fetchContentType(
+  sdk: Pick<FieldAppSDK, 'cma'>,
+  contentTypeId: string,
+): Promise<ContentType> {
+  return sdk.cma.contentType.get({ contentTypeId });
+}
```
colomolo, line 45: It just replaces `sdk.cma.contentType.get({ contentTypeId })` call with `fetchContentType(sdk, contentTypeId)` ? What's the benefit?

packages/_shared/src/QUERY_KEYS_SYNC.md
```diff
@@ -0,0 +1,58 @@
...
+The **user_interface** repository is the source of truth for query key structure:
+
+- Repository: `contentful/user_interface`
+- Location: `src/javascripts/core/react-query/cma/`
+
+## How to Keep In Sync
+
+### When Adding New Query Keys
+
+1. **Check user_interface first**: Look for the corresponding hook in `user_interface/src/javascripts/core/react-query/cma/[entity]/`
```
colomolo: Not sure that we should expose this info here.

packages/_shared/src/QUERY_KEYS_SYNC.md
```diff
@@ -0,0 +1,58 @@
...
+
+- ✅ Array structure matches exactly
+- ✅ Parameter order is identical
+- ✅ Conditional logic (e.g., omitting empty params) is the same
+- ✅ Space and environment IDs are included
+- ✅ Query parameters are included in the key when provided
+
+## Future Improvements
+
+Propose a shared `@contentful/query-keys` package that both repositories can depend on, eliminating the need for manual synchronization.
```
colomolo: +1

packages/_shared/src/queryKeys.ts
```diff
@@ -0,0 +1,109 @@
...
+  contentTypeId: string,
+) => {
+  return ['spaces', spaceId, 'environments', environmentId, 'content_types', 'get', contentTypeId];
+};
+
+/**
+ * Creates a query key for fetching multiple content types.
+ * Matches the structure used in user_interface for cache sharing.
+ *
+ * Source: https://github.com/contentful/user_interface/blob/main/src/javascripts/core/react-query/cma/contentType/useGetManyContentTypes.ts
```
colomolo: Also, you're exposing internals here.

packages/rich-text/src/__fixtures__/FakeSdk.ts
```diff
@@ -114,6 +114,13 @@ export function newReferenceEditorFakeSdk(props?: ReferenceEditorSdkProps): [Fie
           }
           return Promise.reject({});
         },
+        getMany: async () => ({
+          items: Object.values(contentTypes),
+          total: Object.values(contentTypes).length,
```
colomolo, line 109: ~~Can we calculate`Object.values(contentTypes)` once here?~~ oh, it's fixtures, nevermind then.

.swcrc
```diff
@@ -4,7 +4,6 @@
       "syntax": "typescript",
       "tsx": true
     },
-    "target": "es2020",
```
chrishelgert, line 7: might be dangerous, changing something here produced already a few issues in github - worth checking the output

packages/_shared/src/hooks/useContentTypes.ts
chrishelgert: we could also use `queryClient.fetchQuery(...)`
(example in entity store) where we cant use with the hook 🤔 
ofc the query client needs to be a property then 🤔

packages/_shared/src/QUERY_KEYS_SYNC.md
joseliacosta: @colomolo I will bring it to refinement soon: https://contentful.atlassian.net/browse/TOL-3764

.swcrc
joseliacosta: My bad, i accidentally committed it.  🙈 
When developing, I have to comment this line when running the package and webapp locally. Otherwise I am still facing a compilation error on webapp side 😞 
But I reverted it, thanks for catching, @chrishelgert

packages/_shared/src/queryKeys.ts
joseliacosta: No really but I think it can be nice for future us to see this hint on IDE, no?

PR #2045, opened 2026-01-15
refactor: add SharedQueryClientProvider and query hooks to `_shared` package [TOL-3472]

joseliacosta:
## Description
While working on https://github.com/contentful/field-editors/pull/2043, made sense to lift `react-query` client to `_shared`, so it can be reused and avoid a circular dependency between the `field-editors/packages`

## Changes:

✅ Moved React Query setup from reference to _shared
✅ Made reference re-export from _shared (maintains backwards compatibility)
✅ Set `@tanstack/react-query` as optional peer dependency
✅ Zero bundle impact on packages not using React Query!

## Benefits:

- Single source of truth for React Query configuration
- Other packages can now opt-in to using the shared client
- Paves the way for useContentTypes, introduced by https://github.com/contentful/field-editors/pull/2043 to optionally use caching.

### DIFF
--- a/packages/_shared/package.json
+++ b/packages/_shared/package.json
@@ -39,6 +39,7 @@
     "@contentful/app-sdk": "^4.42.0",
     "@contentful/field-editor-test-utils": "^1.7.0",
     "@lingui/core": "5.3.0",
+    "@tanstack/react-query": "^4.3.9",
     "@testing-library/react": "16.3.0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


joseliacosta:
> LGTM, is the bot question an issue though? Haven't used an optional peer dependency before so not sure if it is or it's just hallucinating an issue

The bot wasn't totally wrong, I adjusted some fix to avoid crashing on runtime. I tested canary and apparently, all good!

PR #2060, opened 2026-01-23
fix: revert "fix(references): improve fetching of scheduled actions"

z0al:
Reverts contentful/field-editors#2054


packages/reference/src/common/EntityStore.tsx
```diff
@@ -15,7 +15,6 @@ import {
   fetchAll,
 } from 'contentful-management';
 import { get } from 'lodash';
-import moment from 'moment';
```
chrishelgert, line 18: we shouldnt have used this anyway 🙈

z0al: It's already a dependency of the package, what do you mean?

chrishelgert: yeah i know but maybe we should not use it on new places as there were plans to drop moment for a long time 😓

PR #2136, opened 2026-04-14
fix(date): trigger onChange while typing if input is already parseable [ZEND-7743]

chrishelgert:
Previously onChange only fired on blur, so time values typed without blurring (e.g. clicking Publish immediately) were never saved. Now onChange fires on every change if the current input parses to a valid time, without reformatting the display string. Blur still normalises the display.


YvesRijckaert (approved):
🙏

2wce (approved):

PR #2137, opened 2026-04-16
fix(date): crash when rendering DateEditor with midnight in 24h mode

chrishelgert:
## Summary

- Fixes a crash when `DateEditor` renders a value with midnight time (e.g. `2023-05-03T00:00+02:00`) in 24h clock mode
- Root cause: `TimepickerInput` useEffect always parsed `time + ampm` with the 12h format `hh:mm a`; the `hh` token expects 1–12, so `'00'` (midnight in 24h) produced an `Invalid Date`, which then crashed `format()`
- Fix: when `uses12hClock=false`, parse `time` directly with `HH:mm` instead

### DIFF
--- a/packages/date/src/TimepickerInput.spec.tsx
+++ b/packages/date/src/TimepickerInput.spec.tsx
@@ -1,0 +1,42 @@
+import * as React from 'react';
+
+import '@testing-library/jest-dom/extend-expect';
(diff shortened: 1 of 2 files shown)


2wce (approved):
can we also add a test for the scenario

PR #2138, opened 2026-04-17
feat: [TOL-515] Add embeds as last element

colomolo:
When RTE is not focused, by default we insert embedded element as the last node. If there's an empty paragraph node, we replace it with block element or insert inline element. This aligns embed behavior to, for example, how table injection works currently.

### DIFF
--- a/packages/rich-text/src/plugins/shared/EmbeddedBlockUtil.ts
+++ b/packages/rich-text/src/plugins/shared/EmbeddedBlockUtil.ts
@@ -9,6 +9,7 @@ import {
 } from '../../helpers/config';
 import {
   focus,
+  getAncestorPathFromSelection,
   getNodeEntryFromSelection,
   insertEmptyParagraph,
   moveToTheNextChar,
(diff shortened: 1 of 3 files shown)


packages/rich-text/src/plugins/shared/EmbeddedBlockUtil.ts
```diff
@@ -72,16 +78,12 @@ export async function selectEntityAndInsert(
...
-  const { selection } = editor;
   const rteSlide = watchCurrentSlide(sdk.navigator);
   const entity = await selectEntity(config);
 
   if (!entity) {
     logAction('cancelCreateEmbedDialog', { nodeType });
   } else {
-    // Selection prevents incorrect position of inserted ref when RTE doesn't have focus
-    // (i.e. when using hotkeys and slide-in)
-    select(editor, selection);
```
colomolo, line 84: This part threw error, because with no focus `selection` is null. Seems to doesn't affect anything.

chrishelgert: did you try hotkeys and co in the web app to make sure it still works?
otherwise it might be better to just guard , to only call it if there is a selection

colomolo: @chrishelgert what's the hotkey for Embed? Found that it's actually impossible to insert an item with keyboard only. When navigating to the card, Space/Enter does nothing.

colomolo: Okay, for embed it's Cmd+Shift+E, but it works only when RTE is focused. So it's fine

chrishelgert (approved):

PR #2139, opened 2026-04-21
fix(date): correct 24h time snap-back and timezone date shift [ES-141][ES-143][ES-145]

chrishelgert:
## Summary

- **24h snap-back**: `TimepickerInput.handleBlur` was always emitting `hh:mm` (12h format) via `format(value, 'hh:mm')`, even in 24h mode. Entering `19:00` would fire `onChange({ time: '07:00', ampm: 'PM' })`, causing the field to snap to `07:00` on blur. Fixed by branching on `uses12hClock`.
- **Noon parsed as midnight**: `parseRawInput` tried `'hh:mm'` before any 24h format — `12:00` matched as 12h with no AM/PM suffix, defaulting to AM (midnight). Fixed by inserting `'HH:mm'` before `'hh:mm'` in `validInputFormats`.
- **Date shift across midnight**: `fieldValueToDate` used `parseISO` which converts offset-aware strings to a UTC instant. For e.g. `2026-04-27T23:00-02:00` (= `2026-04-28T01:00Z`), the local Date object carried April 28, causing the stored value to shift by one day. Fixed by extracting the `yyyy-MM-dd` part directly from the ISO string, bypassing timezone conversion entirely.

### DIFF
--- a/packages/date/src/TimepickerInput.spec.tsx
+++ b/packages/date/src/TimepickerInput.spec.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import '@testing-library/jest-dom/extend-expect';
-import { cleanup, configure, render } from '@testing-library/react';
+import { cleanup, configure, fireEvent, render } from '@testing-library/react';
 
 import { TimepickerInput } from './TimepickerInput';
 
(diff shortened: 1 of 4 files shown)


2wce (approved):

joseliacosta (approved):
Nice test coverage 🙏🏼 ✨

PR #2146, opened 2026-04-23
feat: remove moment.js from reference and rich-text packages [HK-000]

chrishelgert:
## Summary

- Replace `moment.js` with `date-fns` in `formatDateAndTime` utilities for both `packages/reference` and `packages/rich-text`
- Add `date-fns ^2.30.0` as a dependency and remove `moment` from both packages


packages/reference/src/components/ScheduledIconWithTooltip/formatDateAndTime.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    expect(formatDate(d)).toBe(`Tomorrow, ${format(d, 'dd MMM yyyy')}`);
+  });
+
+  it('returns "Yesterday" (short) for the previous calendar day', () => {
+    expect(formatDate('2024-06-14T08:00:00.000Z', true)).toBe('Yesterday');
+  });
+
+  it('returns long form for yesterday', () => {
+    const d = new Date('2024-06-14T08:00:00.000Z');
+    expect(formatDate(d)).toBe(`Yesterday, ${format(d, 'dd MMM yyyy')}`);
```
z0al: Nit: Since the date value is constant can you hard-code the expected value here instead of calling format again please? same for other tests. That way it's more clearly what is expected

z0al (approved):

packages/reference/src/components/ScheduledIconWithTooltip/formatDateAndTime.spec.ts
chrishelgert: was already on it 🤝

chrishelgert: here was also the same expectactions running against moment: https://github.com/contentful/field-editors/pull/2146/commits/0fc9b78ffb2f026d58699882222ca430cce16742

joseliacosta (approved):

PR #2147, opened 2026-04-23
fix(slug): don't convert $ to usd

z0al:
`$` today gets converted to `usd` but that can be misleading if the locale is English (New Zealand) or English (Australia). Since the underlying library doesn't support distinct locale codes for English variants the simplest fix is to get rid of the symbol completely. 

Edit: in future, we may want to consider disabling `symbols` option in `speakingurl` which removes the $ and all sorts of other symbols e.g. emojis


2wce (approved):

chrishelgert (commented):
Is this a breaking change? 🙈

z0al:
> Is this a breaking change? 🙈

@chrishelgert technically yes but I hope it won't be too breaking 🙈

PR #2152, opened 2026-04-30
feat(shared): widen entity-shaped sys params

YvesRijckaert:
Widen the entity-shaped `sys` parameters in `_shared/` from entry / asset-specific types to the structural `EntitySys` shape.
The goal is to make shared helpers work across all supported entity types, including `Entry`, `Asset`, `Experience`, and `Fragment`.

### DIFF
--- a/packages/_shared/src/hooks/useLocalePublishStatus.ts
+++ b/packages/_shared/src/hooks/useLocalePublishStatus.ts
@@ -3,7 +3,7 @@ import { useMemo } from 'react';
 import type { LocalesAPI } from '@contentful/app-sdk';
 import type { LocaleProps } from 'contentful-management';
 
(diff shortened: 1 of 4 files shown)


packages/_shared/src/hooks/useLocalePublishStatus.ts
```diff
@@ -15,7 +15,7 @@ export type LocalePublishStatus = {
 export type LocalePublishStatusMap = Map<string, LocalePublishStatus>;
 
 function getLocalePublishStatusMap(
-  entity: Pick<AssetProps | EntryProps, 'sys'>,
+  entity: { sys: entityHelpers.EntitySys },
```
chrishelgert: can we just import type EntitySys and not take it from entityHelpers please?

chrishelgert (approved):

PR #2167, opened 2026-05-27
chore: migrate test runner from jest to vitest

chrishelgert:
## Summary

Migrate the entire test suite (21 packages, ~300+ tests) from Jest to Vitest, using `@vitejs/plugin-react-swc` to keep the existing SWC transform.

- Per-package `vitest.config.ts` extending a shared `vitest.shared.ts` factory; per-package JUnit reports written to `reports/<pkg>-results.xml`.
- Explicit imports throughout (`import { describe, it, expect, vi } from 'vitest'`) — no auto-globals. `jest.*` → `vi.*`. `jest.requireActual` → `vi.importActual`. `jest.Mocked<T>` → `Mocked<T>`.
- Single root `vitest.setup.ts` provides `@testing-library/jest-dom/vitest`, lingui macro mocks, and `afterEach(cleanup)` (Vitest doesn't auto-cleanup like Jest did).
- `reference` keeps its `TZ='UTC'` override; `rich-text` overrides the React plugin to use the classic JSX runtime for `plate-test-utils` `/** @jsx jsx */` pragmas.
- Removed root jest infra: `baseJestConfig.js`, `jest.setup.js`, and deps `jest`, `jest-cli`, `jest-environment-jsdom`, `jest-junit`, `@swc/jest`, `@types/jest`, `eslint-plugin-jest`.
- Test logic and expectations unchanged.


.circleci/config.yml
```diff
@@ -51,7 +51,6 @@ jobs:
       - run: yarn build
       - run: yarn tsc
   unit-tests:
-    parallelism: 3
```
chrishelgert, line 54: funny story, we run all the unit tests always 3 times 🙄

chrishelgert:
Reduced unit test time to ~40 from ~55-60s (install node_modules takes longer than unit tests 😓 )

2wce (approved):
Maybe we can improve install time by switching to pnpm?

chrishelgert:
we have a ticket for that in the backlog 👍

PR #2182, opened 2026-07-01
fix: show max char count capped by technical limits, add missing translations [ES-443]

MayaGillilan:
* Caps char validation to the technical limit based on field type, so that users are not confused by custom limits that go beyond them
* Adds a few missing translations to char counters & validation

### DIFF
--- a/packages/_shared/src/utils/constraints.ts
+++ b/packages/_shared/src/utils/constraints.ts
@@ -29,7 +29,7 @@ export function fromFieldValidations(
     return {
       type: 'min-max',
       min,
-      max: Math.max(max, MAX_LIMITS[fieldType]),
+      max: Math.min(max, MAX_LIMITS[fieldType]),
     };
   } else if (isNumber(min)) {
     return {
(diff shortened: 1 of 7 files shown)


chrishelgert (approved):

packages/_shared/src/utils/constraints.ts
```diff
@@ -29,7 +29,7 @@ export function fromFieldValidations(
     return {
       type: 'min-max',
       min,
-      max,
+      max: Math.max(max, MAX_LIMITS[fieldType]),
```
chrishelgert: shouldnt it be Math.min?
as we want the one with the lower one to apply

chrishelgert: e.g. techinical limit is 200 but the given one to the field is 1000000 we want to use 200 instead of the overwriten one which would not work 🤔

MayaGillilan: Duh 😣 thanks, silly mistake

chrishelgert (approved):

chrishelgert (approved):

PR #2185, opened 2026-07-10
fix: [TOL-4271] normalize react-query v5 loading status across field editors

chrishelgert:
## Summary

Fixes [TOL-4271](https://contentful.atlassian.net/browse/TOL-4271) — under `@tanstack/react-query` v5, four rich-text inline components crash reading `entry.sys` because their loading-status guards only match v4's `'loading' | 'idle'` strings. v5 renamed `'loading'` to `'pending'`, so the guard falls through and the success branch runs on undefined data.

Our packages already declare `"@tanstack/react-query": "^4.3.9 || ^5.0.0"` as a peer dep, so we must actually support both — this PR delivers on that promise.

### What changed

- Adds `isLoadingStatus(status: string | undefined): boolean` to `@contentful/field-editor-shared` under its `./react-query` submodule export. Returns `true` for `'loading' | 'idle' | 'pending'`.
- Fixes the four crashing components with regression tests that mock the fetch hook to return `status: 'pending'` and assert the loading card renders (each test faithfully reproduces the pre-fix crash — verified against pre-fix code).
- Normalizes 9 other in-repo loading-status checks (across `reference`, `rich-text`, `rich-text-alpha`) onto the same helper. Three drifted variants of the check are gone. Zero intended behavior change (verified: none of these sites pass `enabled: false`, so `'idle'` was already unreachable in practice).

### Version bumps (via Lerna conventional commits)

- `@contentful/field-editor-shared` — minor (`feat:` — new export)
- `@contentful/field-editor-reference` — patch
- `@contentful/field-editor-rich-text` — patch
- `@contentful/field-editor-rich-text-alpha` — patch

## Test plan

- [x] Unit tests for `isLoadingStatus` (`'loading' | 'idle' | 'pending'` → true; `'success' | 'error' | undefined` → false)
- [x] Regression test for each of the 4 crash-fix sites — mocks fetch hook with `status: 'pending'`, asserts loading state renders (verified: each test crashes against pre-fix code)
- [x] `yarn workspace @contentful/field-editor-shared test:ci` — 148/148 pass
- [x] `yarn workspace @contentful/field-editor-reference test:ci` — 67/67 pass
- [x] `yarn workspace @contentful/field-editor-rich-text test:ci` — 174/174 pass
- [x] `yarn workspace @contentful/field-editor-rich-text-alpha test:ci` — 3/3 pass
- [x] `tsc` clean across all 4 affected packages
- [x] `yarn workspace @contentful/field-editor-rich-text lint` — 0 errors (20 pre-existing warnings in untouched files)
- [x] Remnant sweep: `rg "(status as string) === 'pending'" packages/` returns no matches
- [ ] Consumer verification: install this branch's `@contentful/field-editor-rich-text` in an app hoisting `@tanstack/react-query@5` and confirm inline entry cards no longer crash on load

[TOL-4271]: https://contentful.atlassian.net/browse/TOL-4271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


colomolo (approved):
Yeah, changes make sense to me. So that we can support both. 👍

PR #2193, opened 2026-07-27
fix(markdown): cap paragraph indentation at 2 spaces to prevent code block rendering [ES-507]

2wce:
## Summary

Fixes a regression introduced in 2.3.2 where the indent/dedent toolbar buttons were disabled entirely for non-list lines. This removed a feature customers actively use (ES-507: Prada reported that indenting plain paragraphs via the toolbar caused the preview to render as a `<pre>` code block).

The root cause: `remark` (the preview renderer) treats 4+ leading spaces as a code block. The 2.3.2 fix blocked indentation on plain paragraphs altogether rather than preventing the threshold from being crossed.

Ticket: https://contentful.atlassian.net/browse/ES-507

## What changed

**New `packages/markdown/src/utils/indentation.ts`** — single source of truth for all indentation decisions:
- `canIndent(line, indentUnit)` — true for list items always; true for plain paragraphs only if adding `indentUnit` spaces would stay at or below 2 leading spaces
- `canDedent(line)` — true whenever the line has any leading spaces
- `MAX_PARAGRAPH_LEADING_SPACES = 2` — named constant with a comment explaining the remark 4-space threshold

**Three call sites, one decision function:**
- `MarkdownCommands.ts` — `indent`/`dedent` now clamp themselves via the util; correctness no longer depends on the toolbar button being disabled upstream
- `CodeMirrorWrapper.ts` — Tab key uses `cursor.ch > leadingSpaces` (correct discriminator: is the cursor in the indentation zone or the text body?) combined with `canIndent`
- `MarkdownEditor.tsx` / `ZenModeModalDialog.tsx` — React state split into separate `canIndentLine`/`canDedentLine` flags, both purely advisory for button affordance

**`MarkdownToolbar.tsx`** — `indentationDisabled` replaced with separate `indentDisabled` and `dedentDisabled` props, allowing the two buttons to reflect their independent states.

Also fixes a latent dedent bug: the previous implementation used `lineStartsWith(indentation)` which could not dedent a line with fewer leading spaces than `indentUnit` (e.g. a 1-space-indented line with `indentUnit = 2`). The new `Math.min(unit, leadingSpaces)` always removes cleanly.

## Why 2 spaces and not 3

The all-or-nothing approach: each indent press always inserts a full `indentUnit` (2 spaces), or is blocked entirely. A partial last step (0 → 2 → 3) would insert different amounts for the same action, which is confusing. Capping at 2 means the sequence is always `0 → 2 → blocked` — consistent and predictable.

## Why 4 spaces triggers the bug

Verified directly against the renderer in use:

```
unified().use(remarkParse).use(remarkGfm).parse('   text').children[0].type  // → 'paragraph'
unified().use(remarkParse).use(remarkGfm).parse('    text').children[0].type // → 'code'
```

Not an assumption from the CommonMark spec — confirmed against the actual parse tree.

## Test plan

- [ ] Plain paragraph with no indentation: indent button enabled, dedent button disabled
- [ ] After one indent press (2 spaces): indent button disabled, dedent button enabled
- [ ] Dedent returns to 0 spaces: indent button re-enabled, dedent button disabled
- [ ] List item: indent and dedent both work freely (no cap)
- [ ] Tab key on empty line: inserts 2 spaces
- [ ] Tab key on a 2-space-indented line: no change (blocked)
- [ ] Tab key mid-line (cursor past indentation zone): inserts spaces freely
- [ ] Preview tab: indented plain paragraph renders as paragraph, not `<pre>`
- [ ] Cypress component tests: `npm run cy:run -- --spec "cypress/component/markdown/MarkdownEditorSimpleActions.spec.ts"`

### DIFF
--- a/cypress/component/markdown/MarkdownEditorSimpleActions.spec.ts
+++ b/cypress/component/markdown/MarkdownEditorSimpleActions.spec.ts
@@ -427,14 +427,37 @@ describe('Markdown Editor / Simple Actions', () => {
       clickVisibleButtonByName('Decrease indentation');
     };
 
-    it('disables indentation for a plain paragraph', () => {
+    it('enables indent button for a plain paragraph with no indentation', () => {
       renderMarkdownEditor();
 
       unveilAdditionalButtonsRow();
       type('A plain paragraph');
 
-      selectors.getIndentButton().should('be.disabled');
+      selectors.getIndentButton().should('not.be.disabled');
       selectors.getDedentButton().should('be.disabled');
+    });
+
+    it('disables indent button for a plain paragraph at the 2-space cap', () => {
+      renderMarkdownEditor({ spyOnSetValue: true });
+
+      unveilAdditionalButtonsRow();
+      type('A plain paragraph');
+      clickIndentButton();
+      checkValue('  A plain paragraph');
+
+      selectors.getIndentButton().should('be.disabled');
+      selectors.getDedentButton().should('not.be.disabled');
+    });
+
+    it('indents and dedents a plain paragraph up to the 2-space cap', () => {
+      renderMarkdownEditor({ spyOnSetValue: true });
+
+      unveilAdditionalButtonsRow();
+      type('A plain paragraph');
+      clickIndentButton();
+      checkValue('  A plain paragraph');
+      clickDedentButton();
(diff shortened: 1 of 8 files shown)


z0al:
Closed as not needed

PR #2195, opened 2026-07-29
refactor: convert field editor class components [TOL-3560]

YvesRijckaert:
## Summary
- Convert remaining React class components in field editors to function components
- TypeScriptify HyperlinkDialog by replacing the JSX component with a TSX implementation
- Preserve connector/editor lifecycle behavior with hooks, refs, and cleanup effects

Manual testing can be done with this URL: [https://app.contentful.com?ui_version=TOL-3560](https://app.contentful.com/?ui_version=TOL-3560)


packages/slug/src/TrackingFieldConnector.tsx
```diff
@@ -27,84 +27,68 @@ function getTitleField(sdk: FieldAppSDK, trackingFieldId?: string) {
...
-    this.state = {
+    return {
       titleValue: titleField ? titleField.getValue() : '',
       isPublished: Boolean(entrySys.publishedVersion),
-      isSame,
+      isSame: titleField ? props.field.id === titleField.id : false
     };
-  }
+  });
+  const propsRef = React.useRef(props);
```
chrishelgert: NIT intead of `React.` could we just use `useRef`, ...

chrishelgert (approved):
code overall looks good, please also check on the web app if we have now more/less rerenders 🙇

packages/_shared/src/FieldConnector.ts
```diff
@@ -29,129 +29,121 @@ type FieldConnectorProps<ValueType> = {
...
-      return deepEqual(value1, value2);
-    },
-    debounce: 300,
-  };
+const defaultIsEmptyValue = (value: unknown) => value === null || value === '';
+const defaultIsEqualValues = (value1: unknown, value2: unknown) => deepEqual(value1, value2);
 
-  constructor(props: FieldConnectorProps<ValueType>) {
-    super(props);
+export function FieldConnector<ValueType>(props: FieldConnectorProps<ValueType>) {
```
chrishelgert, line 46: wondering if that could be a hook instead `useFieldConnector` instead of an HoC

YvesRijckaert: Wouldn't that break the existing API though? We are using `<FieldConnector>` as a component everywhere right now, including in packages outside of field-editors

PR #2227, opened 2026-08-05
fix(rich-text): prevent heading hotkey from firing inside table cells [ES-533]

joseliacosta:
## Summary

Heading toggle shortcuts (`mod+alt+1` … `mod+alt+6`) could fire while the selection was inside a table cell, toggling a heading node type that table cells don't support. The schema normalizer would then immediately revert the change, producing a visible cursor jump with no other effect.

This PR makes `buildHeadingEventHandler` bail out early when the current selection sits inside a `TABLE_CELL` or `TABLE_HEADER_CELL`, so the keystroke falls through as ordinary character input instead of toggling a heading.

Typing `€` (`Ctrl`+`Alt`+`5`) inside a table cell will now correctly insert `€` instead of jumping columns, and `Ctrl`+`Alt`+`5` used anywhere else in the document still converts text to `Heading 5` as expected.

## Scope

- Only the table-cell case is changed.
- Heading toggling in plain paragraphs (toolbar, Command Palette, and hotkey) is unaffected — covered by a regression test.
- No API/schema changes.

## Test plan

- Added unit tests for `buildHeadingEventHandler`/`onKeyDown` covering: table cell, table header cell, a list item nested inside a table cell, and a plain-paragraph regression case.
- `yarn lint`, `yarn tsc`, and `yarn vitest run` all pass for `packages/rich-text` (178/178 tests, no regressions).

### DIFF
--- a/packages/rich-text/src/plugins/Heading/__tests__/createHeadingPlugin.test.tsx
+++ b/packages/rich-text/src/plugins/Heading/__tests__/createHeadingPlugin.test.tsx
@@ -1,8 +1,10 @@
 /* eslint-disable react/no-unknown-property */
 /** @jsx jsx */
-import { describe, it } from 'vitest';
+import { BLOCKS } from '@contentful/rich-text-types';
+import { describe, expect, it, vi } from 'vitest';
 
-import { assertOutput, jsx } from '../../../test-utils';
+import { assertOutput, createTestEditor, jsx } from '../../../test-utils';
+import { createHeadingPlugin } from '../createHeadingPlugin';
 
 describe('normalization', () => {
   it('can contain inline entries & hyperlinks', () => {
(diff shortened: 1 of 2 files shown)


packages/rich-text/src/plugins/Heading/__tests__/createHeadingPlugin.test.tsx
```diff
@@ -293,3 +295,149 @@ describe('normalization', () => {
...
+
+    expect(
+      editor.children.some((n: any) =>
+        JSON.stringify(n).includes(`"type":"${BLOCKS.HEADING_5}"`),
+      ),
+    ).toBe(false);
+    expect(trackingHandler).not.toHaveBeenCalled();
+  });
+
+  it('still toggles a heading in a plain paragraph outside a table (ZEND-6088 regression guard)', () => {
```
YvesRijckaert: please let's not add jira ticket numbers in the code :)

joseliacosta: ops! Thanks for catching it 🙏🏼

luizfonseca (approved):
That's some high level detective work :D

chrishelgert:
Wouldn't it be the same problem with paragraphs and co?
According to this a list item can be a headline, so you wouldnt be able to use hotkeys to assign them?
https://github.com/contentful/rich-text/blob/3b796938337959c30fd4792828ec269b132f4f73/packages/rich-text-types/src/schemaConstraints.ts#L44

Wondering if it breaks more assumptions than it fixes?

@z0al for input

chrishelgert:
> Wouldn't it be the same problem with paragraphs and co? According to this a list item can be a headline, so you wouldnt be able to use hotkeys to assign them? https://github.com/contentful/rich-text/blob/3b796938337959c30fd4792828ec269b132f4f73/packages/rich-text-types/src/schemaConstraints.ts#L44
> 
> Wondering if it breaks more assumptions than it fixes?
> 
> z0al for input

nvm, it's tables not lists

packages/rich-text/src/plugins/Heading/createHeadingPlugin.ts
```diff
@@ -13,6 +13,14 @@ const buildHeadingEventHandler =
   (editor, { options: { hotkey } }) =>
   (event) => {
     if (editor.selection && hotkey && isHotkey(hotkey, event)) {
+      if (
+        getAboveNode(editor, {
```
chrishelgert, line 17: generally the hotkey should not trigger if the block is not not supported inside of this node

chrishelgert (approved):

packages/rich-text/src/plugins/Heading/createHeadingPlugin.ts
chrishelgert: i think we had this once as a ticket in the backlog 😓

luizfonseca:
> Wouldn't it be the same problem with paragraphs and co? According to this a list item can be a headline, so you wouldnt be able to use hotkeys to assign them? https://github.com/contentful/rich-text/blob/3b796938337959c30fd4792828ec269b132f4f73/packages/rich-text-types/src/schemaConstraints.ts#L44
> 
> Wondering if it breaks more assumptions than it fixes?
> 
> z0al for input

btw, this is a valid consideration for the other types that shouldn't have heading within it that might also fail (`QUOTE` for example), but the question would be if we need to fix it/if it's even an issue atm. For the tables this seems to be isolated enough ? 🤔

z0al (approved):
Ideally we don't have to do these hacks but thanks to slate we can't do that easily with the normalisation-based architecture. 

Since this looks contained to tables I'm ok with the change but I think the bigger issue should be addressed as a follow up with product: we have a conflict with OS shortcuts, maybe we should follow Google Doc example and mirror their shortcuts

